PR #12, opened 2021-05-04
feat: Implement basic menu builder

lemarier:
Implement basic menu builder for macOS and windows.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

Basic menu builder for windows, macOS and Linux.
<s>I've checked Linux and it shouldn't be hard to add. I think it shouldn't be a blocker to get started with `wry` integration as it uses the same struct and</s> right now it's a simple `noop` like ios and android.

### API

Please see the [example](https://github.com/tauri-apps/tao/blob/feat/menu-bar/examples/custom_menu.rs#L15-L61)

In macOS, it uses the latest registered menu, if you open 2 windows with different menus, the latest window opened will be the application menu.

Declaring [custom menu](https://github.com/tauri-apps/tao/blob/feat/menu-bar/examples/custom_menu.rs#L55-L59) can include an accelerator (keyboard shortcut). (only supported on macOS right now -- windows seems quite complex) 

Customs menu is supported on Windows and macOS.

The menu builder exposes pre-defined menu action and they are only supported on macOS for now. 
In windows, nothing is rendered.

Click [events](https://github.com/tauri-apps/tao/blob/feat/menu-bar/examples/custom_menu.rs#L78-L93) are sent into the event loop with the `unique_menu_id` declared when we [create a custom menu](https://github.com/tauri-apps/tao/blob/feat/menu-bar/examples/custom_menu.rs#L33)

I think in wry maybe we should provide a default menu for macOS, to include copy, paste, select all, etc... to fix keyboard shortcuts, but the user can overwrite it? In windows, we can make it optional.

### Run the example
```
cargo run --example custom_menu
```

### macOS
https://user-images.githubusercontent.com/22237916/117036350-54759100-acd3-11eb-9008-8233dd0dbaa4.mov

### Windows
https://user-images.githubusercontent.com/22237916/117037201-352b3380-acd4-11eb-9a7a-7db5fdd71bed.mov

### DIFF
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -4,13 +4,13 @@ use std::{
 };
 
 #[derive(Debug, Clone)]
-pub struct Menu<'a> {
+pub struct Menu {
   pub title: String,
-  pub items: Vec<MenuItem<'a>>,
+  pub items: Vec<MenuItem>,
 }
 
-impl<'a> Menu<'a> {
-  pub fn new(title: &str, items: Vec<MenuItem<'a>>) -> Self {
+impl Menu {
+  pub fn new(title: &str, items: Vec<MenuItem>) -> Self {
     Self {
       title: String::from(title),
       items,
@@ -20,17 +20,17 @@ impl<'a> Menu<'a> {
 
 #[derive(Debug, Clone, Hash, Copy)]
 /// CustomMenu is a custom menu who emit an event inside the EventLoop.
-pub struct CustomMenu<'a> {
+pub struct CustomMenu {
   pub _id: MenuId,
-  pub name: &'a str,
+  pub name: &'static str,
   pub keyboard_accelerators: Option<&'static str>,
 }
 
 /// A menu item, binded to a pre-defined action or `Custom` emit an event.
 #[derive(Debug, Clone, Copy)]
-pub enum MenuItem<'a> {
+pub enum MenuItem {
   /// A custom menu emit an event inside the EventLoop.
-  Custom(CustomMenu<'a>),
+  Custom(CustomMenu),
 
   /// Shows a standard "About" item
   ///
@@ -38,7 +38,7 @@ pub enum MenuItem<'a> {
   ///
   /// - **Windows / Linux / Android / iOS:** Unsupported (noop).
   ///
-  About(&'a str),
+  About(&'static str),
 
   /// A standard "hide the app" menu item.
   ///
@@ -171,10 +171,10 @@ pub enum MenuItem<'a> {
   Separator,
 }
 
-impl<'a> MenuItem<'a> {
+impl MenuItem {
   /// Create new custom menu item.
   /// unique_menu_id is the unique ID for the menu item returned in the EventLoop `Event::MenuEvent(unique_menu_id)`
-  pub fn new(title: &'a str) -> Self {
+  pub fn new(title: &'static str) -> Self {
     MenuItem::Custom(CustomMenu {
       _id: MenuId::new(title),
       name: title,
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -14,6 +14,7 @@ impl Menu {
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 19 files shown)


src/menu.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+#[derive(Debug, Clone)]
+/// CustomMenu is a custom menu who emit an event inside the EventLoop.
+pub struct CustomMenu {
+  // todo: replace id by a type that
+  // we can send from the with_menu::<T>()
+  // or we could check to use the UserEvent::<T> but
+  // my latest test failed
+  pub id: String,
```
wusyong: Maybe `id` can be the hash of the `name` and users could retrieve a `MenuId` from the method.
This could also remove the usage of thread local on Windows.

src/menu.rs
```diff
@@ -0,0 +1,195 @@
...
+#[derive(Debug, Clone)]
+pub enum MenuItem {
+  /// A custom menu emit an event inside the EventLoop.
+  Custom(CustomMenu),
+
+  /// Shows a standard "About" item
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows/Linux:** Unsupported (noop).
```
wusyong: Lets list them all lol
```suggestion
  /// - **Windows / Linux / Android / iOS:** Unsupported (noop).
```

src/platform_impl/linux/window.rs
```diff
@@ -402,6 +403,16 @@ impl Window {
...
   }
 
+  /// Set menu
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows/Linux:** Unsupported (noop).
+  ///
+  pub fn set_menu(&self, _menu: Option<Vec<Menu>>) {
+    debug!("`Window::set_menu` is ignored on linux")
```
wusyong, line 417: I can help implementing this.

wusyong: image
Initial implementation looks promising. Do you want me to push it or wait until next PR?

src/menu.rs
lemarier: Sounds good! I wonder if they have the same `name` it can cause issue? 

Because, i wanted to use the same menu builder, and same event loop event for the tray app menu.

It may be harder to make different actions, if you have a menu  example `Quit` in the menu bar and `Quit` in the tray menu

src/platform_impl/linux/window.rs
lemarier: Sure!

wusyong: great! I'll push some implementation first.

src/menu.rs
wusyong: Hmm maybe we could give them different type like `MenuId` & `TrayId`.

lemarier: Done. Let me know what you think

src/menu.rs
```diff
@@ -1,43 +1,44 @@
...
-pub struct CustomMenu {
-  // todo: replace id by a type that
-  // we can send from the with_menu::<T>()
-  // or we could check to use the UserEvent::<T> but
-  // my latest test failed
-  pub id: String,
-  pub name: String,
+pub struct CustomMenu<'a> {
+  pub _id: MenuId,
+  pub name: &'a str,
```
wusyong: I would prefer `String` at this rate so we don't have to affect so many types with explicit lifetime.

lemarier: yeah the problem is i need the copy trait if we want to use what i’ve done with custom_menu.id() inside the event loop

wusyong (approved):
I'll merge this so we can get a cleaner PR of tray. Would also open a tracking issue for implementing menu item for Linux and Windows.

PR #70, opened 2021-05-19
add Window::request_close method

chippers:
This should be called from programs that want to close the associated window.
The event loop handler should decide if the CloseRequested event should do
anything.

(linux only until i know this interface will work)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
this is a linux only draft until we know this is the way we want to approach it.

Adds `Window::request_close` which triggers `WindowEvent::CloseRequested` with the window id. This allows code that only has a reference to the window (like in certain places in `wry`) to request the window to be closed programmatically. This comes up in wry when wanting to let `window.close()` in javascript work, or allowing webdriver clients to close the window after finishing their tests.

It is up to the event loop handler to decide what to do with the request, and dropping it if they wish to close the window.

### DIFF
--- a/src/platform_impl/linux/event_loop.rs
+++ b/src/platform_impl/linux/event_loop.rs
@@ -585,6 +585,14 @@ impl<T: 'static> EventLoop<T> {
                 menu.show_all();
               }
             }
+            WindowRequest::Close => {
+              if let Err(e) = event_tx.send(Event::WindowEvent {
+                window_id: RootWindowId(id),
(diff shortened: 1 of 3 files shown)


chippers:
we are going to use this method in wry, but without any additional code in `tao` as closing the window from the webview should trigger the same event anyways.

PR #78, opened 2021-05-29
feat: add `skip_taskabr` impl for windows

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

- [X] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)

**Other information:**

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -103,7 +103,7 @@ pub trait WindowExtWindows {
   /// Returns the current window theme.
   fn theme(&self) -> Theme;
 
-  /// Not to display window icon in the task bar.
+  /// Removes the window icon from the task bar.
   fn skip_taskbar(&self);
 }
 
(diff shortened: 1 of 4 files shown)


src/platform/windows.rs
```diff
@@ -102,6 +102,9 @@ pub trait WindowExtWindows {
 
   /// Returns the current window theme.
   fn theme(&self) -> Theme;
+
+  /// Not to display window icon in the task bar.
```
lemarier: Maybe replace with; Whether to display window icon in the taskbar.

lemarier (approved):

src/platform/windows.rs
amrbashir: yeah that sounds better, I will change the linux one too since that's where I copied it from xD

amrbashir: Actually, I think a better comment would be `Removes the window icon from the task bar` since `skip_taskbar` is a one way function (I think we can make it both way but it is better be in another PR ).

PR #93, opened 2021-06-22
feat(windows): allow menu update

lemarier:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -44,7 +44,7 @@ use crate::{
   menu::MenuType,
   monitor::MonitorHandle as RootMonitorHandle,
   platform_impl::platform::{
-    accelerator::register_accel,
+    accelerator::{drop_accel_table, register_accel},
     dark_mode::try_theme,
     dpi::{dpi_to_scale_factor, hwnd_dpi},
     drop_handler::FileDropHandler,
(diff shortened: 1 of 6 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -150,8 +151,29 @@ impl Window {
...
 
-  // TODO (lemarier): allow menu update
-  pub fn set_menu(&self, _new_menu: Option<menu::Menu>) {}
+  #[inline]
+  pub fn set_menu(&mut self, menu: Option<menu::Menu>) {
+    if let Some(menu) = &self.menu {
+      unsafe {
+        winuser::DestroyMenu(menu.0);
+      }
+      drop_accel_table();
```
amrbashir, line 160: why drop the accel_table? AccelTable should be global and not bound to a single menu or a single window.

src/platform_impl/windows/window.rs
```diff
@@ -150,8 +151,29 @@ impl Window {
     }
   }
 
-  // TODO (lemarier): allow menu update
-  pub fn set_menu(&self, _new_menu: Option<menu::Menu>) {}
+  #[inline]
+  pub fn set_menu(&mut self, menu: Option<menu::Menu>) {
+    if let Some(menu) = &self.menu {
+      unsafe {
+        winuser::DestroyMenu(menu.0);
```
amrbashir, line 158: if you destroy this HMENU, it will no longer be valid for other windows, so probably you just want to call `SetMenu(hwnd,NULL)` like you did a few lines later.

amrbashir (commented):
- Menus can be associated with different windows and should not be destroyed unless the user explicitly drops the menu and in this case we `impl drop for Menu` and call `DestroyMenu` there.

	Windows handle Menus and Accelerators weirdly, if you call `SetMenu(hwnd,Null)` it will remove the Menu from a window but the Accelerator will still be valid and will be triggered even if the window doesn't have the Menu anymore and I am not sure how linux Accelerators work or how we should handle this.

- We probably need to refactor the `AccelTable`, Accelerators should be associated with a `Menu` and not a `Window`, when we Drop the menu we remove its accelerators only and not the whole `AccelTable`

src/platform_impl/windows/window.rs
lemarier: Good point!

But, if we do not drop the accel_table, and menu had an accelerator registered, the event is trigerred twice — but yeah i think it may cause issue for other window, and wanted to avoir a static hashmap to keep an active menu list

src/platform_impl/windows/window.rs
lemarier: i’ll re-test but i think i had issues with setting the menu NULL

src/platform_impl/windows/window.rs
amrbashir: yeah we might need to keep a static hashmap

amrbashir:
This is now implemented in #297

PR #106, opened 2021-06-25
feat: implement `set_menu` for system tray

fnune:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**


fnune:
I'm not sure if this works. I'm not sure how `_window_target` is being passed to these functions.

src/system_tray.rs
```diff
@@ -113,4 +114,12 @@ impl SystemTray {
   pub fn set_icon(&mut self, icon: PathBuf) {
     self.0.set_icon(icon)
   }
+
+  pub fn set_menu<T: 'static>(
+    &mut self,
+    tray_menu: MenuPlatform,
```
amrbashir: tray_menu should be of type `ContextMenu` and not `MenuPlatform`

fnune:
Oh sweet! You're extending it. I'll have time to work on this tonight, or maybe over the weekend

lemarier:
Ok, so It works fine on `Linux` and `macOS`.

The only issue we have is `Windows`, once we set the menu the tray menu stops working.

Gonna try to dig in. @amrbashir you have an idea?

amrbashir:
yeah I saw that issue yesterday, It was because the `Menu` gets dropped which calls `DestroyMenu()` and it destroys the `HMENU` stored in the `SystemTray` and the stash

fnune:
You guys rock! I wasn't expecting you to just take over the PR :smile:

lemarier (approved):

amrbashir (commented):
@lemarier Why disabling this for windows ?
I think it is fine that the menu is dropped, if they want to set a new menu, they probably need to keep it alive as possible either build it within the same lifetime of event_loop or just static lifetime.

lemarier:
> @lemarier Why disabling this for windows ?
> I think it is fine that the menu is dropped, if they want to set a new menu, they probably need to keep it alive as possible either build it within the same lifetime of event_loop or just static lifetime.

I wonder if we shouldn't handle it internally? Because it's not how macOS and Linux work, but you are right... maybe we can ship it anyway and see how it goes

amrbashir:
> I wonder if we shouldn't handle it internally? Because it's not how macOS and Linux work, but you are right... maybe we can ship it anyway and see how it goes

Well, we can either add drop impl for linux and macOS or remove all drop impls.

lemarier:
> > I wonder if we shouldn't handle it internally? Because it's not how macOS and Linux work, but you are right... maybe we can ship it anyway and see how it goes
> 
> Well, we can either add drop impl for linux and macOS or remove all drop impls.

on windows did the menus are linked to the app? so when closed they are all dropped anyway?

amrbashir:
> on windows did the menus are linked to the app? so when closed they are all dropped anyway?

I think so.

lemarier:
Thanks for your contribution @fnune we do appreciate it a lot!

I'm sure the community will enjoy this new feature.

Have a great day!

PR #133, opened 2021-07-16
[Linux] Replace libappindicator dependency by ksni.

emirror-de:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [X] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

**Other information:**

Because `libappindicator` is [deprecated in Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=895037) applications should switch to `libayatana-appindicator` as stated in the bug report.

To reduce the external dependencies a cleaner way would be to switch to the `ksni` crate that implements the KDE/freedesktop StatusNotifierItem Specification and handles the communication via DBus directly. The specification is used by both, `libappindicator` as well as `libayatana-appindicator` and is at least implemented in KDE, GNOME, Cinnamon and XFCE.


src/platform_impl/linux/system_tray.rs
```diff
@@ -41,43 +32,109 @@ impl SystemTrayBuilder {
...
+  }
+
+  /// Updates the icon.
+  pub fn set_icon(&mut self, icon: &PathBuf) {
+    let (icon_name, icon_theme_path) = Self::split_icon(&icon);
+    self.icon_name = icon_name;
+    self.icon_theme_path = icon_theme_path;
+  }
+
+  fn split_icon(icon: &PathBuf) -> (String, String) {
```
lemarier, line 148: For the icon could we use a Vec<u8> like other platforms? It could be really good if we can remove that `PathBuf`

emirror-de: A method `icon_pixmap` exists that holds the icon in binary representation, so it should be possible. It requires an ARGB32 representation of the image.
https://docs.rs/ksni/0.1.3/ksni/trait.Tray.html#method.icon_pixmap

emirror-de:
**Issues**

1. Currently the set_menu function does not work properly. The menu seems to be updated in the correct way, but somehow the resulting tray menu does render a Submenu where it should just render a StandardItem (see examples/system_tray.rs, tested on Kubuntu 21.04).

**Thoughts on the current state**

1. The menu currently needs to be cloned into the KsniTray struct. Maybe there is a better solution for this without cloning?
2. The accelerators only work when they are set on Submenus. This seems to be a generic desktop environment issue, as KDE uses ALT + KEY for handling shortcuts by default and GNOME seems to not support shortcuts in the tray menu at all. Using libappindicator faces the same behavior.

amrbashir:
> 2. The accelerators only work when they are set on Submenus. This seems to be a generic desktop environment issue, as KDE uses ALT + KEY for handling shortcuts by default and GNOME seems to not support shortcuts in the tray menu at all. Using libappindicator faces the same behavior.

IMO, Accelerators is not really necessary for tray menu and we could just ignore it .

src/platform_impl/linux/menu.rs
```diff
@@ -26,6 +26,25 @@ macro_rules! menuitem {
...
+    ksni::menu::StandardItem {
+      label: $description.into(),
+      ..Default::default()
+    }
+    .into()
+  };
+  ( $description:expr, $modifier:expr, $key:expr ) => {
+    ksni::menu::StandardItem {
+      label: $description.into(),
+      shortcut: vec![vec![$modifier.into(), $key.into()]],
```
amrbashir, line 41: shouldn't this be removed since we dropped accelerator support in tray?

src/platform_impl/linux/menu.rs
```diff
@@ -26,6 +26,25 @@ macro_rules! menuitem {
...
+      ..Default::default()
+    }
+    .into()
+  };
```
amrbashir: This should be removed since we dropped support for accelerators in tray menus.

src/platform_impl/linux/menu.rs
```diff
@@ -176,6 +198,145 @@ impl Menu {
...
+  fn generate_ksnimenu_item(
+    item: &GtkMenuInfo,
+    tx: Sender<(WindowId, WindowRequest)>,
+    window_id: WindowId,
+  ) -> Option<super::system_tray::KsniMenuItem> {
+    match item {
+      GtkMenuInfo {
+        menu_type: GtkMenuType::Custom,
+        sub_menu: None,
+        custom_menu_item: Some(i),
```
amrbashir: I don't like one letter variables
```suggestion
        custom_menu_item: Some(item),
```

src/platform_impl/linux/system_tray.rs
```diff
@@ -8,76 +8,181 @@ use crate::{
...
+      tray.set_menu(tray_menu.clone());
+    });
   }
 }
 
-pub struct SystemTray {
-  app_indicator: AppIndicator,
+/// Type alias for ksni menu.
+pub(crate) type KsniMenu = Vec<ksni::MenuItem<KsniTray>>;
+pub(crate) type KsniMenuItem = ksni::MenuItem<KsniTray>;
```
amrbashir: should probably be `pub(super)`

amrbashir (commented):
Awesome PR, thanks for your efforts. 

Aside from small nit-picks, I don't see any code related to enabling/disabling selecting/unselecting the menu item at runtime.

src/platform_impl/linux/menu.rs
emirror-de: I removed the part from the macro and all invocations using this in c839fd3

src/platform_impl/linux/menu.rs
emirror-de: Variable has been renamed to `custom_item` in e6cff2b. This name has been chosen to prevent any further confusion with the `item` variable that is created earlier.

emirror-de:
> I don't see any code related to enabling/disabling selecting/unselecting the menu item at runtime.

Yes, you are right. I am on it.

src/platform_impl/linux/system_tray.rs
emirror-de: Updated in e67bdb4

emirror-de:
Enabling/disabling and changing the label fixed in d95ad32.

Selecting/Unselecting still needs a bit. In the example nothing changes when calling it during runtime (e.g. when opening the new window). Neither in dev branch, nor in f/ksni.
What is the expected behavior of the `set_selected` function?
Should the item be converted to a Checkmark and automatically get selected? If setting to `false`, should it be converted to a StandardItem instead of unchecking?

amrbashir:
> Selecting/Unselecting still needs a bit. In the example nothing changes when calling it during runtime (e.g. when opening the new window). Neither in dev branch, nor in f/ksni.
> What is the expected behavior of the `set_selected` function?
> Should the item be converted to a Checkmark and automatically get selected? If setting to `false`, should it be converted to a StandardItem instead of unchecking?

Currently in dev branch, `set_selected` will only work if it was initially set to true. I think you should just align it with dev for now until we see a better strategy because menus in linux is a bit different than other platforms (I mean selecting/unselecting a menu item).

emirror-de:
> Currently in dev branch, set_selected will only work if it was initially set to true. I think you should just align it with dev for now until we see a better strategy because menus in linux is a bit different than other platforms (I mean selecting/unselecting a menu item).

Ok, so they are aligned then, except:
A Checkmark item currently does not change between selected and unselected state automatically on click with ksni. It is currently not simple to switch its state internally because there is no link between the `activate` function that handles a click and the item itself. I think this would need a request in ksni repo.
But if that behavior is ok for you, this PR is ready for review.

amrbashir:
Yeah I think we can live with this for now, cc @lemarier @wusyong.

lemarier:
> Yeah I think we can live with this for now, cc @lemarier @wusyong.

I do agree, lets keep it like this for now

wusyong:
I'm also fine with it. Just remember to document the difference of system tray on Linux.

wusyong (approved):

lemarier:
LGTM

src/platform_impl/linux/menu.rs
```diff
@@ -67,12 +77,35 @@ pub struct MenuItemAttributes {
 }
 
 impl MenuItemAttributes {
+  pub fn new(
```
amrbashir, line 80: seems a bit redundant to me, why is this needed?

src/platform_impl/linux/menu.rs
```diff
@@ -176,6 +209,153 @@ impl Menu {
...
+    let mut v = vec![];
+    for item in &self.gtk_items {
+      if let Some(m) = Self::generate_ksnimenu_item(item, tx.clone(), window_id) {
+        v.push(m);
+      }
+    }
+    v
+  }
+
+  fn generate_ksnimenu_item(
```
amrbashir, line 228: can we move these functions behind a trait in the `system_tray.rs`

```rs
trait ToKsniMenu {
}

impl ToKsniMenu for Menu {
}
```

amrbashir (commented):
By just looking at the code I am still not sure how something like this is possible https://github.com/tauri-apps/tao/blob/0d6fcde47a1dd587a7ac2c209d330169496d4c46/examples/system_tray.rs#L102-L105
I know you said selecting/deselecting doesn't work but how about enable/disabling and setting the title.

src/platform_impl/linux/menu.rs
wusyong: They didn't expose to public. It should be okay to stay as is imho.
Maybe we make it `into_ksnimenu` just like `into_gtkmenu`

amrbashir: mmm yeah, makes sense. lets keep it as is and just change the name.

emirror-de:
> how about enable/disabling and setting the title.

this works because the tray gets diffed and rerendered on click. what i did not test yet, now that you address this, is changing title and enabled without interaction with the menu. I can have a look at this tomorrow.
currently the title and enabled get pulled from the gtk_item properties. otherwise it would be required to have a separate list of the menu items.

the new method is required because otherwise the initial set of enabled and checked is not triggered.

amrbashir:
> the new method is required because otherwise the initial set of enabled and checked is not triggered.

Ah I see, you're getting the label and enabled from the actual gtk_item, you can remove `::new()` and pull these states from `MenuItemAtrributes` directly.
so instead of `self.gtk_item.get_label()` just do `self.title`

wusyong:
Sorry I pushed the commit without realising you probablly want to continue the work.
If you still want to work on it, I'll revert the commit.
Or if you are okay with me taking it, I'll try to make editting menuitem work this weekend.

Edit: I reverted 2 commits that changed whole PR since this doesn't respect what you contributed.
Since existing imple can already edit some menuitem fields, I'll explore the rest and see how to achieve it.

wusyong:
Since new version of `libayatana-appindicator` and `libappindicator` is merged, I'll close this PR.
Thanks again for all you contributions! @emirror-de

PR #153, opened 2021-07-21
refactor: refactor `system_tray` impl on windows

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [X] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

- [X] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)

**Other information:**

### DIFF
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -32,7 +32,6 @@ use crate::{
 ///
 /// ## Platform-specific
 ///
-/// **Windows:** The native menu is destroyed when this struct is dropped and can no longer be used.
 pub struct ContextMenu(pub(crate) Menu);
 /// Object that allows you to create a `MenuBar`, menu.
 ///
@@ -40,7 +39,6 @@ pub struct ContextMenu(pub(crate) Menu);
 ///
 /// **macOs:** The menu will show in the **Menu Bar**.
 /// **Linux / Windows:** The menu will be show at the top of the window.
-/// **Windows:** The native menu is destroyed when this struct is dropped and can no longer be used.
 pub struct MenuBar(pub(crate) Menu);
 
 /// A custom menu item.
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -29,10 +29,18 @@ use crate::{
 };
 
 /// Object that allows you to create a `ContextMenu`.
+///
+/// ## Platform-specific
+///
+/// **Windows:** The native menu is destroyed when this stuct is dropped and can no longer be used.
 pub struct ContextMenu(pub(crate) Menu);
 /// Object that allows you to create a `MenuBar`, menu.
 ///
-/// Used by the **Window** menu in Windows and Linux and the **Menu bar** on macOS
+/// ## Platform-specific
+///
+/// **macOs:** The menu will show in the **Menu Bar**.
+/// **Linux / Windows:** The menu will be show at the top of the window.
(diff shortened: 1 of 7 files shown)


src/menu.rs
```diff
@@ -29,10 +29,18 @@ use crate::{
 };
 
 /// Object that allows you to create a `ContextMenu`.
+///
+/// ## Platform-specific
+///
+/// **Windows:** The native menu is destroyed when this stuct is dropped and can no longer be used.
```
wusyong: struct

src/menu.rs
```diff
@@ -29,10 +29,18 @@ use crate::{
...
 pub struct ContextMenu(pub(crate) Menu);
 /// Object that allows you to create a `MenuBar`, menu.
 ///
-/// Used by the **Window** menu in Windows and Linux and the **Menu bar** on macOS
+/// ## Platform-specific
+///
+/// **macOs:** The menu will show in the **Menu Bar**.
+/// **Linux / Windows:** The menu will be show at the top of the window.
+/// **Windows:** The native menu is destroyed when this stuct is dropped and can no longer be used.
```
wusyong: Ditto

src/platform_impl/windows/system_tray.rs
```diff
@@ -118,30 +114,27 @@ impl SystemTrayBuilder {
...
           }
         }),
-        MenuType::ContextMenu,
-        None,
+      };
+      commctrl::SetWindowSubclass(
+        hwnd,
+        Some(tray_subclass_proc),
+        TRAY_SUBCLASS_ID,
+        Box::into_raw(Box::new(traydata)) as _,
```
wusyong, line 136: I just realize there's no `from_raw` for `traydata` and `menu_handler`.
This may cause some memory leaks. Maybe add these to the tray type and release them in drop trait?

src/menu.rs
amrbashir: I forgot we removed the drop impl that destroys the native window, will remove this comment

src/platform_impl/windows/system_tray.rs
amrbashir: Would it be enough to just use `Box::from_raw()` in `WM_DESTROY` which will be triggered when tray type is dropped? don't wanna try to add them to tray type and fight the borrow checker.

wusyong: `Box::into_raw` will return a pointer and the owner of the type is sort of ignored by the borrow checker.
This pointer won't be released on `WM_DESTROY` I suppose. I tried `free` from libc on macos before and that didn't work
The safest way is `from_raw` and make the type drop out of scope.

amrbashir: We are already doing something similar with `Box::from_raw`, https://github.com/tauri-apps/tao/blob/999a2cc524a2ba2f2126f72ec4f88774d61b008c/src/platform_impl/windows/event_loop.rs#L843-L845

wusyong: yes something like this will work.

amrbashir: alright, I just pushed a commit and let all `WM_DESTORY` execute where I call `Box::from_raw`

wusyong (approved):

PR #156, opened 2021-07-22
Revert "refactor(windows): remove window subclassing"

lemarier:
Reverts tauri-apps/tao#140


lemarier:
Reverted because `transparent` wasn't working correctly.

PR #160, opened 2021-07-27
feat(macOS): Allow creation of child Window

lemarier:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**


src/platform/macos.rs
```diff
@@ -301,6 +301,8 @@ impl NativeImage {
 ///  - `with_titlebar_buttons_hidden`
 ///  - `with_fullsize_content_view`
 pub trait WindowBuilderExtMacOS {
+  /// Sets a parent to the window to be created.
+  fn with_parent_window(self, parent: *mut c_void) -> WindowBuilder;
```
amrbashir, line 308: Now with your awesome PR, the 3 major platform has support for child windows so this should be moved inside the normal window attributes now for all platforms.

lemarier: I've made some tests and it may require some works to get there as each platform require different value. Also we need to be able to `Debug` and `Clone` the `WindowBuilder`, so if we want to pass a `Window` it may get complex.

I think it's fair to let it under the platform_impl for now

amrbashir: I think windows HWND is considered a `*mut c_void` but I see your point, gtk will be a pain. lets keep them under platform trait then

PR #206, opened 2021-09-13
Replace winapi with windows crate bindings shared with WRY

wravery:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Documentation
- [x] Build-related changes
- [x] Other, please describe:

By itself, this change should not have much impact on TAO. I've replaced all of the winapi crate references with the windows crate. The generated bindings are in the webview2-com-sys crate, which is important to be able to share types with WRY later in a separate PR. That's already working in my fork of WRY, 

Mostly that's a matter of 1:1 replacements, where the types and functions in winapi matched what the windows crate generated, although there are a few less obvious changes. For one thing, the winapi crate seems to declare enums for most of the `#define` macros in the Windows headers, and the windows crate declares them as `const` values. That means you can't refer to them in a pattern (i.e. match arms), because that just shadows the const with a local variable using the same name. So that's why I had to replace so many patterns with something like `_ if msg == WM_FOO =>`, using a pattern guard expression.

There were a few macros like `GET_X_LPARAM`/`GET_Y_LAPARAM` and `PRIMARYLANGID` that did bit manipulation which winapi reimplemented as public functions, and the windows crate doesn't. I ended up reimplementing those inline where I found them, but I might go back and replace some of those with `const fn`s. On the bright side, while researching those macros I also found a few places where TAO was not quite passing the correct values. For instance, `drag_window` passed a pointer to a `POINTS` struct instead of encoding the x/y coordinates as `i16` values in a `u32`. I think I saw the same bug in at least one other place.

One thing the windows crate does much better is argument coercions, so for example you can pass `false` and it will automatically convert that to a `BOOL(0)` with a conversion trait. It also implements the Default trait for many of the structs, so you don't need to use `std::mem::zeroed` in an unsafe block to initialize them.

The windows crate also wins hands down when dealing with COM, especially if you need to implement them. For instance, TAO benefits from this in the `FileDropHandler` and when dealing with `ITaskbarList2`. I was able to remove a lot of code in `FileDropHandler` dealing with ref-counts and v-tables and replace it with `#[windows::implement(Windows::Win32::System::Com::IDropTarget)]`.

- [ ] Yes. Issue #___
- [x] No

- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

This is related to https://github.com/tauri-apps/wry/pull/133, which was later removed because it only worked for Windows 10 and it required redistributing DLLs along with the app. Both of those have been addressed since then, plus the windows crate now supports the `#[implement]` macro. I think it's worth another try.


wravery:
Marking this ready for review. Now that it's targeting the next branch I don't think there's any reason to hold it back.

lemarier:
Personally, it looks AMAZING. 
I'll not tell you how the Windows API work :P

We'll run different tests/scenarios on different platforms, and we'll wait for the complete review from @amrbashir (he's away for few days) as it's our Windows specialist ;)

Meanwhile, there is a little `fmt` error, would be great if you can fix it.

This is the kind of gift we didn't expect and we do appreciate it a lot.
It'll change a lot of things.

Thanks again from the whole Tauri community.

wravery:
> We'll run different tests/scenarios on different platforms, and we'll wait for the complete review from @amrbashir (he's away for few days) as it's our Windows specialist ;)

Sounds great, LMK if I can help. 👍🏼 

> Meanwhile, there is a little `fmt` error, would be great if you can fix it.

It looks like the errors came from clippy running on macOS. I have a Mac (and Linux machine), so I can take a look, but I wonder if it would make sense to run the clippy checks on Windows and Linux as well. When I ran clippy on Windows I found a couple of other pre-existing errors that I fixed on Windows and the workflow doesn't seem to have caught those.

lemarier:
> but I wonder if it would make sense to run the clippy checks on Windows and Linux as well

Yes totally! I was sure it was checking all platforms already. We can add it in another PR, you don't have to deal with the CI fun. You've already done enough hehehe ;)

Thanks for the clippy fix!!

src/platform/windows.rs
```diff
@@ -79,11 +76,11 @@ impl<T> EventLoopExtWindows for EventLoop<T> {
 /// Additional methods on `Window` that are specific to Windows.
 pub trait WindowExtWindows {
   /// Returns the HINSTANCE of the window
-  fn hinstance(&self) -> *mut libc::c_void;
+  fn hinstance(&self) -> HINSTANCE;
```
amrbashir: Can't we just keep `*mut libc::c_void`, it would make it less annoying for users to use it, since they won't need to depend on `windows` or `webview2-com-sys` crates to use these types.

I think we can re-export these types but I prefer to stick to `libc` types since they are more commonly used I think.

src/platform/windows.rs
```diff
@@ -79,11 +76,11 @@ impl<T> EventLoopExtWindows for EventLoop<T> {
...
   /// Returns the native handle that is used by this window.
   ///
   /// The pointer will become invalid when the native window was destroyed.
-  fn hwnd(&self) -> *mut libc::c_void;
+  fn hwnd(&self) -> HWND;
```
amrbashir: Same here and to not bloat this review, every public function should not return types from `windows` crate.

src/platform_impl/windows/accelerator.rs
```diff
@@ -7,19 +7,20 @@ use std::{
...
+use webview2_com_sys::Windows::Win32::{Foundation::HWND, UI::WindowsAndMessaging::*};
 
 // NOTE:
 // https://docs.microsoft.com/en-us/windows/win32/wsw/thread-safety
 // All handles you obtain from functions in Kernel32 are thread-safe,
 // unless the MSDN Library article for the function explicitly mentions it is not.
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
-struct WindowHandle(HWND);
+struct WindowHandle(isize);
```
amrbashir, line 19: Why not use `HWND` here? It would be more verbose as to what type we are wrapping here.

src/platform_impl/windows/accelerator.rs
```diff
@@ -7,19 +7,20 @@ use std::{
...
 unsafe impl Send for WindowHandle {}
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
-struct AccelHandle(HACCEL);
+struct AccelHandle(isize);
```
amrbashir, line 23: Ditto.

src/platform_impl/windows/event_loop.rs
```diff
@@ -521,103 +541,111 @@ lazy_static! {
...
-        let class_name= util::to_wstring("Tao Thread Event Target");
-
-        let class = winuser::WNDCLASSEXW {
-            cbSize: mem::size_of::<winuser::WNDCLASSEXW>() as UINT,
-            style: 0,
-            lpfnWndProc: Some(winuser::DefWindowProcW),
+        let class = WNDCLASSEXW {
+            cbSize: mem::size_of::<WNDCLASSEXW>() as u32,
+            style: WNDCLASS_STYLES(0),
+            lpfnWndProc: Some(call_default_window_proc),
```
amrbashir: Just curious, why cant we just pass `DefWindowProcW` directly here ?

src/platform_impl/windows/event_loop.rs
```diff
@@ -1063,14 +1095,14 @@ unsafe fn public_window_callback_inner<T: 'static>(
       result = ProcResult::DefSubclassProc;
     }
 
-    winuser::WM_SIZE => {
+    win32wm::WM_SIZE => {
       use crate::event::WindowEvent::Resized;
-      let w = LOWORD(lparam as DWORD) as u32;
-      let h = HIWORD(lparam as DWORD) as u32;
+      let w = lparam.0 as u32 & 0xFFFF;
+      let h = (lparam.0 as u32 & 0xFFFF_0000) >> 16;
```
amrbashir: This is the second time I've seen this low/high word manipulation, can we somehow extract these operations into a utility function ? `to_low_word()` and `get_low_word_val()` maybe ?

src/platform_impl/windows/event_loop.rs
```diff
@@ -1340,16 +1372,16 @@ unsafe fn public_window_callback_inner<T: 'static>(
       result = ProcResult::Value(0);
     }
 
-    winuser::WM_XBUTTONDOWN => {
+    win32wm::WM_XBUTTONDOWN => {
       use crate::event::{ElementState::Pressed, MouseButton::Other, WindowEvent::MouseInput};
-      let xbutton = winuser::GET_XBUTTON_WPARAM(wparam);
+      let xbutton = (wparam.0 >> 16) as u16;
```
amrbashir: Same here, util functions with a descriptive name can help maintain this code.

This applies to more places in the PR that involves low word or some bit shifting.

src/platform_impl/windows/event_loop.rs
```diff
@@ -2135,9 +2162,12 @@ unsafe fn handle_raw_input<T: 'static>(
...
     }
 
-    if util::has_flag(mouse.usButtonFlags, winuser::RI_MOUSE_WHEEL) {
+    if util::has_flag(
+      mouse.Anonymous.Anonymous.usButtonFlags,
+      RI_MOUSE_WHEEL as u16,
+    ) {
       // We must cast to SHORT first, becaues `usButtonData` must be interpreted as signed.
-      let delta = mouse.usButtonData as SHORT as f32 / winuser::WHEEL_DELTA as f32;
+      let delta = mouse.Anonymous.Anonymous.usButtonData as f32 / WHEEL_DELTA as f32;
```
amrbashir, line 2159: Why this `Anonymous` fields exist and will they change in the future ?

src/platform_impl/windows/menu.rs
```diff
@@ -193,11 +197,11 @@ impl Menu {
         format_hotkey(accelerators, &mut anno_title);
       }
 
-      winuser::AppendMenuW(
+      AppendMenuW(
         self.hmenu,
         flags,
         menu_id.0 as _,
-        to_wstring(&anno_title).as_mut_ptr(),
+        PWSTR(to_wstring(&anno_title).as_mut_ptr()),
```
amrbashir: Seems like `to_wstring` is obsolete now since we can just use string directly here.

src/platform_impl/windows/menu.rs
```diff
@@ -434,62 +438,62 @@ fn execute_edit_command(command: EditCommand) {
...
-    inputs[2].u.ki_mut().dwFlags = winuser::KEYEVENTF_KEYUP;
+    inputs[2].r#type = INPUT_KEYBOARD;
+    inputs[2].Anonymous.ki.wVk = key;
+    inputs[2].Anonymous.ki.dwFlags = KEYEVENTF_KEYUP;
 
-    inputs[3].type_ = winuser::INPUT_KEYBOARD;
-    inputs[3].u.ki_mut().wVk = winuser::VK_CONTROL as _;
-    inputs[3].u.ki_mut().dwFlags = winuser::KEYEVENTF_KEYUP;
+    inputs[3].r#type = INPUT_KEYBOARD;
+    inputs[3].Anonymous.ki.wVk = VK_CONTROL as _;
```
amrbashir, line 403: Again not comfortable about these Anonymous fields

src/platform_impl/windows/monitor.rs
```diff
@@ -140,16 +131,16 @@ impl Window {
   }
 }
 
-pub(crate) fn get_monitor_info(hmonitor: HMONITOR) -> Result<winuser::MONITORINFOEXW, io::Error> {
-  let mut monitor_info: winuser::MONITORINFOEXW = unsafe { mem::zeroed() };
-  monitor_info.cbSize = mem::size_of::<winuser::MONITORINFOEXW>() as DWORD;
+pub(crate) fn get_monitor_info(hmonitor: HMONITOR) -> Result<MONITORINFOEXW, io::Error> {
+  let mut monitor_info = MONITORINFOEXW::default();
+  monitor_info.__AnonymousBase_winuser_L13558_C43.cbSize = mem::size_of::<MONITORINFOEXW>() as u32;
```
amrbashir, line 136: Again this Anonymous field.

src/platform_impl/windows/window.rs
```diff
@@ -1168,19 +1145,19 @@ pub fn hit_test(hwnd: HWND, cx: i32, cy: i32) -> LRESULT {
...
+        BOTTOM => LRESULT(HTBOTTOM as i32),
+        TOPLEFT => LRESULT(HTTOPLEFT as i32),
+        TOPRIGHT => LRESULT(HTTOPRIGHT as i32),
+        BOTTOMLEFT => LRESULT(HTBOTTOMLEFT as i32),
+        BOTTOMRIGHT => LRESULT(HTBOTTOMRIGHT as i32),
+        _ => LRESULT(HTNOWHERE as i32),
       }
     } else {
-      HTNOWHERE
+      LRESULT(HTNOWHERE as i32)
```
amrbashir: It might be better to assign the match to a variable, and return `LRESULT` with that variable in the end of the function.

amrbashir (changes requested):
Thanks for the awesome PR and I apologize for the nit-picks.

There is 2 concerns I have:
1. `DWORD`, `UINT` and similar types are replaced with rust types, I think I prefer to stick with win32 types if possible, so it is consistent with original c++ win32 api.
2. There is a lot of `Anonymous` fields on structs in this PR which tbh kinda scary and I don't understand why they exist and how much maintenance it will require.

A totally useless request is can we rename `webview2-com-sys`, seems like `win32-com-sys` might be a better name, as my OCD is triggered 😅 (A window management crate depending on a webview crate) .

src/platform/windows.rs
wravery: I agree, removing the reexported types from the public interface is cleaner. How about the internal Rust type that windows-rs uses instead? In this example, I think that would be `isize`. Or TAO could export its own opaque tuple types for cases like this, so it's still strongly typed.

src/platform_impl/windows/accelerator.rs
wravery: If I recall correctly, it's because `HWND` and several other tuple types don't implement `Send`, but the simple handle values inside can be passed. I should go double check why they don't implement that trait, but I don't think it's anything like implementing `Drop` (which would be bad if the internal value were stolen).

BTW, something like this is what I had in mind for the types in the public interface. Strongly typed, passed around by value, but opaque to consumers.

src/platform_impl/windows/event_loop.rs
wravery: The bindings don't quite match the expected type. One is an `fn` type which takes the parameter types without conversion, and the binding for the specific exported function has `IntoParam` generic trait types for the parameters which turn into a pass through if the type already matches but can also convert more idiomatic Rust types like `&str` automatically when you call it.

Seems like there ought to be a conversion to raw `fn` for this scenario, it may even exist already. If not, I'll suggest it for `windows-rs `.

src/platform_impl/windows/event_loop.rs
wravery: Sure, that and the equivalent to `GET_X_LPARAM`/`GET_Y_LPARAM`, which should use `i16` instead of `u16`.

src/platform_impl/windows/event_loop.rs
wravery: The `Anonymous... ` fields are how `win32metadata` represents unnamed nested structs and unions.

src/platform_impl/windows/monitor.rs
wravery: Yeah, this one is a bit... scarier. It's disambiguating multiple struct/union fields in the same struct. It could require minor maintenance if the Windows SDK header changes, `win32metadata` regenerates from the new headers, and we update `windows-rs` to a version with the new name. But it is frozen as long as not all of those things happen, so it should not happen often. The compiler error in that case should even suggest the new name as something similar.

src/platform/windows.rs
amrbashir: Exporting the inner type, i.e. `isize` is good enough I guess, but since other platforms other than windows with similar methods are exporting `*mut libc::c_void`, we can cast `isize` to it for consistency across the code 
base.

src/platform_impl/windows/accelerator.rs
amrbashir: I see, then wrapping `isize` here is fine.

src/platform_impl/windows/event_loop.rs
amrbashir: Alright, can we move this `call_default_window_proc` function to `util` mod and use it instead of defining it multiple times.

src/platform_impl/windows/event_loop.rs
amrbashir: yeah that would be nice, Also do you have any idea why `windows-rs` doesn't have these functions ?

src/platform_impl/windows/event_loop.rs
amrbashir: I don't like it tbh, but I will accept it since the PR is too good to be held of by this.

src/platform_impl/windows/monitor.rs
amrbashir: I don't know when does win32metdata will generate these types, is it on every compile of `webview2-com-sys` or on every release of it?
I just don't want this to break builds unexpectedly for users of a certain tao version.

wravery: These `Anonymous` members come from the `win32metadata` for the Windows Platform SDK, which is included in the `windows` crate and published with it for each version. The `webview2-com-sys` has a fixed version of the `windows` crate in its dependencies where it gets these types, and it mixes in the additional `win32metadata` for the WebView2 SDK. It should only change with an update to the `webview2-com-sys` which updated its own version of the `windows` crate, which had its own update to the Platform SDK `win32metadata`.

A re-build of TAO or Tauri won't do it, but theoretically it could require a code change when bumping the version of `webview2-com-sys`, if the Platform SDK itself changed and all of the upstream dependencies for `webview2-com-sys` had been updated in that version of `webview2-com-sys`. So at the point where someone was bumping the version of `webview2-com-sys` in TAO itself, they might run into a break they'd need to fix at the same time, but otherwise it should always be stable.

amrbashir: cool, this should be okay then.

src/platform_impl/windows/event_loop.rs
wravery: They might make sense to add by hand to `windows-rs` itself, but the reason they don't exist in the generated bindings is that they're macros rather than functions.

wravery:
Couple of remaining issues that I haven't really addressed, hopefully neither of them is a blocker. 😄 

> 1. `DWORD`, `UINT` and similar types are replaced with rust types, I think I prefer to stick with win32 types if possible, so it is consistent with original c++ win32 api.

The `windows-rs` crate takes a somewhat different approach by preferring Rust types as much as possible with automatic conversions. It keeps the same function/struct names to make it easier to find documentation even though they don't use snake-case, but where there's an equivalent simple type it will use them. It doesn't include any structs for [DWORD](https://microsoft.github.io/windows-docs-rs/doc/bindings/index.html?search=DWORD) or [UINT](https://microsoft.github.io/windows-docs-rs/doc/bindings/index.html?search=UINT), so we'd need to include our own. I think that would just make it a little harder to consume from the Rust side though, and it wouldn't apply to any other platforms.

> A totally useless request is can we rename `webview2-com-sys`, seems like `win32-com-sys` might be a better name, as my OCD is triggered 😅 (A window management crate depending on a webview crate) .

Maybe... but then I think we should probably create a TAO-specific bindings crate for it, under the tauri-apps org. The thing that worries me about that is the types wouldn't be the same as the bindings consumed by WRY, but if all of the types from `webview2-com-sys` are removed from the TAO public interface (e.g. putting back `*mut libc::c_void`), I guess that should be OK. In the worst case, it might define 2 different implementations of any associated functions that are referenced in each crate and bloat the binary a little bit.

amrbashir:
No these are not exactly blockers, it was just a mere preference. 

I will try to do a final review of this PR soon and make sure it is all fine before merge.

src/platform/windows.rs
```diff
@@ -156,14 +154,14 @@ impl WindowExtWindows for Window {
...
-      winuser::PostMessageW(
-        self.hwnd() as _,
-        button as minwindef::UINT,
-        edge as minwindef::WPARAM,
-        &point as *const _ as minwindef::LPARAM,
-      );
+      let w_param = WPARAM(edge as usize);
+
+      let low_word = x as u32 & 0xFFFF;
+      let high_word = (y as u32 & 0xFFFF) << 16;
```
amrbashir: you should use the utils functions you introduced in a previous commit here.

src/platform/windows.rs
```diff
@@ -281,7 +280,7 @@ pub trait MonitorHandleExtWindows {
   fn native_id(&self) -> String;
 
   /// Returns the handle of the monitor - `HMONITOR`.
-  fn hmonitor(&self) -> *mut c_void;
+  fn hmonitor(&self) -> HMONITOR;
```
amrbashir: Should return a `libc::c_void`

src/platform_impl/windows/event_loop.rs
```diff
@@ -1340,16 +1363,16 @@ unsafe fn public_window_callback_inner<T: 'static>(
       result = ProcResult::Value(0);
     }
 
-    winuser::WM_XBUTTONDOWN => {
+    win32wm::WM_XBUTTONDOWN => {
       use crate::event::{ElementState::Pressed, MouseButton::Other, WindowEvent::MouseInput};
-      let xbutton = winuser::GET_XBUTTON_WPARAM(wparam);
+      let xbutton = util::to_hiword(wparam.0 as u32);
```
amrbashir: is this correct or should it be `from_hiword` ?

src/platform_impl/windows/event_loop.rs
```diff
@@ -1360,16 +1383,16 @@ unsafe fn public_window_callback_inner<T: 'static>(
       result = ProcResult::Value(0);
     }
 
-    winuser::WM_XBUTTONUP => {
+    win32wm::WM_XBUTTONUP => {
       use crate::event::{ElementState::Released, MouseButton::Other, WindowEvent::MouseInput};
-      let xbutton = winuser::GET_XBUTTON_WPARAM(wparam);
+      let xbutton = (wparam.0 >> 16) as u16;
```
amrbashir: should be extracted to a util function

src/platform_impl/windows/event_loop.rs
```diff
@@ -1652,14 +1682,14 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
-    winuser::WM_DPICHANGED => {
+    win32wm::WM_DPICHANGED => {
       use crate::event::WindowEvent::ScaleFactorChanged;
 
       // This message actually provides two DPI values - x and y. However MSDN says that
       // "you only need to use either the X-axis or the Y-axis value when scaling your
       // application since they are the same".
       // https://msdn.microsoft.com/en-us/library/windows/desktop/dn312083(v=vs.85).aspx
-      let new_dpi_x = u32::from(LOWORD(wparam as DWORD));
+      let new_dpi_x = util::to_loword(wparam.0 as u32) as u32;
```
amrbashir: again `from_loword`? I am a bit confused about these conversions, that's why I wanted them into a util functions so the name can clear any confusion.

src/platform_impl/windows/event_loop.rs
```diff
@@ -1115,23 +1138,23 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
+          dwFlags: TME_LEAVE,
           hwndTrack: window,
-          dwHoverTime: winuser::HOVER_DEFAULT,
+          dwHoverTime: HOVER_DEFAULT,
         });
       }
 
-      let x = windowsx::GET_X_LPARAM(lparam) as f64;
-      let y = windowsx::GET_Y_LPARAM(lparam) as f64;
+      let x = util::to_x_param(lparam) as f64;
```
amrbashir: shouldn't it be `get_x_param()` ?

src/platform_impl/windows/keycode.rs
```diff
@@ -1,17 +1,13 @@
...
+use webview2_com_sys::Windows::Win32::{
+  System::SystemServices::LANG_KOREAN, UI::KeyboardAndMouseInput::GetKeyboardLayout,
 };
 
 pub fn keycode_to_scancode(code: KeyCode) -> Option<u32> {
   // See `from_scancode` for more info
   let hkl = unsafe { GetKeyboardLayout(0) };
 
-  let primary_lang_id = PRIMARYLANGID(LOWORD(hkl as u32));
+  let primary_lang_id = ((hkl.0 as usize) & 0x3FF) as u32;
```
amrbashir: Should this be moved into a util function too? I am not sure, your call

amrbashir (changes requested):

src/platform_impl/windows/keycode.rs
wravery: Sure, I implemented this as `primary_lang_id`.

src/platform_impl/windows/event_loop.rs
wravery: I just implemented `util::get_xbutton_wparam` to handle these 2 cases

src/platform/windows.rs
wravery: Moved the function body to `platform_impl` (for visibility into `util`) and replaced this pattern with `util::make_x_y_lparam`.

amrbashir (approved):
Thanks again for the awesome work, the PR looks fantastic.

wusyong (approved):
Thanks both of your works! Let's move to wry and make this happen!

PR #229, opened 2021-10-27
feat: Allow more strings to parse to keycode

probablykasper:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Allow more strings to parse to keycode, for example allows `CmdOrCtrl+,` as an alternative to `CmdOrCtrl+Comma`


src/keyboard.rs
```diff
@@ -747,16 +747,15 @@ impl FromStr for KeyCode {
...
-      "NUMPADDIVIDE" => KeyCode::NumpadDivide,
-      "NUMPADSUBSTRACT" => KeyCode::NumpadSubtract,
-      "NUMPADENTER" => KeyCode::NumpadEnter,
+      "NUMADD" | "NUMPADADD" => KeyCode::NumpadAdd,
+      "NUMBACKSPACE" | "NUMPADBACKSPACE" => KeyCode::NumpadBackspace,
+      "NUMCLEAR" | "NUMPADCLEAR" => KeyCode::NumpadClear,
+      "NUMCOMMA" | "NUMPADCOMMA" => KeyCode::NumpadComma,
+      "NUMDIVIDE" | "NUMPADDIVIDE" => KeyCode::NumpadDivide,
+      "NUMSUBSTRACT" | "NUMPADSUBSTRACT" => KeyCode::NumpadSubtract,
+      "NUMENTER" | "NUMPADENTER" => KeyCode::NumpadEnter,
```
amrbashir, line 756: does these even exist ? I can't find `Enter` in the numbers row for example.

amrbashir (changes requested):
<s>Thanks but I don't think I like there is two possible values for keys in `from_str`, it breaks consistency with `to_str`.ex:
```rs
let key = KeyCode::from_str("`"); //parsed as KeyCode::Backquote which is correct
key.to_str(); // will return "BACKQUOTE` instead of what you originally entered which is " `"
```
</s>

Apologies, disregard that, we don't have `KeyCode::to_str()` lol.

I guess there is only the comment I left and we can merge this.

src/keyboard.rs
kasper9n: It should at least be possible for them to exist on keyboards. There are a lot of weird key codes

amrbashir: The meaning of the "NUM" prefix is that these keys exist in the numbers row which some of them don't and even if they do exist, they should have a different corresponding KeyCode. I don't see any benefits from adding this, so it is probably better to remove it.

src/keyboard.rs
```diff
@@ -682,20 +682,20 @@ impl FromStr for KeyCode {
...
+      "NUM0" | "NUMPAD0" => KeyCode::Numpad0,
+      "NUM1" | "NUMPAD1" => KeyCode::Numpad1,
+      "NUM2" | "NUMPAD2" => KeyCode::Numpad2,
+      "NUM3" | "NUMPAD3" => KeyCode::Numpad3,
+      "NUM4" | "NUMPAD4" => KeyCode::Numpad4,
+      "NUM5" | "NUMPAD5" => KeyCode::Numpad5,
+      "NUM6" | "NUMPAD6" => KeyCode::Numpad6,
+      "NUM7" | "NUMPAD7" => KeyCode::Numpad7,
+      "NUM8" | "NUMPAD8" => KeyCode::Numpad8,
+      "NUM9" | "NUMPAD9" => KeyCode::Numpad9,
```
amrbashir, line 694: ~~The `NUM0-9` should correspond to `KeyCode::Digit0-9` not `KeyCode::Numpad0-9`~~

amrbashir (changes requested):

src/keyboard.rs
kasper9n: Ah, thought you were asking if `NumpadEnter` exists in the numpad, my bad.

In Electron the `num` prefix is for the numpad, so that's why I added it like this. Would be confusing if `num0` means `numpad0` in Electron and `0` in tao/tauri. If it's too ambiguous, I'll just remove the `NUM` prefix.

amrbashir: We are not trying to follow electron blindly so there will be differences.

kasper9n: Not suggesting that, just would never expect `num0` to not be the `numpad0` key (which is why I misunderstood you earlier). Wikipedia's numpad page even states that `Num 0` can be used for `numpad0` to remove ambiguity, so I think it would be a poor decision to parse `num0` as `0`. Will go and remove the `NUM` prefixes if you prefer.

amrbashir: You do raise a valid concern though. lets wait on this PR a bit, I will properly review it again when I am back (2-3 days). I need to check our code base and other stuff that I can't do on my phone right now.

amrbashir: Now that I got time to look at this, I think this PR is fine as is, thanks.

amrbashir (approved):

Issue #289, opened 2022-01-19
Off screen rendering


PooyaEimandar:
Hello
Thank you for contributing to this amazing project.
I find out that tao uses gtk and with gtk, we can use the following C codes in order to render the whole window into the pixels.

```
void webview_screenshot(struct webview *w) 
{
    GdkWindow* window;
    GdkPixbuf* buffer;
    gint x, y, width, height;

    window = gtk_widget_get_window(GTK_WIDGET(w->priv.webview));
    gdk_window_get_geometry (window, &x, &y, &width, &height);
    buffer = gdk_pixbuf_get_from_window (window, x, y, width, height);
    
    //test
    printf("w=%d h=%d \r\n", width, height);
    gdk_pixbuf_savev(buffer, "/path/1.jpg", "jpeg", NULL, NULL, NULL);
}
```

This approach can help us to [enable off-screen rendering for wry (issue #391)](https://github.com/tauri-apps/wry/issues/391) feature for wry and then we can use wry in the field of 3D applications & games. (e.g  html user interface in game ).

I could not find any rust function related to `gdk_pixbuf_get_from_window`
am I miss something?

wusyong:
Most gtk-rs crate documents have search alias that you can just type with c functions.
Here's the one I got from gdk:
https://docs.rs/gdk/latest/gdk/prelude/trait.WindowExtManual.html#tymethod.pixbuf

blaind:
Did some experimentation on this, and opened the pull request for easier discussion (with code).

Seems like this would be doable at least for Linux, need to investigate other platforms.

https://github.com/blaind/wry/blob/headless/examples/headless.rs#L111 is an example to use together with the pull req, although the Event is currently a wrong one and only first two frames are being sent.

wusyong:
@blaind Thanks for your attempt. When I was saying this need to do it yourself with tao window creation, I mean everyone can retrieve a raw handle (in this case [gtk_window](https://docs.rs/tao/latest/tao/platform/unix/trait.WindowExtUnix.html#tymethod.gtk_window)). You don't need to implement into the event loop.
If you are only interested in off screen rendering on webview, let's move back to wry. I might have some results for you.

blaind:
[...]

PR #292, opened 2022-01-24
WIP: Off-screen / headless rendering 

blaind:
This is a WIP pull request for discussing approaches to solving #289, hopefully it's okay to open this. This "works" (as in, prints texture data, and even sends them as an events), but the code is definitely all over in wrong places at `tao`.

I will put most of the questions inline into the code. I can modify according to the feedback.

One non-inlined question:
- this example is for gtk / Linux, not sure if other platforms would allow the access to textures as easily. Any experiences on this?

Also, a few notes:
- input forwarding = keyboard, mouse events (from the caller to tao) should also be investigated
- window could maybe be set hidden by https://docs.rs/gtk/latest/gtk/builders/struct.WindowBuilder.html#method.visible

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist
- [ ] When resolving issues, they are referenced in the PR's title (e.g `fix #___, #___`)

### DIFF
--- a/examples/capture_as_texture.rs
+++ b/examples/capture_as_texture.rs
@@ -1,0 +1,39 @@
+// Copyright 2019-2021 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+
+use tao::{
+  event::{Event, WindowEvent},
+  event_loop::{ControlFlow, EventLoop},
+  window::WindowBuilder,
+};
+
+#[allow(clippy::single_match)]
+fn main() {
+  env_logger::init();
+  let event_loop = EventLoop::new();
+
+  let window = WindowBuilder::new()
+    .with_title("A fantastic window!")
+    .with_inner_size(tao::dpi::LogicalSize::new(128.0, 128.0))
+    .build(&event_loop)
+    .unwrap();
+
+  event_loop.run(move |event, _, control_flow| {
+    *control_flow = ControlFlow::Wait;
+
+    match event {
+      Event::WindowEvent {
+        event: WindowEvent::CloseRequested,
+        window_id,
+        ..
+      } if window_id == window.id() => *control_flow = ControlFlow::Exit,
+      Event::MainEventsCleared => {
+        window.request_redraw();
+      }
+      Event::RedrawRequested(_, textures) => {
+        println!("Received texture bytes: {:?}", textures);
+      }
+      _ => (),
+    }
+  });
+}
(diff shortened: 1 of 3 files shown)


examples/capture_as_texture.rs
```diff
@@ -0,0 +1,39 @@
...
+    match event {
+      Event::WindowEvent {
+        event: WindowEvent::CloseRequested,
+        window_id,
+        ..
+      } if window_id == window.id() => *control_flow = ControlFlow::Exit,
+      Event::MainEventsCleared => {
+        window.request_redraw();
+      }
+      Event::RedrawRequested(_, textures) => {
```
blaind, line 33: `RedrawRequested` is probably a wrong place for the texture event. Should it go to `MainEventsCleared`, or maybe a new event type should be created?

src/platform_impl/linux/event_loop.rs
```diff
@@ -669,6 +672,20 @@ impl<T: 'static> EventLoop<T> {
               log::warn!("Failed to send redraw event to event channel: {}", e);
             }
 
+            let buffer: gdk_pixbuf::Pixbuf = window.window().unwrap().pixbuf(0, 0, 6, 4).unwrap();
```
blaind, line 675: What would be the correct place to access the `window.pixbuf`? Definitely not here.

Basically a `gtk::Window` access is required. Need to know the window size as well.

src/platform_impl/linux/event_loop.rs
```diff
@@ -903,7 +922,7 @@ impl<T: 'static> EventLoop<T> {
           }
           _ => match draws.try_recv() {
             Ok(id) => callback(
-              Event::RedrawRequested(RootWindowId(id)),
+              Event::RedrawRequested(RootWindowId(id), Some(textures.try_recv().unwrap())),
```
blaind, line 925: Is it possible to send two events in an iteration? Or is it okay to just try to pull from textures channel, and send even if something is available?

src/platform_impl/linux/event_loop.rs
```diff
@@ -81,6 +81,8 @@ pub struct EventLoop<T: 'static> {
   events: crossbeam_channel::Receiver<Event<'static, T>>,
   /// Draw queue of EventLoop
   draws: crossbeam_channel::Receiver<WindowId>,
+  /// Textures
+  textures: crossbeam_channel::Receiver<Vec<u8>>,
```
blaind, line 85: Could wrap into some generic Texture-struct?

Performance-wise, any challenges for using channels? There's probably only one item in queue anyway.

src/platform_impl/linux/event_loop.rs
blaind: Actually, maybe the texture capture should happen as a response to an event request by the client (library that's using tao), if possible?

PR #298, opened 2022-01-28
macOS: Support set_activation_policy during runtime

probablykasper:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix #___, #___`)

### Other information

Example usage:

```rust
event_loop.run(move |event, event_loop_wt, control_flow| {
  // ...
  #[cfg(target_os = "macos")]
  event_loop_wt.set_activation_policy(ActivationPolicy::Accessory)
}
```

### DIFF
--- a/examples/system_tray.rs
+++ b/examples/system_tray.rs
@@ -10,10 +10,7 @@ fn main() {
   #[cfg(target_os = "linux")]
   use std::path::Path;
   #[cfg(target_os = "macos")]
-  use tao::platform::macos::{
-    ActivationPolicy, CustomMenuItemExtMacOS, EventLoopWindowTargetExtMacOS, NativeImage,
-    SystemTrayBuilderExtMacOS,
-  };
+  use tao::platform::macos::{CustomMenuItemExtMacOS, NativeImage, SystemTrayBuilderExtMacOS};
   use tao::{
     event::{Event, WindowEvent},
     event_loop::{ControlFlow, EventLoop},
@@ -32,13 +29,6 @@ fn main() {
 
   // open new window menu item
   let open_new_window_element = submenu.add_item(MenuItemAttributes::new("Open new window"));
-
-  #[cfg(target_os = "macos")]
-  let mut is_accessory = false;
-  // Accessory activation policy toggle
-  #[cfg(target_os = "macos")]
-  let mut activation_policy_accessory =
-    submenu.add_item(MenuItemAttributes::new("Accessory Activation Policy").with_selected(false));
 
   // set default icon
   #[cfg(target_os = "macos")]
@@ -163,18 +153,6 @@ fn main() {
           || menu_id == focus_all_window.clone().id()
         {
           create_window_or_focus();
-        }
-        #[cfg(target_os = "macos")]
-        if menu_id == activation_policy_accessory.clone().id() {
-          if is_accessory {
-            is_accessory = false;
-            event_loop.set_activation_policy_at_runtime(ActivationPolicy::Regular);
-            activation_policy_accessory.set_selected(false);
-          } else {
-            is_accessory = true;
-            event_loop.set_activation_policy_at_runtime(ActivationPolicy::Accessory);
-            activation_policy_accessory.set_selected(true);
-          }
         }
         // click on `quit` item
         if menu_id == quit_element.clone().id() {
--- a/examples/system_tray.rs
+++ b/examples/system_tray.rs
@@ -34,12 +34,11 @@ fn main() {
   let open_new_window_element = submenu.add_item(MenuItemAttributes::new("Open new window"));
 
   #[cfg(target_os = "macos")]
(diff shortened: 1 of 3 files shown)


amrbashir:
you mentioned that the window will be hidden when switching the activation policy to Accessory, is this still the case ?

probablykasper:
Yes, as far as I can tell that's just how macOS handles it

amrbashir:
Not according to the docs.
> from: https://developer.apple.com/documentation/appkit/nsapplicationactivationpolicy/nsapplicationactivationpolicyaccessory?language=objc
> The application doesn’t appear in the Dock and doesn’t have a menu bar, but it may be activated programmatically or by clicking on one of its windows.

probablykasper:
I can't see anything about the behavior at all in Apple's documentation, so it just doesn't seem to be documented.

[This SO answer](https://stackoverflow.com/a/68057340/6553404) mentions the behavior:
>  2. App windows **went backward/hidden** after App icon in Dock is **disabled** (after setting `NSApplication.ActivationPolicy.accessory`). You need to launch "Mission control" to reveal App windows.

amrbashir:
tbh I don't think this brings much value and the main reason was to hide/show dock icon, so I think we should just close this and look for a proper way to do it by following electron implementation.

probablykasper:
Was the main reason for adding set_activation_policy to show/hide the Dock icon?

I do think this would be very useful for an overlay search bar that wants to toggle a settings window

amrbashir:
> Was the main reason for adding set_activation_policy to show/hide the Dock icon?

yeah and we already have `set_activation_policy()` it just doesn't work at runtime, yours does but not perfect, I guess? that's why I feel it is not worth the effort and people would be confused as to why we have two methods with the same name. 
 
> I do think this would be very useful for an overlay search bar that wants to toggle a settings window

sure, but we should implement dock management the right way.

probablykasper:
> confused as to why we have two methods

Should the original one be removed? Or is there some way to make it only one method?

> sure, but we should implement dock management the right way.

I'm not really talking about toggling the dock icon, I mean that's a use case for switching the activation policy on runtime. The activation policy API isn't there just for toggling the dock icon

amrbashir:
> > confused as to why we have two methods
> 
> Should the original one be removed? Or is there some way to make it only one method?

No, lets keep it and change this one to a different name, `set_activation_policy_at_runtime()` and you should probably document any bugs it has if any, when called during runtime.

> > sure, but we should implement dock management the right way.
> 
> I'm not really talking about toggling the dock icon, I mean that's a use case for switching the activation policy on runtime. The activation policy API isn't there just for toggling the dock icon

Ah okay.

probablykasper:
Renamed it to that, though maybe it would make more sense for the runtime one to be `set_activation_policy`, and the pre-runtime one (which isn't calling `setActivationPolicy`) to be `with_activation_policy`.

No bugs as far as I can tell.

slantview:
👍  Need this asap, thanks for the work!

examples/system_tray.rs
```diff
@@ -30,6 +33,13 @@ fn main() {
   // open new window menu item
   let open_new_window_element = submenu.add_item(MenuItemAttributes::new("Open new window"));
 
+  #[cfg(target_os = "macos")]
+  let mut is_accessory = false;
+  // Accessory activation policy toggle
+  #[cfg(target_os = "macos")]
+  let mut activation_policy_accessory =
+    submenu.add_item(MenuItemAttributes::new("Accessory Activation Policy").with_selected(false));
```
amrbashir: No need to add an example of it since it is platform-specific.

src/platform/macos.rs
```diff
@@ -390,11 +390,11 @@ impl WindowBuilderExtMacOS for WindowBuilder {
 }
 
 pub trait EventLoopExtMacOS {
-  /// Sets the activation policy for the application. It is set to
+  /// Sets the initial activation policy for the application. It is set to
```
amrbashir:
```suggestion
  /// Sets the activation policy for the application. It is set to
```
Lets keep this as is.

amrbashir (approved):

examples/system_tray.rs
kasper9n: Removed it

amrbashir:
Next time please wait for @wusyong to approve and merge macOS specific PRs.

My approval of macOS PRs should not be considered as a green-light for merge.

lucasfernog:
Oops, changed the branch protection rules for `next`. Thanks @amrbashir

probablykasper:
Ah sorry, will keep that in mind

PR #308, opened 2022-02-08
Implement global shortcut on Linux, close #307

### LINKED ISSUE #307: [bug] Global shortcuts are never triggered on Linux
### Describe the bug

Even using API example it's possible to register global shortcut, but it never triggers on linux 
[This line](https://github.com/tauri-apps/tauri/blob/46f2eae8aad7c6a228eaf48480d5603dae6454b4/examples/api/src/components/Shortcuts.svelte#L16) never reached. 

### Reproduction

Compile latest tauri (git commit 46f2eae8aad7c6a228eaf48480d5603dae6454b4) and run api example as per readme. Observe lack of global shortcut trigger - both JS and Rust examples don't work.

### Expected behavior

Shortcut triggered - from JS and Rust

### Platform and versions

```shell
Operating System - Pop!_OS, version 20.04 X64
```

### Stack trace

_No response_

### Additional context

_No response_

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


.changes/linux-shortcuts.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Add global shortcut support on Linux (both x11 and wayland).
```
amrbashir:
```suggestion
Fix global shortcut support on Linux (both x11 and wayland).
```
We did support Linux, it was just broken and was only for x11 so I think "Fix" is better than "Add" here.

src/platform_impl/linux/event_loop.rs
```diff
@@ -600,6 +628,21 @@ impl<T: 'static> EventLoop<T> {
...
+                        tx_clone.send(Event::GlobalShortcutEvent(accelerator.clone().id()))
+                      {
+                        log::warn!(
+                          "Failed to send global shortcut event to event channel: {}",
+                          e
+                        );
+                      }
+                    }
+                  }
+                }
```
amrbashir: GlobalShortcuts are supposed to be "Global" and not tied to a specific window but this code is ran inside a window specific event.

It should be possible to have a tray-only app and still have GlobalShortcuts.

amrbashir (commented):
I have been thinking we should remove globalShortcut module anyway and use `DeviceEvent::Key` and tauri or any tao user can keep their own hashmap of global shortcuts and build their global shortcut manager as they see fit on top of `DeviceEvent::Key`, this has been also proposed by a maintainer at winit. This could also help us when we move back to winit.

`DeviceEvent::Key` is currently not implemented on gtk I think. 

cc @lucasfernog

lucasfernog:
> I have been thinking we should remove globalShortcut module anyway and use `DeviceEvent::Key` and tauri or any tao user can keep their own hashmap of global shortcuts and build their global shortcut manager as they see fit on top of `DeviceEvent::Key`, this has been also proposed by a maintainer at winit. This could also help us when we move back to winit.
> 
> `DeviceEvent::Key` is currently not implemented on gtk I think.
> 
> cc @lucasfernog

I think that can be part of a next iteration, maybe for Tauri 1.1 since it's too big of a change to do it now.

amrbashir:
Sounds good to me.

Now we only need to make this PR global and not specific to a window.

src/platform_impl/linux/event_loop.rs
wusyong: Oh no. There's no such signal for GtkApplication. I think I'll have to find other way or maybe back for GAction. :(

amrbashir: if you want, you can have a hidden window for the non-window events (DeviceEvents, or to catch the shortcuts), kinda like how it is done on Windows, in `src/platform_impl/windows/event_loop.rs` there is a `subclass_event_target_window()` function which has a different event loop logic than the one used for windows created by the user.

I am not sure if this approach would work the same as it works on Windows but this also might be the only way we can use to implement `DeviceEvent::Key` in the future.

wusyong: I tried GAction instead. It's the original plan but the API only accepts strings. I wire through `key_to_raw_key` to convert them. So there might be some keys missing.

src/platform_impl/linux/event_loop.rs
```diff
@@ -563,9 +601,9 @@ impl<T: 'static> EventLoop<T> {
 
             let tx_clone = event_tx.clone();
             let keyboard_handler = Rc::new(move |event_key: EventKey, element_state| {
-              // if we have a modifier lets send it
               let mut mods = keyboard::get_modifiers(event_key.clone());
-              if !mods.is_empty() {
```
amrbashir, line 568: Why was this removed? Just making sure it isn't a typo.

src/platform_impl/linux/global_shortcut.rs
```diff
@@ -1,389 +1,110 @@
...
-        Ok(HotkeyMessage::RegisterHotkeyResult(Err(err))) => Err(err),
-        Err(err) => Err(ShortcutManagerError::InvalidAccelerator(err.to_string())),
-        _ => Err(ShortcutManagerError::InvalidAccelerator(
-          "Unknown error".into(),
-        )),
-      };
+      Ok(RootGlobalShortcut(GlobalShortcut { accelerator }))
+    } else {
+      Err(ShortcutManagerError::InvalidAccelerator(
+        "Unable to register global shortcut".into(),
```
amrbashir: Can we make this error more helpful? i.e. why it failed and maybe include a dump of the failed `Accelerator` struct.

src/platform_impl/linux/global_shortcut.rs
```diff
@@ -1,389 +1,110 @@
...
 
-// required for event but we use dummy window Id
-// so it shouldn't be a problem
-unsafe impl Send for WindowId {}
-unsafe impl Sync for WindowId {}
-// simple enum, no pointer, shouldn't be a problem
-// to use send + sync
-unsafe impl Send for WindowRequest {}
-unsafe impl Sync for WindowRequest {}
+pub enum ShortcutEvent {
```
amrbashir:
```suggestion
pub enum GloablShortcutRequest {
```
to align with `WindowRequest` enum. Also the error messages should reflect this change.

amrbashir (changes requested):

src/platform_impl/linux/event_loop.rs
wusyong: This is replaced by `if event_key.is_modifier()`. So event_key with modifier like `<Shift>T` can get the modifier state in the event.

amrbashir (approved):

PR #328, opened 2022-02-28
refactor: system tray icons

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist

### Other information
This PR aims at unifying the interface around using icons in system trays, it also makes it possible to use other image formats than `.ico` for windows, and since Linux needs a file path for the system tray icon, we have to create it at runtime in a temp dir.


amrbashir:
I haven't tested this on Linux, because my WSL is debian and can't install needed tray dependencies.

amrbashir:
MacOS is not implemented too, so I need help with that.

FabianLars:
Doesn't seem to work correctly on linux (fedora-gnome w/ libappindicator and ubuntu-gnome w/ libappindicator & ayatana).
It doesn't crash but it seems to display some placeholder image instead of the one in the examples folder.
Bildschirmfoto-20220228132025-184x59
Talking about the "3 dots" icon. Here's another placeholder image displayed by a different tray gnome extension:
gnome-shell-screenshot-bodvdx

amrbashir:
Pushed a commit to revert some of the logic used before this PR. I hope it works

FabianLars:
Still doesn't work. And as a hint, if i log the path (`/tmp/.tmpkCaXM1` in this case) it doesn't exist (and nothing with a different name either)

amrbashir:
I missed a call to `set_theme_path`, the tmp file might get deleted before it can be used? and I think it also needs to have `.png` and I am not sure if the icon is written correctly too. Anyways, pushed another commit.

FabianLars:
1) It's missing a create_dir call (getting an error because tao/ doesn't exist in /tmp/
2) (After resolving 1.) the file indeed seems to be invalid. systray still shows "3 dots" icon and opening the file in gnome's default image viewer gives me a "Not a png file" error

amrbashir:
Thanks @FabianLars, I will convert this PR to a draft until I can switch to my Linux system.

amrbashir:
@FabianLars well this should work now, at least the image is written correctly.

I had to use `image` crate to correctly write it though. If there is a better way, please let me know.

FabianLars:
There is this mini crate which covers _only_ this use case: <https://github.com/ideasman42/png-encode-mini-rs>. idk if it works for us too but i found it in some stackoverflow thread were they talked about doing this in plain rust (and the crate is basically the suggested solution).

~~Either way i will test out your commit first in a second.~~ It works.

amrbashir:
I think we should stick with `image` crate, it is more general than `png-encode-mini-rs` and tauri probably will depend on it too if it wants to use `png` icons on Windows.

amrbashir:
Thanks @FabianLars for testing. 

I guess only macOS is left, I'll keep the PR as a draft until someone can help with it.

Cargo.toml
```diff
@@ -118,3 +118,5 @@ gdk-pixbuf = { version = "0.15", features = [ "v2_36_8" ] }
 libayatana-appindicator = { version = "0.2", optional = true }
 libappindicator = { version = "0.7", optional = true }
 x11-dl = "2.18"
+uuid = { version = "0.8", features = [ "v4" ] }
+image = "0.24"
```
FabianLars:
```suggestion
image = { version = "0.24", default-features = false }
```

amrbashir: I don't think we actually need any features for release. `png` is only needed for reading the `.png` image in the examples.

FabianLars: Yep, you're right :+1:

amrbashir:
The PR passed all tests but needs to be tested on macOS first before merge.

wusyong:
Hey sorry for the long wait. I can have some more time after android poc.
The icon isn't showing on mac unfortunately
Screen Shot 2022-03-16 at 2 26 36 PM

And system_tray_no_menu is removed. Do you think it's not relevant anymore?

amrbashir:
> Hey sorry for the long wait. I can have some more time after android poc.
The icon isn't showing on mac unfortunately

I coded the macOS part of this PR without testing because I don't have access to macOS.

> And system_tray_no_menu is removed. Do you think it's not relevant anymore?

I think `system_tray` example is enough and `system_tray_no_menu` doesn't add much tbh.

examples/system_tray.rs
```diff
@@ -173,6 +148,21 @@ fn main() {
...
+    let image = image::load_from_memory(slice)
+      .expect("Failed to parse icon slice")
+      .into_rgba8();
+    let (width, height) = image.dimensions();
+    let rgba = image.into_raw();
+    (rgba, width, height)
+  };
+  tao::system_tray::Icon::from_rgba(icon_rgba, icon_width, icon_height)
+    .expect("Failed to open icon")
+}
```
wusyong: Looks like icon on mac need to use origin `slice` as NSData instead of rgba's
Maybe we should move rgba creation to linux and windows inside.

amrbashir: This would defeat the purpose of this PR, I will convert this PR to draft for now.

awkj:
Hello, I want use svg as mac tray icon( length and height not equal, is a  rectangle, i notice the master branch limit length and height so can't ) ,  this PR can help me solve it ?

amrbashir:
No. you need to find a way to convert the svg into an 32bpp rgba data.

awkj:
> 

thanks answer, use any format is ok, It can allow me use rectangle as icon ?

amrbashir:
Use any format as long as you can extract its RGBA data and it fits the requirements of this method https://docs.rs/tao/latest/tao/window/struct.Icon.html#method.from_rgba

examples/system_tray.rs
amrbashir: pushed a commit to use the original slice.

awkj:
Hi, I've been following this PR for a long time and would like to ask when this PR will be merged into the master branch?

wusyong (dismissed):
Just checked macOS and Linux.
This icon doesn't disappear this time.

PR #329, opened 2022-02-28
feat(macos): allow to disable system tray menu on left click, closes #317

amrbashir:
### What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [x] I have added a convincing reason for adding this feature, if necessary

PR #291, opened 2022-01-22
Fix click events missing whe tray has menu

wusyong:
### What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist
- [ ] When resolving issues, they are referenced in the PR's title (e.g `fix #___, #___`)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary

PR #345, opened 2022-03-16
Fix inconsist behaviour when setting menu on mac

wusyong:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -464,8 +464,10 @@ impl UnownedWindow {
   }
 
   pub fn set_menu(&self, menu: Option<Menu>) {
-    unsafe {
-      util::set_menu_async(*self.ns_window, menu);
+    // TODO if None we should set an empty menu
+    // On windows we can remove it, in macOS we can't
+    if let Some(menu) = menu {
+      menu::initialize(menu);
     }
   }
 
--- a/.changes/set-menu-mac.md
+++ b/.changes/set-menu-mac.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fix inconsist behaviour when setting menu on mac.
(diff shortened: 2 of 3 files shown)


src/platform_impl/macos/window.rs
```diff
@@ -464,8 +464,10 @@ impl UnownedWindow {
   }
 
   pub fn set_menu(&self, menu: Option<Menu>) {
-    unsafe {
-      util::set_menu_async(*self.ns_window, menu);
+    // TODO if None we should set an empty menu
+    // On windows we can remove it, in macOS we can't
```
amrbashir, line 468: Is this still the case with the recent changes in PR #291 ?

amrbashir (approved):

src/platform_impl/macos/window.rs
wusyong: It still is. That PR isn't related.
I think to work this around we could create a menu with app name as an empty menu

amrbashir:
I thought that PR added the ability to manually hide/show the tray menu when the icon is clicked so theoretically if we call `set_menu(None)`  we should have the ability to not show it at all. Like #329 only shows the tray menu on right click without left click.

src/platform_impl/macos/window.rs
amrbashir: I thought that PR added the ability to manually hide/show the tray menu when the icon is clicked so theoretically if we call `set_menu(None)`  we should have the ability to not show it at all. Like #329 only shows the tray menu on right click without left click.

lucasfernog:
I can test this one on Tauri this afternoon.

lucasfernog (approved):
LGTM (I just checked tauri does not expose an api to change the menu after the window has been created so this change is totally ok).

src/platform_impl/macos/window.rs
wusyong: #291 is about some mouse clicking misbehaviour iirc. But yeah I think we should make `set_menu(None)` not show the menu. On macOS, it will show a menu item with only application's name on it. We'll do this in future PR.

PR #347, opened 2022-03-25
refactor: fix and enhance the about menu on Linux

lucasfernog:

- [x] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -187,6 +187,42 @@ impl Default for MenuBar {
   }
 }
 
+/// Application metadata for the [`MenuItem::About`] action.
+#[derive(Debug, Clone)]
+pub struct AboutApplicationData {
+  /// The application name.
+  pub name: String,
+  /// The application name. Currently only used on Linux.
+  pub version: Option<String>,
+  /// The authors of the application. Currently only used on Linux.
+  pub authors: Option<Vec<String>>,
+  /// Application comments. Currently only used on Linux.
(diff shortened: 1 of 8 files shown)


src/menu.rs
```diff
@@ -187,6 +187,25 @@ impl Default for MenuBar {
   }
 }
 
+/// Application metadata for the [`MenuItem::About`] action.
+#[derive(Debug, Clone, Default)]
+pub struct AboutApplicationData {
```
amrbashir: Does `AboutMetadata` sound like a better name?

src/menu.rs
```diff
@@ -187,6 +187,25 @@ impl Default for MenuBar {
   }
 }
 
+/// Application metadata for the [`MenuItem::About`] action.
```
amrbashir, line 190: We should document platform specific notes here instead of each field.

amrbashir (changes requested):

src/menu.rs
lucasfernog: good idea, pushed

src/menu.rs
lucasfernog: yeah, pushed

amrbashir (approved):

PR #352, opened 2022-03-28
feat(windows): add `Event::DecorationsClick`

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/decorations-click-event.md
+++ b/.changes/decorations-click-event.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Added `Event::DecorationsClick` (Windows only).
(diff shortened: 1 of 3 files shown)


lucasfernog:
Tauri needs this event to focus the webview when the user clicks on the window decoration to fix the issue of <select> elements not being moved when the window is moved.

amrbashir (approved):
I am gonna approve but with our recent decision to move back to `winit`, this event will not exist and there is a good chance they won't implement it. However, they are working on a [way](https://github.com/rust-windowing/winit/issues/2120) to allow direct access to the win32 event loop so it is gonna need some refactor down the road.

PR #353, opened 2022-03-28
Set activation policy at runtime

betamos:
There's already a set_activation_policy API, but it's at build time only. This change simply adds an equivalent for runtime use -- using the standard OS X API (just like show/hide application does today). Thanks!

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/add-runtime-activation-policy-mac.md
+++ b/.changes/add-runtime-activation-policy-mac.md
@@ -1,0 +1,5 @@
+---
+"tao": minor
+---
+
+Add `set_activation_policy` as a runtime API.
--- a/.changes/add-runtime-activation-policy-mac.md
+++ b/.changes/add-runtime-activation-policy-mac.md
@@ -2,4 +2,4 @@
 "tao": minor
 ---
 
-Add `set_activation_policy_at_runtime`.
+Add `EventLoopWindowTargetExtMacOS::set_activation_policy_at_runtime`.
--- a/.changes/add-runtime-activation-policy-mac.md
+++ b/.changes/add-runtime-activation-policy-mac.md
@@ -2,4 +2,4 @@
 "tao": minor
 ---
 
-Add `set_activation_policy` as a runtime API.
+Add `set_activation_policy_at_runtime`.
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -17,7 +17,12 @@ use crate::{
 #[cfg(feature = "tray")]
 use crate::system_tray::{SystemTray, SystemTrayBuilder};
 
-use cocoa::{appkit, base::id};
+use cocoa::{appkit::{
+  self, NSApplicationActivationPolicy,
+  NSApplicationActivationPolicyRegular,
+  NSApplicationActivationPolicyAccessory,
+  NSApplicationActivationPolicyProhibited,
+}, base::id};
 
 /// Additional methods on `Window` that are specific to MacOS.
 pub trait WindowExtMacOS {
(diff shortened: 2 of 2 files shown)


.changes/add-runtime-activation-policy-mac.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": minor
+---
+
+Add `set_activation_policy` as a runtime API.
```
amrbashir:
```suggestion
Add `set_activation_policy_at_runtime`.
```

src/platform/macos.rs
```diff
@@ -449,6 +464,8 @@ pub trait EventLoopWindowTargetExtMacOS {
   fn show_application(&self);
   /// Hide the other applications. In most applications this is typically triggered with Command+Option-H.
   fn hide_other_applications(&self);
+  /// Set activation policy at runtime.
+  fn set_activation_policy(&self, activation_policy: ActivationPolicy);
```
amrbashir:
```suggestion
  fn set_activation_policy_at_runtime(&self, activation_policy: ActivationPolicy);
```

src/platform/macos.rs
```diff
@@ -469,6 +486,13 @@ impl<T> EventLoopWindowTargetExtMacOS for EventLoopWindowTarget<T> {
     let app: cocoa::base::id = unsafe { msg_send![cls, sharedApplication] };
     unsafe { msg_send![app, hideOtherApplications: 0] }
   }
+
+  fn set_activation_policy(&self, activation_policy: ActivationPolicy) {
```
amrbashir:
```suggestion
  fn set_activation_policy_at_runtime(&self, activation_policy: ActivationPolicy) {
```

amrbashir (changes requested):
I guess we merged #298 into next only and not to dev.

I am still not sure why winit devs chose to require activation policy to be set before the event loop runs. I am not experienced with macOS so I don't exactly know the side effects of this but here is an interesting comments from winit devs:
https://github.com/tauri-apps/tao/blob/34be12dc7aa4e60d11498af329b269b36de7c398/src/platform_impl/macos/app_state.rs#L470-L472

betamos:
Wow, that PR was almost identical and I had no knowledge about it. It does have examples and docs though, so it's preferable to mine. Let's replicate that change instead.

What's the difference between next and dev branches, and what processes should be followed to get changes across different branches?

FabianLars:
@betamos the next branch is currently inactive so don't worry about it. It was our main branch while dev was frozen because of the audit.

betamos:
On closer look, it appears as the other PR had a bug where the type was wrong - it uses the Rust-defined `ActivationPolicy` enum instead of `NSApplicationActivationPolicy`. I assume it went under the radar due to unsafe-ness in the `msg_send!` macro.

I included Kasper's docstrings in my PR instead. Since it never made it to the `dev` branch, we can just use this PR instead? PTAL.

.changes/add-runtime-activation-policy-mac.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": minor
+---
+
+Add `set_activation_policy_at_runtime`.
```
amrbashir:
```suggestion
Add `EventLoopWindowTargetExtMacOS::set_activation_policy_at_runtime`.
```

amrbashir (changes requested):

amrbashir (dismissed):

wusyong:
Could you format it via `cargo fmt`?
It's good to me too and I'll merge it once it's done.

betamos:
Thanks. Done!

wusyong (approved):

PR #354, opened 2022-03-29
fix(Windows): fix `set_inner_size` setting a bigger size, closes #194

### LINKED ISSUE #194: set_size adds extra pixels on on undecorated windows or windows without a menu.
edit: ahh, i was supposed to open this at tao, urgh... ~~But i guess it kinda makes sense to start here, but maybe that's just me~~

### **Describe the bug**
Please see this discord message and those below: https://discord.com/channels/616186924390023171/625037620996734986/876967731130421248

Bug seems to be windows only.

The sizes are correct until here: https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/windows/util.rs#L101
The function itself gets the correct values (x,y), but outer_y calculated with the value from adjust_window_rect is higher than expected.

With `"decorations": true` it adds 20px to the specified height.
With `"decorations": false` it adds _something_ to both the height and the width.

## To reproduce
~~I wasn't able to reproduce this in tao's examples, didn't try wry (yet?)~~.
edit: reproducable in wry,
edit2: it actually is reproducable in tao, no idea why it didn't work on my first try 🤷 

The tao one should basically be enough, but i could reproduce the behavior only with disabled decorations. On tauri and wry it's enough to not have a menu (but decorations enabled).

### in tauri
1. tauri's api example
2. in fn `main()` comment out `.menu()`
3. Replace Welcome.svelte with this:
```js
<script>
  import { getCurrent, PhysicalSize } from '@tauri-apps/api/window'

  async function resize() {
    const window = getCurrent();
    const size = await window.innerSize();
    await window.setSize(new PhysicalSize(size.width, size.height));
  }
</script>
<p on:mouseenter={resize}>
  Tauri's API capabilities using the ` @tauri-apps/api ` package. It's used as
  the main validation app, serving as the testbed of our development process. In
  the future, this app will be used on Tauri's integration tests.
</p>
```
4. hover a couple of times over the p element, the window height should increase everytime.  

### in wry
for example in custom_titlebar.rs, add the following to the WindowEvent match and focus/unfocus the window a few times.
```rs
WindowEvent::Focused(_) => {
    let w = webviews[&window_id].window();
    w.set_inner_size(w.inner_size());
}
```
p.s. don't get tricked by the visuals if you remove the `Resized` match arm for testing. The window size still increases (visible if you log `outer_x` in tao for example)

### in tao
replace (or create) an example with something like this:
```rs
use tao::{
  event::{Event, WindowEvent},
  window::Window,
};

fn main() {

  let window = Window::new(&event_loop).unwrap();

    match event {
        WindowEvent::Focused(_) => window.set_inner_size(window.inner_size()),
        _ => (),
      },
      _ => (),
    }
  });
}
```

### update
https://github.com/tauri-apps/wry/commit/6eb10d4e10ce86c8403c80fb41ba5e37072dc61e seems to be the first commit where you can observe this behavior. That doesn't make much sense to me tho, as this commit with wry 0.11 and tao 0.4 was part of beta.5. (the issue is gone if you downgrade tauri to beta.5 in a project)

### update 30 or something, i lost count
The first commit that shows this behavior is this one, at least if i didn't trick myself lol: https://github.com/tauri-apps/tao/commit/f35dd03dc6f15d51fb348c6b404c195ba2401339, i wasn't able to figure out which change exactly breaks it, because i wasn't able to fix it by reverting the changes one by one (isolated)...
and i don't get how that commit can be the first one to have this bug, because that commit was part of tauri's beta.5 and it worked fine there. i think i'm loosing my mind 🤔 

### update 24.09
with wravery's PR merged into tauri, the behavior with menus changed (in tauri). Now, without a menu it works as expected, with a menu the window shrinks instead!
Undecorated windows seem to be behave the same way as before.
tbh I don't get why/how that change happened at all, because all of the code related to this seems to be unchanged as far as i can see.

amrbashir:

- [X] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist

### DIFF
--- a/.changes/windows-set_size-overflow.md
+++ b/.changes/windows-set_size-overflow.md
@@ -1,0 +1,5 @@
+---
+"tao": "patch"
+---
+
+On Windows, Fix `Window::set_inner_size` setting a bigger size than requested.
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1710,7 +1710,7 @@ unsafe fn public_window_callback_inner<T: 'static>(
       let new_scale_factor = dpi_to_scale_factor(new_dpi_x);
       let old_scale_factor: f64;
 
-      let allow_resize = {
+      let (allow_resize, is_decorated) = {
         let mut window_state = subclass_input.window_state.lock();
         old_scale_factor = window_state.scale_factor;
         window_state.scale_factor = new_scale_factor;
@@ -1720,11 +1720,21 @@ unsafe fn public_window_callback_inner<T: 'static>(
           return;
         }
 
-        window_state.fullscreen.is_none()
-          && !window_state.window_flags().contains(WindowFlags::MAXIMIZED)
+        let window_flags = window_state.window_flags();
+        (
+          window_state.fullscreen.is_none() && !window_flags.contains(WindowFlags::MAXIMIZED),
+          window_flags.contains(WindowFlags::DECORATIONS),
+        )
       };
 
-      let style = GetWindowLongW(window, GWL_STYLE) as WINDOW_STYLE;
+      let mut style = GetWindowLongW(window, GWL_STYLE) as WINDOW_STYLE;
+      // if the window isn't decorated, remove `WS_SIZEBOX` and `WS_CAPTION` so
+      // `AdjustWindowRect*` functions doesn't account for the hidden caption and borders and
+      // calculates a correct size for the client area.
+      if !is_decorated {
+        style &= !WS_CAPTION;
+        style &= !WS_SIZEBOX;
+      }
       let style_ex = GetWindowLongW(window, GWL_EXSTYLE) as WINDOW_EX_STYLE;
 
       // New size as suggested by Windows.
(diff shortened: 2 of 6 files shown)


amrbashir:
@FabianLars can you test this extensively, and move the window between different monitors with different dpis?

FabianLars:
@amrbashir Yeah moving between monitors with different dpis is not working great. The window size increases each time (no matter if the new monitor it moves to has a bigger or smaller dpi value).
That said the set_inner_size function itself (like i wrote about in 194) works on non-default dpi monitors just fine now.

P.S. It looks like changing the dpi while the window is open increases the window size by the same(?) lengths as dragging it across monitors.

Edit: Oh and if it's of concern, decorated windows are fine.

amrbashir:
Can you see if winit@master has the same effect?

FabianLars:
seems to work fine on winit@master

amrbashir:
@FabianLars can you try this one more time?

FabianLars:
Hmm, looks good so far 👍

PR #359, opened 2022-03-30
docs: document wayland behavior for `Moved` event, closes #259

### LINKED ISSUE #259: Moved event not fired correctly on Wayland
This problem can be split up into 2:
1. The Moved event doesn't get triggered when you move the window, you need to trigger another event to trigger the Moved event _once_ (there is no queue that empties once you trigger some other event, the triggered Moved events match the amount of triggers of the other event)
2. The reported position is always 0, 0

This behavior doesn't seem to be new, as it behaves the same in tauri@beta.8 (== the last release of tao), i also rolled back some commits in tao and didn't see a difference.

System is fedora 35 with gnome
Also: This works fine on X11.

## To reproduce
Imo the easiest way to do that is to add this to the resizable example (or whatever example that already has WindowEvents other than CloseRequested.
```rs
WindowEvent::Moved(position) => {
    dbg!(position);
}
```
So in the resizable example you would end up with this:
```rs
match event {
    WindowEvent::Moved(position) => {
      dbg!(position);
    }
    WindowEvent::KeyboardInput {
      event:
        KeyEvent {
          physical_key: KeyCode::Space,
          ..
        },
      ..
    } => {
      resizable = !resizable;
      println!("Resizable: {}", resizable);
      window.set_resizable(resizable);
    }
    _ => (),
  },
  _ => (),
};
```

Edit: just a small heads up, because i just trolled myself. To properly test this, the cargo run command must be executed from a wayland context like a native terminal. vscode's built-in terminal for example won't work, because it uses XWayland!

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -320,6 +320,10 @@ pub enum WindowEvent<'a> {
   Resized(PhysicalSize<u32>),
 
   /// The position of the window has changed. Contains the window's new position.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux**: Wayland, do not support a global coordinate system, and thus this event will always be (0, 0)
   Moved(PhysicalPosition<i32>),
 
   /// The window has been requested to close.
--- a/src/event.rs
+++ b/src/event.rs
@@ -323,7 +323,7 @@ pub enum WindowEvent<'a> {
   ///
   /// ## Platform-specific
   ///
-  /// - **Linux**: Wayland, do not support a global coordinate system, and thus this event will always be (0, 0)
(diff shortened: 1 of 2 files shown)


FabianLars:
Should we do the same for the inner_position/outer_position getters since they should have the same limitation?

PR #362, opened 2022-04-04
refactor(tray): split gtk and ayatana appindicator features

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

This allows users to have more control on what appindicator package to use, while giving us the possibility to allow both features to be enabled, which would make us resolve the library at runtime.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,10 +27,10 @@ targets = [
 ]
 
 [features]
-default = [ "tray", "ayatana-appindicator" ]
+default = [ "tray", "ayatana-tray" ]
 tray = []
-gtk-appindicator = [ "tray", "libappindicator" ]
-ayatana-appindicator = [ "tray", "libayatana-appindicator" ]
+gtk-tray = [ "tray", "libappindicator" ]
+ayatana-tray = [ "tray", "libayatana-appindicator" ]
 dox = [ "gtk/dox" ]
 
 [dependencies]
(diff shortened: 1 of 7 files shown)


lucasfernog:
I have no idea what I need to install on Ubuntu for this to work lol

Cargo.toml
```diff
@@ -27,9 +27,10 @@ targets = [
 ]
 
 [features]
-default = [ "tray" ]
-tray = [ "libappindicator" ]
-ayatana = [ "libayatana-appindicator" ]
+default = [ "tray", "ayatana-appindicator" ]
```
wusyong: Can we still default to old libappindicator? The majority of most distros are still using it.

Cargo.toml
```diff
@@ -27,9 +27,10 @@ targets = [
...
+tray = []
+gtk-appindicator = [ "tray", "libappindicator" ]
+ayatana-appindicator = [ "tray", "libayatana-appindicator" ]
```
wusyong: Could we shorten the name a bit like `gtk-tray` and `ayatana-try`?
The length is getting a bit longer imo.😅

lucasfernog: makes sense

lucasfernog: pushed @wusyong

wusyong (approved):

PR #371, opened 2022-04-21
On Windows, not transparent and not decorated windows have shadows.

### LINKED ISSUE #157: feat: `with_has_shadow`, `has_shadow` and `set_has_shadow`
When using `decorations: false` the app removes not only the topbar & border, but also the shadow. 

I would like to see the shadow separated into it's own config property. So you could e.g. do:
```
"windows": [
  {
    "shadow": true,
    "decorations": false
  }
]
```
Which would result in a window without the top bar, without borders but *with* the OS shadow.

I tried building this in my app, but this comes with a lot of edge cases. You can't click through the shadow, you have to dynamically change the spacing/shadow when you are full screen, etc.

**Additional context**
This was experienced on Windows, I have not tested behavior on other OS'.

blemelin:

- [X] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist

### Other information

Fixes #157 for Windows. Non-decorated non-transparent windows now have shadows.

### DIFF
--- a/examples/frameless.rs
+++ b/examples/frameless.rs
@@ -1,0 +1,41 @@
+// Copyright 2019-2021 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+
+use tao::{
+  event::{Event, WindowEvent},
+  event_loop::{ControlFlow, EventLoop},
+  window::WindowBuilder,
+};
+
+#[allow(clippy::single_match)]
+fn main() {
+  env_logger::init();
+  let event_loop = EventLoop::new();
+
+  let window = WindowBuilder::new()
+    .with_title("A frameless window!")
+    .with_inner_size(tao::dpi::LogicalSize::new(128.0, 128.0))
+    .with_decorations(false)
+    .with_transparent(false)
+    .build(&event_loop)
+    .unwrap();
+
+  event_loop.run(move |event, _, control_flow| {
+    *control_flow = ControlFlow::Wait;
+    println!("{:?}", event);
+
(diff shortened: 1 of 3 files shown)


blemelin:
Since this changes the behaviour when creating a window on windows, does this require a change file ?

FabianLars:
> Since this changes the behaviour when creating a window on windows, does this require a change file ?

Yes, but don't expect to see this merged anytime soon. I think we don't add new APIs (unless absolutely needed) until after tauri v1 is released (I may be wrong about that since i'm not a wry/tao maintainer).
Also there's https://github.com/tauri-apps/window-shadows because of that.

Furthermore this doesn't really fix #157 since that one is about an API to set it. Your PR just always enables it, which is most likely not what's wanted? (think of overlay-like apps for example)

blemelin:
> Furthermore this doesn't really fix #157 since that one is about an API to set it. Your PR just always enables it, which is most likely not what's wanted? (think of overlay-like apps for example)

Oh right. Understood.

> I think we don't add new APIs (unless absolutely needed).

Then, should I just document how to implement this feature and close this PR ?

amrbashir:
Thanks but the `window-shadows` crate can be used with `tao` directly or `tauri` or `winit`.

I think we will keep the shadows in its separate crate instead of directly in `tao` and `tauri` v2 can depend on it and provide its option.

PR #372, opened 2022-04-22
feat(macos): Fix `with_titlebar_transparent`, add `WindowExtMacOS::set_titlebar_transparent`

probablykasper:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Used @JonasKruckenberg's code from https://github.com/tauri-apps/tauri/issues/2663, and it looks like it works!

Fixes `with_titlebar_transparent` by setting a style mask, and adds `WindowExtMacOS::set_titlebar_transparent`

### DIFF
--- a/.changes/fix-titlebar-transparent.md
+++ b/.changes/fix-titlebar-transparent.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fix macOS `WindowBuilderExtMacOS::with_titlebar_transparent`
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -36,7 +36,7 @@ use cocoa::{
   appkit::{
     self, CGFloat, NSApp, NSApplication, NSApplicationPresentationOptions, NSColor,
     NSRequestUserAttentionType, NSScreen, NSView, NSWindow, NSWindowButton, NSWindowOrderingMode,
-    NSWindowStyleMask, NSWindowTitleVisibility,
+    NSWindowStyleMask,
   },
   base::{id, nil},
   foundation::{NSAutoreleasePool, NSDictionary, NSPoint, NSRect, NSSize, NSUInteger},
@@ -254,7 +254,6 @@ fn create_window(
       if !pl_attrs.has_shadow {
         ns_window.setHasShadow_(NO);
       }
-
       if attrs.position.is_none() {
         ns_window.center();
       }
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -212,7 +212,6 @@ fn create_window(
         );
         ns_window.setStyleMask_(style_mask);
 
-        ns_window.setTitleVisibility_(NSWindowTitleVisibility::NSWindowTitleHidden);
         ns_window.setTitlebarAppearsTransparent_(YES);
       }
       if pl_attrs.title_hidden {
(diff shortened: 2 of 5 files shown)


JonasKruckenberg:
there is also `with_titlebar_transparent` already, which (I think?) was supposed to do this already, but doesn't. Maye we could just fix `with_titlebar_transparent` with this implementation? 🤔

src/platform_impl/macos/window.rs
```diff
@@ -247,6 +249,29 @@ fn create_window(
...
+          NSWindowTitleVisibility::NSWindowTitleHidden
+        } else {
+          NSWindowTitleVisibility::NSWindowTitleVisible
+        });
+        ns_window.setTitlebarAppearsTransparent_(if pl_attrs.titlebar_style_hidden {
+          cocoa::base::YES
+        } else {
+          cocoa::base::NO
+        });
+      }
```
JonasKruckenberg: You can move the if statement to the top of this block and get rid of the inline checks. All the else branches are just the default values.

JonasKruckenberg (changes requested):

src/platform_impl/macos/window.rs
kasper9n: Good call, that's much better!

probablykasper:
> there is also `with_titlebar_transparent` already, which (I think?) was supposed to do this already, but doesn't. Maye we could just fix `with_titlebar_transparent` with this implementation? 🤔

Ah, looks like you're right!

I made it so `with_titlebar_transparent` is what's used, and I also made it not hide the title because there's a separate option for that.

src/platform_impl/macos/window.rs
```diff
@@ -205,6 +205,13 @@ fn create_window(
       ns_window.setAcceptsMouseMovedEvents_(YES);
 
       if pl_attrs.titlebar_transparent {
+        let mut style_mask = ns_window.styleMask();
+        style_mask.set(
+          NSWindowStyleMask::NSFullSizeContentViewWindowMask,
+          true,
+        );
+        ns_window.setStyleMask_(style_mask);
+
```
amrbashir, line 214: There is `WindowBuilderExtMacOS::with_fullsize_content_view` for exactly doing this. Users should just use it.
```suggestion
```

src/platform_impl/macos/window.rs
```diff
@@ -1233,6 +1240,21 @@ impl WindowExtMacOS for UnownedWindow {
...
+  #[inline]
+  fn set_titlebar_transparent(&self, transparent: bool) {
+    unsafe {
+      let id = self.ns_window() as cocoa::base::id;
+      let mut style_mask = id.styleMask();
+      style_mask.set(
+        NSWindowStyleMask::NSFullSizeContentViewWindowMask,
+        transparent,
+      );
+      self.ns_window.setStyleMask_(style_mask);
```
amrbashir, line 1253: This shouldn't be the responsibility of this method, it should probably be split into another method `WindowExtMacOS::set_fullsize_content_view`

.changes/fix-titlebar-transparent.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fix macOS `WindowBuilderExtMacOS::with_titlebar_transparent`
```
amrbashir, line 5: I don't think it was broken in the first place, because it seems like users needed to just pair it with `WindowBuilderExtMacOS::with_fullsize_content_view` in wry/tauri apps.

amrbashir (changes requested):

probablykasper:
Alright, looks like tao already supports what would be needed for making titlebars transparent:
`with_titlebar_transparent` + `with_fullsize_content_view` + `with_title_hidden`

PR #385, opened 2022-05-07
Implement standalone webview ndk

wusyong:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+target = "aarch64-linux-android"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,0 @@
-[build]
-target = "aarch64-linux-android"
--- a/.changes/ndk.md
+++ b/.changes/ndk.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Update ndk to 0.6.
--- a/.changes/ndk.md
+++ b/.changes/ndk.md
@@ -2,4 +2,4 @@
 "tao": patch
 ---
 
-Update ndk to 0.6.
+Add standalone webview ndk port.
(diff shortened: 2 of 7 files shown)


wusyong:
This PR will rewrite whole android stack.
The benefit from it is whole workflow on android will be just like others.
The downside is I need to write many webview code into tao since they are under the event loop.
@lucasfernog @amrbashir Do you have any opinion on this?
This is like we alter to gtk stack on Linux. I change the implementation on Android to be more webview specific.

amrbashir:
The work you've done is remarkable but given our decision to move back to use winit, I don't know how will this fit into that.

wusyong:
Yeah that's my concern too.
But we still need this one way or another. winit's android port is pure NDK.
We can't even define any class in that way, and this is hard requirement to implement IPC.

I think the reasonable choice would be android and linux port in tao will end up remain, and we maintain these ourselves. For the rest platforms, we will re-export from winit.
Does this match what you plan on moving back to winit?

amrbashir:
yeah that sounds about right. 

I just thought we were gonna create a `winit_gtk` crate and have it under `rust-windowing` umbrella since the winit team offered to do so but I guess keeping gtk and android-webview in tao and reexport winit on other platforms is the optimal choice.

wusyong:
Oh right, we can still propose `winit_gtk`. But that might not be the case on Android. Our implementation will be very webview specific.
On maintenance side, it contains around ~2000 LOC. It's fairly maintainable IMHO.

lucasfernog (dismissed):

amrbashir (dismissed):
We can decide later whether we keep gtk and android webview glue code under `tao` or make `winit_gtk` and `tao` would re-export it. For now, this PR is fine. Great work as always.

PR #400, opened 2022-05-20
Update the windows-rs crate to 0.37.0

wravery:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

The `windows-implement` macro which lets you implement COM interfaces with the `windows` crate depends on `const` generic support which was just stabilized in `rustc` 1.61. This effectively raises the MSRV for `tao` on Windows to 1.61.

### DIFF
--- a/.changes/windows-0.37.0.md
+++ b/.changes/windows-0.37.0.md
@@ -2,6 +2,4 @@
 "tao": patch
 ---
 
-Update the `windows` crate to the latest 0.37.0 release.
-
-The `#[implement]` macro in `windows-implement` depends on `const` generic features which were just stabilized in `rustc` version 1.61, so this change also raises the MSRV from 1.56 to 1.61.
+Update the windows-rs crate to the latest 0.37.0 release. This depends on rustc version 1.61 for some `const` generic support which was just stabilized, so on Windows the MSRV is effectively 1.61 now.
--- a/.changes/windows-0.37.0.md
+++ b/.changes/windows-0.37.0.md
@@ -2,4 +2,6 @@
 "tao": patch
 ---
 
-Update the windows-rs crate to the latest 0.37.0 release. This depends on rustc version 1.61 for some `const` generic support which was just stabilized, so on Windows the MSRV is effectively 1.61 now.
+Update the `windows` crate to the latest 0.37.0 release.
+
+The `#[implement]` macro in `windows-implement` depends on `const` generic features which were just stabilized in `rustc` version 1.61, so this change also raises the MSRV from 1.56 to 1.61.
--- a/.changes/windows-0.37.0.md
+++ b/.changes/windows-0.37.0.md
@@ -2,4 +2,8 @@
 "tao": patch
 ---
 
-Update the windows-rs crate to the latest 0.37.0 release. This depends on rustc version 1.61 for some `const` generic support which was just stabilized, so on Windows the MSRV is effectively 1.61 now.
+Update the `windows` crate to the latest 0.37.0 release.
+
+The `#[implement]` macro in `windows-implement` and the `implement` feature in `windows` depend on some `const` generic features which stabilized in `rustc` 1.61. The MSRV on Windows targets is effectively 1.61, but other targets do not require these features.
+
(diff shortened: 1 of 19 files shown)


.changes/windows-0.37.0.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Update the windows-rs crate to the latest 0.37.0 release. This depends on rustc version 1.61 for some `const` generic support which was just stabilized, so on Windows the MSRV is effectively 1.61 now.
```
amrbashir:
```suggestion
Update the windows-rs crate to the latest 0.37.0 release. This depends on rustc version 1.61 for some `const` generic support which was just stabilized, so on Windows the MSRV is effectively 1.61 now.
```
Can you also update thr MSRV in `Cargo.toml`?

amrbashir (changes requested):
As always, we appreciate your help.

.changes/windows-0.37.0.md
wravery: Sure, I figured it didn't need to be raised for other platforms since they won't include the `windows-rs` dependency, but I can see how keeping track of different MSRVs on each platform would be riskier.

I was also thinking that the MSRV in `webview2-com` would take care of enforcement, but that's only applicable on `wry`, and `tao` does use the `windows-implement` macro for some COM interfaces.

FabianLars: i'd rather not increase it for all OSes. There were many Linux users complaining about the raise to 1.57 (in tauri) for quite some time because not everyone uses rustup and they apparently have to wait up to a few months for a new rust version 🙄 Don't reallyyy want to go through that _again_ tbh.

Also, idk if we still plan to go down to 1.56 for all tauri crates to align with debian's build pipeline or whatever?

lucasfernog: I saw some Rust maintainers complaining about that too, because the Debian package is supposed to be a build dependency for other packages, not used for developers since it locks you to an old version. I think not having the same MSRV across all platforms is tricky since most users cannot see the build error as an outdated Rust version error.

wravery:
The change in https://github.com/tauri-apps/tao/pull/400/commits/793c1e6515988f1bbd17aac955a11a26c805cdf3 goes with https://github.com/tauri-apps/wry/pull/592/commits/b63db59bb2361c3bc30fa0a469046cab9826a10e to break a dependency on matching `windows` types in `tao` and `wry`. That should make it possible for `tao` and `wry` to update `windows` and use different versions independently so the PRs and crates.io updates don't need to be staged like this in the future. They can be temporarily out of sync for their `windows` crate versions and still work together.

.changes/windows-0.37.0.md
wravery: Just noticed the new comments, is the consensus that I should revert that change to the MSRV for `tao` and `wry`?

Windows developers are probably more likely to use `rustup` since the OS doesn't come with redistributable toolchain packages (unless you count cross-compilation), so I think it's fair to require them to perform an update. Unfortunately, none of the Windows dependencies in `tao` specify an MSRV, so if they try to build an app with only `tao`, the errors they get will probably be about using unstable features instead of needing to run the upgrade. I'll try to merge a PR to add an MSRV to `windows-implement` since that specific crate is using those features (although you also get that MSRV dependency if you include the `implement` feature in the base `windows` crate), and `tao` depends on that crate so it'll start getting the MSRV errors after that.

The `wry` MSRV should be fine if it's just unspecified. It depends on `webview2-com` (and `windows-implement`, plus the `implement` feature in `windows`), and `webview2-com` specifies the MSRV.

wravery: Here's the PR I started for `windows-implement`: https://github.com/microsoft/windows-rs/pull/1773.

wravery:
I undid the changes in https://github.com/tauri-apps/tao/pull/400/commits/b94e19bbb8792ea047294540d341d9e1a9e86d51 (bumping the MSRV for `tao` on all platforms) and documented it in the change log: https://github.com/tauri-apps/tao/pull/400/commits/ef3f1b7c9970db8d8b22feafe6da10a60b43a12a.

amrbashir (approved):
LGTM, thanks.

wusyong (approved):
Sorry for the late merge. been busy on termite crisis whole weekend 😅

PR #404, opened 2022-05-24
probe libappindicator or libayatana-appindicator at runtime

### LINKED ISSUE #3859: [feat] Auto-detect appindicator library during runtime
### Describe the problem

Right now there's two libraries being used for system tray on Linux: libappindicator and libayatana-appindicator. ayatana is former's fork since the old one is unmaintained. This becomes an issue because different distros might use different one. I would say the ratio is around 50:50.
Decide one during compile time means the other half won't work.

Add a feature or default to auto detect the library. We can use pkg_config crate to probe if the library is exist on the end user device.

### Alternatives considered

_No response_

### Additional context

_No response_

henry40408:
closes tauri-apps/tauri#3859

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist

### Other information

- The pull request removes features `gtk-tray` and `ayatana-tray` and should be considered as a breaking change

### DIFF
--- a/src/platform_impl/linux/system_tray.rs
+++ b/src/platform_impl/linux/system_tray.rs
@@ -9,27 +9,142 @@ use glib::Sender;
 use std::path::PathBuf;
 
 use gtk::{prelude::WidgetExt, AccelGroup};
-#[cfg(feature = "gtk-tray")]
-use libappindicator::{AppIndicator, AppIndicatorStatus};
-#[cfg(feature = "ayatana-tray")]
-use libayatana_appindicator::{AppIndicator, AppIndicatorStatus};
+
+use libappindicator::{
+  AppIndicator as GtkAppIndicator, AppIndicatorStatus as GtkAppIndicatorStatus,
+};
+use libayatana_appindicator::{
+  AppIndicator as AyatanaAppIndicator, AppIndicatorStatus as AyatanaAppIndicatorStatus,
+};
 
 use super::{menu::Menu, window::WindowRequest, WindowId};
 
+enum AppIndicatorStatus {
+  Passive,
+  Active,
+  Attention,
+}
+
+impl From<AppIndicatorStatus> for GtkAppIndicatorStatus {
+  fn from(s: AppIndicatorStatus) -> Self {
+    match s {
+      AppIndicatorStatus::Passive => GtkAppIndicatorStatus::Passive,
+      AppIndicatorStatus::Active => GtkAppIndicatorStatus::Active,
+      AppIndicatorStatus::Attention => GtkAppIndicatorStatus::Attention,
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


src/platform_impl/linux/system_tray.rs
```diff
@@ -9,27 +9,142 @@ use glib::Sender;
...
   tray_menu: Option<Menu>,
-  app_indicator: AppIndicator,
+  app_indicator: Box<dyn AppIndicator>,
+}
+
+fn probe_libayatana() -> bool {
+  if let Err(_) = pkg_config::probe_library("ayatana-appindicator3-0.1") {
+    return false;
+  }
+  return true;
```
amrbashir, line 116:
```suggestion
  pkg_config::probe_library("ayatana-appindicator3-0.1").is_ok()
```

src/platform_impl/linux/system_tray.rs
```diff
@@ -9,27 +9,142 @@ use glib::Sender;
...
+}
+
+fn probe_libindicator() -> bool {
+  if let Err(_) = pkg_config::probe_library("appindicator3") {
+    if let Err(_) = pkg_config::probe_library("appindicator3-0.1") {
+      return false;
+    }
+  }
+  return true;
```
amrbashir, line 125:
```suggestion
pkg_config::probe_library("appindicator3").or(
  pkg_config::probe_library("appindicator3-0.1")
).is_ok()
```

amrbashir (changes requested):
Thank you! A couple of nitpicks and could you add a changefile in `.changes` directory?

Edit: sorry, I didn't notice the that PR is still a draft

FabianLars:
The problem with this approach is that the dev machine needs to have both libs installed. This makes compiling on, for example, fedora completely impossible because libayatana isn't available.

lucasfernog:
My intent when I added both feature flags is that we only probe at runtime if both are enabled, or we add a third feature flag (`runtime-tray` idk).

wusyong:
@FabianLars What about we move this to [libappindicator-rs](https://github.com/tauri-apps/libappindicator-rs)?
The reason it will panic is because both crate check in build script. Maybe we can change how libappindicator-rs checks it.
Probing on runtime don't need the lib actually installed.

FabianLars:
Wouldn't we end up here https://github.com/tauri-apps/libappindicator-rs/pull/28#issuecomment-1063791225 again?
I don't know much about this stuff btw

wusyong:
That one is during compile time. After searching a bit, I found a crate similar to dlopen.
We might use it to bind both old and ayatana one.
I'll close this in favor of https://github.com/tauri-apps/libappindicator-rs/issues/31

PR #415, opened 2022-06-04
feat: support child window on linux, closes #273

### LINKED ISSUE #273: [feat] Support child windows on Linux
Child windows are currently supported on [windows](https://github.com/tauri-apps/tao/blob/f8866a084afe18983c91c67401d8759697578a9c/src/platform/windows.rs#L171) and [mac](https://github.com/tauri-apps/tao/blob/f8866a084afe18983c91c67401d8759697578a9c/src/platform/macos.rs#L308) under the platform specific window APIs, however there does not appear to be an equivalent API for linux.
According to Amr Bashir on discord, child windows are possible on linux, so creating an API in tao for linux would then make it possible to have child windows on all supported platforms.

I am not sure if a common API would be a good idea or not, since child windows apparently behave differently on different platforms, but it may be good to have and to just document the differing behaviour.

keiya01:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -39,7 +39,7 @@ pub trait WindowBuilderExtUnix {
   /// Whether to create the window icon with the taskbar icon or not.
   fn with_skip_taskbar(self, skip: bool) -> WindowBuilder;
   /// Set this window as a transient dialog for `parent`
-  /// <https://gtk-rs.org/gtk3-rs/stable/latest/docs/gdk/struct.Window.html#method.set_transient_for> 
+  /// <https://gtk-rs.org/gtk3-rs/stable/latest/docs/gdk/struct.Window.html#method.set_transient_for>
   fn with_transient_for(self, parent: gtk::ApplicationWindow) -> WindowBuilder;
 }
 
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -12,6 +12,7 @@
 pub use crate::platform_impl::{hit_test, EventLoop as UnixEventLoop};
 use crate::{
   event_loop::EventLoop,
+  platform_impl::Parent,
   window::{Window, WindowBuilder},
 };
 
@@ -37,11 +38,18 @@ impl WindowExtUnix for Window {
 pub trait WindowBuilderExtUnix {
   /// Whether to create the window icon with the taskbar icon or not.
   fn with_skip_taskbar(self, skip: bool) -> WindowBuilder;
+  /// Sets a parent to the window to be created.
+  fn with_parent_window(self, parent: gtk::ApplicationWindow) -> WindowBuilder;
 }
 
 impl WindowBuilderExtUnix for WindowBuilder {
   fn with_skip_taskbar(mut self, skip: bool) -> WindowBuilder {
     self.platform_specific.skip_taskbar = skip;
+    self
+  }
+
+  fn with_parent_window(mut self, parent: gtk::ApplicationWindow) -> WindowBuilder {
+    self.platform_specific.parent = Parent::ChildOf(parent);
     self
   }
 }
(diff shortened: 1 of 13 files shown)


wusyong:
Tested it and looking great! Thanks!
I think at this rate we could bring this method to general window builder method?
This feature isn't relevant to mobile afaik.

vdvman1:
If it is made a general window builder method, it would probably be best to document the behaviour differences for each platform, if there are any.
I know at the very least that "child window" on MacOS and Windows have different behaviour, where the "child window" behaviour on MacOS is more akin to "owned window" in Windows, AFAIK

keiya01:
I fixed these review.
Thank you for reviewing!

vdvman1:
Might be good to use `OwnedBy` on windows rather than `ChildOf`, as I believe that would be more consistent window behaviour with the other platforms (I'm not sure though, that should probably be tested, I haven't/can't test how it behaves on each platform myself)

wusyong:
It's look good to me but I feel like @amrbashir is best to review the difference between `OwnedBy` and `ChildOf`

amrbashir:
as @vdvman1 said, `ChildOf` on Windows doesn't behave like other platforms. `OwnedBy` is the closest to other platforms.

With that said, I don't think it is a good idea to provide a general builder method for all 3 platforms. Instead we should have them under extensions trait so macOS would have `WindowBuilderExtMacOS::with_parent`, and Windows  would have `WindowBuilderExtWindows::with_parent` and `WindowBuilderExtWindows::with_owner` and Linux would have `WindowBuilderExtLinux::with_transient_for`.

keiya01:
I see.
There are some slight difference between these three methods.
If we split these method for each platform, users may not be confused.

FabianLars:
On a somewhat off-topic note, can you do normal pushes instead of force-pushing in the future? Would make it much easier to follow along :)

Imo we should add this to our contributing guidelines too 🤔

keiya01:
> On a somewhat off-topic note, can you do normal pushes instead of force-pushing in the future? Would make it much easier to follow along :)

Oh, sure.
Thank you for your advice 🙇‍♂️

> Imo we should add this to our contributing guidelines too 

yeah even if we separate these method, I think we should keep a context as comment.

wusyong:
@keiya01 So sorry you have to revert. I think this is the part I'll usually miss some context. 🙇‍♂️

@FabianLars I'm okay with both as long as commits are clear. I also will do some force push if commits are all just want to make clippy and fmt happy.

keiya01:
I reverted in [Revert "fix: ganeral window builder"](https://github.com/tauri-apps/tao/pull/415/commits/0fa2d6c2bc43aa110c817ab787b2e81d5332f771).
BTW by bundling commit with force push, I somtimes be eased to revert lol

keiya01:
I will add some description later.

src/platform/unix.rs
```diff
@@ -37,13 +38,20 @@ impl WindowExtUnix for Window {
 pub trait WindowBuilderExtUnix {
   /// Whether to create the window icon with the taskbar icon or not.
   fn with_skip_taskbar(self, skip: bool) -> WindowBuilder;
+  /// Sets a parent to the window to be created.
+  fn with_parent_window(self, parent: gtk::ApplicationWindow) -> WindowBuilder;
```
```suggestion
  /// Set this window as a transient dialog for `parent`
  /// <https://gtk-rs.org/gtk3-rs/stable/latest/docs/gdk/struct.Window.html#method.set_transient_for> 
  fn with_transient_for(self, parent: gtk::ApplicationWindow) -> WindowBuilder;
```

amrbashir (changes requested):

keiya01:
I wonder why does android fail...

wusyong:
Probably github CI being funny again. I'll rerun it tomorrow.

wusyong:
Looks like it's because of https://github.com/rust-windowing/android-ndk-rs/pull/286
Let's ignore this for now. I'll update ndk later.
@amrbashir Does this look good to you?

amrbashir (approved):

PR #422, opened 2022-06-12
feat: add support to deeplink and file association on macOS

meowtec:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This feature is necessary for the file-association function: open files with tauri-based app.
ref: https://github.com/tauri-apps/tauri/pull/4320


src/platform_impl/macos/app_delegate.rs
```diff
@@ -92,3 +101,17 @@ extern "C" fn application_will_terminate(_: &Object, _: Sel, _: id) {
   AppState::exit();
   trace!("Completed `applicationWillTerminate`");
 }
+
+extern "C" fn application_open_file(_: &Object, _: Sel, _: id, file: id) -> i8 {
```
keiya01:
```suggestion
extern "C" fn application_open_file(_: &Object, _: Sel, _: id, file: id) -> bool {
```

src/platform_impl/macos/app_delegate.rs
```diff
@@ -44,6 +49,10 @@ lazy_static! {
       sel!(applicationWillTerminate:),
       application_will_terminate as extern "C" fn(&Object, Sel, id),
     );
+    decl.add_method(
+      sel!(application:openFile:),
+      application_open_file as extern "C" fn(&Object, Sel, id, id) -> i8,
```
keiya01:
```suggestion
      application_open_file as extern "C" fn(&Object, Sel, id, id) -> bool,
```

keiya01 (changes requested):
Thank you for your PR. Please check my review.

src/event.rs
```diff
@@ -100,6 +100,11 @@ pub enum Event<'a, T: 'static> {
     position: PhysicalPosition<f64>,
   },
 
+  /// Emitted when open an external file with the app
+  /// ## Platform-specific
+  /// - **Windows / Android / Linux:** Unsupported.
```
keiya01:
```suggestion
  /// Emitted when open an external file with the app
  ///
  /// ## Platform-specific
  ///
  /// - **Windows / Android / Linux:** Unsupported.
```

src/platform_impl/macos/app_delegate.rs
meowtec: It seems should to be `BOOL`(`cocoa::base::BOOL`)

FabianLars:
Did you try using <https://developer.apple.com/documentation/appkit/nsapplicationdelegate/2887193-application?language=objc> instead?

I'm asking because that's what we need for "deep linking" support and it conflicts with openFile, and i'm not sure yet how we want to handle this conflict. I guess if we can use one interface for both that would be the easiest/best solution.

Anyway, i wanted to try making deep linking work on macos tomorrow, so i can try it out myself too.

meowtec:
> Did you try using https://developer.apple.com/documentation/appkit/nsapplicationdelegate/2887193-application?language=objc instead?
> 
> I'm asking because that's what we need for "deep linking" support and it conflicts with openFile, and i'm not sure yet how we want to handle this conflict. I guess if we can use one interface for both that would be the easiest/best solution.
> 
> Anyway, i wanted to try making deep linking work on macos tomorrow, so i can try it out myself too.

I did not aware of the conflict before, and i will try `openURLs` and find how to make both situations work.

meowtec:
> Did you try using https://developer.apple.com/documentation/appkit/nsapplicationdelegate/2887193-application?language=objc instead?
> 
> I'm asking because that's what we need for "deep linking" support and it conflicts with openFile, and i'm not sure yet how we want to handle this conflict. I guess if we can use one interface for both that would be the easiest/best solution.
> 
> Anyway, i wanted to try making deep linking work on macos tomorrow, so i can try it out myself too.

I have tried `openURLs` and it seems that we can use `openURLs` instead of `openFile` for handling file. When using `openURLs`, a `NSArray<NSURL>` are received, in which files are like `file:///path/to/file`

lucasfernog:
I think `OpenFile` is the correct way to implement this. Let's implement deep link after this so we can see how to handle it.

Electron has both events btw.

lucasfernog:
Actually I saw it gets weird on multiple files :( 

This scares me about openURLs `urls` argument though:
> An array of URLs to open. The list does not include URLs for which your app has a defined document type.

amrbashir:
maybe we should add both events? `OpenFile` and `OpenUrl`? seems like both can be used in different cases and I don't see a harm in adding both

amrbashir:
after a quick look at the documentation, it seems like it needs a return value to indicate the sucess/failure so maybe we need to do this in a similar way to this winit PR https://github.com/rust-windowing/winit/pull/1759

FabianLars:
> maybe we should add both events? OpenFile and OpenUrl? seems like both can be used in different cases and I don't see a harm in adding both

If past-me is any trustworthy, we can't because openfile isn't triggered if an app listens to openurl

amrbashir:
is there an resource confirming that? electron seems to have it

FabianLars:
The docs i linked above mention this: <https://developer.apple.com/documentation/appkit/nsapplicationdelegate/2887193-application?language=objc>

But these are apple's docs so it's likely that i'm misunderstanding them (or that they are wrong lol)

amrbashir:
electron doesn't mention anything about this in their docs so I wonder how electron behaves if you register for both events, anyways, lets follow the official docs.

And in light of this limitation, we need to use the callback style then like this PR https://github.com/rust-windowing/winit/pull/1759 so it can be opt-in for whichever event you want.

lucasfernog:
We could also emit OpenFile if the url is `file://` and OpenUrl for other urls. I couldn't find it in the electron codebase, maybe it's related to chromium.

amrbashir:
maybe that's something that could be offloaded to tauri instead and tao just allows registering both and documents that only one should be registered.

Toldoven:
What is the progress on this issue?

amrbashir:
@wusyong could you check out this PR when you have some time? especially the `OpenFile` event and see if modifying the mutable reference `success` has any effect at all.

wusyong:
@amrbashir How do I test this? Is there an example to run?

amrbashir:
@wusyong 
```rs
use tao::{
  event::{Event, WindowEvent},
  window::WindowBuilder,
};

fn main() {

  let window = Some(
    WindowBuilder::new()
      .with_inner_size(tao::dpi::LogicalSize::new(128.0, 128.0))
      .build(&event_loop)
      .unwrap(),
  );

    println!("{:?}", event);

    match event {
      Event::OpenFile { filename, success } => {
        dbg!(filename);
        *success = false; // reject the drop, change to true to accept it
      }
      Event::WindowEvent {
        event: WindowEvent::Destroyed,
        window_id: _,
        ..
      } => {
      }
      Event::MainEventsCleared => {
        if let Some(w) = &window {
          w.request_redraw();
        }
      }
      _ => (),
    }
  });
}

```

wusyong:
I tired example above but it couldn't really test if events work. If anyone knows how to test it, please let me know.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -96,3 +122,55 @@ extern "C" fn application_will_terminate(_: &Object, _: Sel, _: id) {
...
+  let filename = unsafe { CStr::from_ptr(file.UTF8String()) }
+    .to_string_lossy()
+    .to_string();
+  let mut success = true;
+
+  trace!("Get `application:openFile:` URLs: {:?}", filename);
+  AppState::open_file(filename, &mut success);
+  trace!("Completed `application:openFile:`");
+
+  success as i8
```
TrueCarry: This line throws error since return should be BOOL

lucasfernog:
I'm testing it again and seems like `openURLs` is enough for our use case; The URL scheme is `file://` when opening files, and the deeplink scheme otherwise. Did I forget something @FabianLars ? This PR is so old I always forget something, but from my tests it looks enough and we could drop the openFile[s] logic.

FabianLars:
I honestly forgot most of it too, but yeah, openurls is enough unless we need to respond to the openfile event (tho tbh i've not seen a usecase for that yet) and and could be used to simplify the implementation 🤷 

One thing tho, did you test it with multiple files too? we've had issues with that when we tried to abuse my deep link plugin for that.
Also testing the behavior of different plist configs, like only urls, only files, and having both. Again using the deep link plugin as an example, some users had to add the `file` scheme to the plist for some messed up reason.

So if all of that works i personally don't see much of a reason to keep openFile/s around

lucasfernog:
I did test with multiple files, seems to work fine. I guess we can go with the openURLs approach and wait for community feedback.

amrbashir:
I'd support openFile but default to openURLs because this will and should be upstreamed to winit and could serve as a reference the en

lucasfernog:
@amrbashir when we upstream we could always use [`cfc768f` (#422)](https://github.com/tauri-apps/tao/pull/422/commits/cfc768f14af8dcd148bff2e5de83f4fdfa57f7d9) as a reference. But at least for our own use case it seems like openURLs is enough since it also support file opening.

amrbashir:
yeah I know it is enough but I never want to restrict tao to an API and if possible allow both ways if it is not much hassle. Just to be safe from a future feature request where someone might request it.

amrbashir (approved):

PR #433, opened 2022-06-21
merge build.yml and fmt.yml

henry40408:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: **GitHub workflow configuration**

- [x] No

### Checklist

### Other information

This pull request...

1. merges `build.yml` and `fmt.yml`
2. replaces `hecrj/setup-rust-action@v1` with `actions-rs/cargo@v1`
3. removes the following steps because `serde` feature is not used in `Cargo.toml`:
    - Build with serde enabled
    - Build tests with serde enabled
    - Run tests with serde enabled


amrbashir:
I am fine with merging fmt.yml into build.yml but it needs to be its own job.

.github/workflows/build.yml
```diff
@@ -23,26 +23,27 @@ jobs:
...
-      RUSTFLAGS: "-C debuginfo=0"
-      OPTIONS: ${{ matrix.platform.options }}
-      FEATURES: ${{ format(',{0}', matrix.platform.features ) }}
-      CMD: ${{ matrix.platform.cmd }}
+      RUST_BACKTRACE: 1 # rustc option
+      CARGO_INCREMENTAL: 0 # cargo option
+      RUSTFLAGS: "-C debuginfo=0" # rustc option
+      OPTIONS: ${{ matrix.platform.options }} # as arguments
+      FEATURES: ${{ format(',{0}', matrix.platform.features ) }} # as arguments
+      CMD: ${{ matrix.platform.cmd }} # as arguments
```
amrbashir: The comments are not needed

amrbashir (changes requested):
Can you also rename `build.yml` to `ci.yml`

.github/workflows/build.yml
henry40408: removed in 0b5ddd4

Cargo.toml
```diff
@@ -16,7 +16,7 @@ documentation = "https://docs.rs/tao"
 categories = [ "gui" ]
 
 [package.metadata.docs.rs]
-features = [ "serde", "tray", "dox" ]
```
amrbashir: why was this removed?

.github/workflows/ci.yml
```diff
@@ -81,18 +85,16 @@ jobs:
...
-      - name: Build tests with serde enabled
-        shell: bash
-        run: cargo $CMD test --no-run --verbose --target ${{ matrix.platform.target }} $OPTIONS --features serde,$FEATURES
-
-      - name: Run tests with serde enabled
-        shell: bash
-        if: (
-          !contains(matrix.platform.target, 'android') &&
-          !contains(matrix.platform.target, 'ios'))
-        run: cargo $CMD test --verbose --target ${{ matrix.platform.target }} $OPTIONS --features serde,$FEATURES
```
amrbashir: why was these removed too?

.github/workflows/ci.yml
```diff
@@ -63,6 +64,9 @@ jobs:
         if: contains(matrix.platform.target, 'android')
         run: cargo install cargo-apk
 
+      - name: Download dependencies
+        run: cargo $CMD fetch
+
```
amrbashir: dependencies will be downloaded when building anyways? is there a reason to separate this into its own step?

.github/workflows/ci.yml
amrbashir: looks like we also don't run tests with `tray` feature enabled, maybe we should add that too.

amrbashir (commented):
Sorry, I didn't notice these in the original review

.github/workflows/ci.yml
wusyong: Maybe we can have a test with full features and one without.
It's not practical to run every feature in a single test job.

amrbashir: yeah that sounds better

Cargo.toml
henry40408: add back on commit 71bb9a2

.github/workflows/ci.yml
henry40408: removed on commit 71bb9a2

.github/workflows/ci.yml
henry40408: I added a new `platform` in matrix: `ubuntu-with-all-features`, which includes all features described in `Cargo.toml`

amrbashir: we need to test the tray on all platforms, not just ubuntu, so you should just re-add these 3 steps and add tray to it

amrbashir (approved):
Thank you

PR #449, opened 2022-06-25
feat(linux): Store tray icons in XDG_RUNTIME_DIR

Beanow:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This is preferred over `/tmp`, because this directory (typically `/run/user/{uid}`) is only readable for the current user. While `/tmp` is shared with all users.

Comparing mount options:
```
tmpfs on /tmp type tmpfs (rw,noatime,inode64)
tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,size=3281440k,nr_inodes=820360,mode=700,uid=1000,gid=1000,inode64)
```

Note the `mode=700,uid=1000,gid=1000`.

---

Also added a test case to show fallback behavior, using `/tmp` when the runtime dir is not set.

### DIFF
--- a/src/platform_impl/linux/system_tray.rs
+++ b/src/platform_impl/linux/system_tray.rs
@@ -102,10 +102,24 @@ impl Drop for SystemTray {
 }
 
 fn temp_icon_path() -> std::io::Result<(PathBuf, PathBuf)> {
-  let mut parent_path = std::env::temp_dir();
+  let mut parent_path = dirs_next::runtime_dir().unwrap_or_else(|| std::env::temp_dir());
+
   parent_path.push("tao");
   std::fs::create_dir_all(&parent_path)?;
   let mut icon_path = parent_path.clone();
   icon_path.push(format!("tray-icon-{}.png", uuid::Uuid::new_v4()));
   Ok((parent_path, icon_path))
+}
+
+#[test]
+fn temp_icon_path_prefers_runtime() {
+  let runtime_dir = env!("XDG_RUNTIME_DIR");
+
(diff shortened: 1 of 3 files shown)


amrbashir (approved):
Perfect, thank you

src/platform_impl/linux/system_tray.rs
```diff
@@ -102,10 +102,24 @@ impl Drop for SystemTray {
...
   parent_path.push("tao");
   std::fs::create_dir_all(&parent_path)?;
   let mut icon_path = parent_path.clone();
   icon_path.push(format!("tray-icon-{}.png", uuid::Uuid::new_v4()));
   Ok((parent_path, icon_path))
 }
+
+#[test]
+fn temp_icon_path_prefers_runtime() {
+  let runtime_dir = env!("XDG_RUNTIME_DIR");
```
lucasfernog, line 116: this test will fail if you don't have xdg dirs set up FYI

amrbashir: Nice catch, will push a commit to use `option_env!` asap

Beanow: Ha, yes you're right. Though the test itself shows it would at runtime fall back on `/tmp` for those users :sweat_smile:
Should it have been an early return?

PR #451, opened 2022-06-25
feat(macOS): create native image from file path

ahkohd:
Add support to create `NativeImage` from a given file path.

It enables devs to use an image from a file path as native image for their menu items.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/examples/system_tray.rs
+++ b/examples/system_tray.rs
@@ -7,11 +7,11 @@
 #[cfg(any(feature = "tray", all(target_os = "linux", feature = "ayatana")))]
 fn main() {
   #[cfg(target_os = "macos")]
+  use tao::platform::macos::{CustomMenuItemExtMacOS, NativeImage, SystemTrayBuilderExtMacOS};
   use tao::{
     event::Event,
     event_loop::{ControlFlow, EventLoop},
     menu::{ContextMenu as Menu, MenuItemAttributes, MenuType},
-    platform::macos::{CustomMenuItemExtMacOS, NativeImage, SystemTrayBuilderExtMacOS},
     system_tray::SystemTrayBuilder,
   };
 
--- a/examples/system_tray.rs
+++ b/examples/system_tray.rs
@@ -7,19 +7,16 @@
 #[cfg(any(feature = "tray", all(target_os = "linux", feature = "ayatana")))]
 fn main() {
   #[cfg(target_os = "macos")]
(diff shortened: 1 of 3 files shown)


examples/system_tray.rs
```diff
@@ -7,27 +7,35 @@
...
 fn main() {
   #[cfg(target_os = "macos")]
-  use tao::platform::macos::{CustomMenuItemExtMacOS, NativeImage, SystemTrayBuilderExtMacOS};
   use tao::{
     event::Event,
     event_loop::{ControlFlow, EventLoop},
     menu::{ContextMenu as Menu, MenuItemAttributes, MenuType},
+    platform::macos::{CustomMenuItemExtMacOS, NativeImage, SystemTrayBuilderExtMacOS},
     system_tray::SystemTrayBuilder,
   };
```
Beanow: This breaks other platforms, because instead of:
only using `tao::platform::macos::*` on macos.
Now it:
only uses `tao::*` on macos.

And all of Tao is not imported.
Just revert to what it was and it works fine (tried on linux) :]

ahkohd: @Beanow thanks, I have fixed it,  please refer to commit 4a53076.

Beanow: Resolves compile issues for me, though I don't have a mac environment ready. So I'll trigger the CI tests.

amrbashir (commented):
`NativeImage` is supposed to be only "pre-defined-system-icon" so `FromFile` variant shouldn't be there and this overall seems like what we would do for `CustomMenuItem::set_icon()` which won't be specific to macOS and will be implemented for all platforms.

ahkohd:
> `NativeImage` is supposed to be only "pre-defined-system-icon" so `FromFile` variant shouldn't be there and this overall seems like what we would do for `CustomMenuItem::set_icon()` which won't be specific to macOS and will be implemented for all platforms.

I see, I was following electron's spec,  https://www.electronjs.org/docs/latest/api/native-image
According to electron's doc, NativeImage can be created from PNG, and JPG and it can be used to create tray, dock, and application icons using PNG or JPG files.

Should I go ahead to implement `CustomMenuItem::set_icon()` and revert my changes to `NativeImage`?

amrbashir:
yeah that would be better.

ahkohd:
Closed in favour of  PR #459

PR #468, opened 2022-07-08
feat: add theme feature on Linux, closes #387

### LINKED ISSUE #387: [feat] Read dark/light theme on MacOS and Linux
**Describe the problem**

Currently, the API can only read out the current system theme on Windows, but not on MacOS.

The following code below will result in `light` even though I'm in dark mode.

```js
import { appWindow } from '@tauri-apps/api/window';

appWindow.theme().then((theme) => {
  console.log(theme); // returns 'light' even though I'm in dark mode
});
```

This should work on MacOS as well.

**Alternatives considered**

None.

**Additional context**

N/A

keiya01:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Currently GTK does not support theme changed event...

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -469,7 +469,9 @@ pub enum WindowEvent<'a> {
   /// Applications might wish to react to this to change the theme of the content of the window
   /// when the system changes the window theme.
   ///
-  /// At the moment this is only supported on Windows.
+  /// ## Platform-specific
+  ///
+  /// - **Linux / Android / iOS:** Unsupported
   ThemeChanged(Theme),
 
   /// The window decorations has been clicked.
--- a/src/platform_impl/linux/window.rs
+++ b/src/platform_impl/linux/window.rs
@@ -9,7 +9,7 @@ use std::{
 };
 
 use gdk::{WindowEdge, WindowState};
-use gtk::{prelude::*, AccelGroup, Orientation};
+use gtk::{prelude::*, traits::SettingsExt, AccelGroup, Orientation, Settings};
 use raw_window_handle::{RawWindowHandle, XlibHandle};
 
 use crate::{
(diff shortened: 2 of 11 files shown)


src/event.rs
```diff
@@ -469,7 +469,9 @@ pub enum WindowEvent<'a> {
   /// Applications might wish to react to this to change the theme of the content of the window
   /// when the system changes the window theme.
   ///
-  /// At the moment this is only supported on Windows.
+  /// ## Platform-specific
+  ///
+  /// - **Linux / Android / iOS:** Unsupported
```
wusyong, line 474: I assume it's Windows instead of Linux

keiya01: I think windows is already supported.

wusyong (approved):

src/platform_impl/linux/window.rs
```diff
@@ -647,6 +669,20 @@ impl Window {
...
     }
   }
+
+  pub fn theme(&self) -> Theme {
+    if let Some(settings) = Settings::default() {
+      let theme_name = settings.gtk_theme_name().map(|s| s.as_str().to_owned());
+      if let Some(theme) = theme_name {
+        // Currently GTK doesn't provide feature for detect theme, so we need to check theme manually.
+        // ref: https://github.com/WebKit/WebKit/blob/e44ffaa0d999a9807f76f1805943eea204cfdfbc/Source/WebKit/UIProcess/API/gtk/PageClientImpl.cpp#L587
+        if theme.ends_with("-dark") || theme.ends_with("-Dark") {
```
lucasfernog, line 679: I use Arc-Darker theme :( really sucks having to detect the theme this way @keiya01 @wusyong

amrbashir: probably should be
```suggestion
        if theme.contains("-dark") || theme.contains("-Dark") {
```

keiya01: Darker ... I see. I will fix this issue.

keiya01: @lucasfernog 
I created the PR. Could you check it when you have time?
https://github.com/tauri-apps/tao/pull/511

Issue #484, opened 2022-07-17
[bug] Event can't be cloned


ghostnumber7:
There is an issue with lifetimes when you try to `Clone::clone()` an event inside `event_loop.run()`. If you take any example and add `let _clone = event.clone();` you will see something like the following:
```bash
cargo run --example custom_events
    Blocking waiting for file lock on build directory
   Compiling tao v0.12.1 (/Users/mgodoy/mko/rust/untitled/tao)
warning: unused import: `Icon`
  --> src/platform/macos.rs:14:12
   |
14 |   window::{Icon, Window, WindowBuilder},
   |            ^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `tao` (lib) generated 1 warning
error[E0495]: cannot infer an appropriate lifetime due to conflicting requirements
  --> examples/custom_events.rs:41:24
   |
41 |     let _clone = event.clone();
   |                        ^^^^^
   |
note: first, the lifetime cannot outlive the anonymous lifetime #1 defined here...
  --> examples/custom_events.rs:38:18
   |
38 |     event_loop.run(move |event, _, control_flow| {
   |  __________________^
39 | |     *control_flow = ControlFlow::Wait;
40 | |
41 | |     let _clone = event.clone();
...  |
50 | |     }
51 | |   });
   | |___^
note: ...so that the types are compatible
  --> examples/custom_events.rs:41:24
   |
41 |     let _clone = event.clone();
   |                        ^^^^^
   = note: expected `&Event<'_, CustomEvent>`
              found `&Event<'_, CustomEvent>`
   = note: but, the lifetime must be valid for the static lifetime...
note: ...so that the types are compatible
  --> examples/custom_events.rs:41:24
   |
41 |     let _clone = event.clone();
   |                        ^^^^^
   = note: expected `<Event<'_, CustomEvent> as Clone>`
              found `<Event<'static, CustomEvent> as Clone>`

For more information about this error, try `rustc --explain E0495`.
error: could not compile `tao` due to previous error
```

Due to this, there is no way to pass the event to anything outside the run closure, or any way to run something like `let _non_user_events = event.clone().map_nonuser_event::<()>();` (You can't run it on event directly as it does not implement `Copy`)

The reason behind the issue seams to be that `Clone` is only implemented for `'static` lifetime and the `event_loop.run()` closure requirement of being a `move`.

I got it working properly by changing 2 lines on `src/event.rs`:
```diff
[...]

PR #485, opened 2022-07-17
Fix: Event can't be cloned #484

ghostnumber7:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Bug description in [issue 484](https://github.com/tauri-apps/tao/issues/484)

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -164,7 +164,7 @@ pub enum Event<'a, T: 'static> {
   LoopDestroyed,
 }
 
-impl<T: Clone> Clone for Event<'static, T> {
+impl<T: Clone> Clone for Event<'_, T> {
   fn clone(&self) -> Self {
     use self::Event::*;
     match self {
@@ -482,7 +482,7 @@ pub enum WindowEvent<'a> {
   DecorationsClick,
 }
 
-impl Clone for WindowEvent<'static> {
+impl Clone for WindowEvent<'_> {
   fn clone(&self) -> Self {
     use self::WindowEvent::*;
     return match self {


amrbashir (approved):
does this seem okay with you @wusyong

wusyong:
Fine by me, but need to aware that winit is still `'static`

amrbashir:
right!
@ghostnumber7 do you really need to clone? can't you just use `&event` or `&mut event`? because there is some event variants that could be modified (`WindowEvent::ScaleFactorChanged` iirc)so you'd better take a reference

ghostnumber7:
> right! @ghostnumber7 do you really need to clone? can't you just use `&event` or `&mut event`? because there is some event variants that could be modified (`WindowEvent::ScaleFactorChanged` iirc)so you'd better take a reference

I was trying to do  `event.map_nonuser_event::<SomeOtherUserEventThing>()` to be able to send non user events somewhere else to be handled by a plugin or alike. As in, I don't care to handle menu events for example, but there is something else that needs to, and don't want to map every single possible event one by one. The clone thing is an example to reproduce easily, but the real problem is that I can do nothing to send the event to some other place, except one by one mapping. Felt like that was the whole idea behind `map_nonuser_event` method and could need a fix :)

amrbashir:
does this work for you?

```rs
fn handle_events<T>(event: &mut Event<T>) {};
```
then in your eventloop run callback
```rs
let mut event = event.map_nonuser_event::<SomeOtherUserEventThing>();

// handle the events somewhere elese
handle_events(&mut event);

// rest of the run callback
match event {
}

```

amrbashir:
also have you looked at `Event::to_static` then you can call `clone` on it.

ghostnumber7:
> ```rust
> map_nonuser_event
> ```

If I try that I get an error 
```rust
event_loop.run(move |window_event, event_loop, control_flow| {

        let mut event = window_event.to_static().unwrap().map_nonuser_event::<()>();
        // same happens with let mut event = window_event.map_nonuser_event::<()>();
        println!("moved {:?}", event);

        match window_event {
            Event::NewEvents(StartCause::Init) => { ...
```

```shell
error[E0382]: use of moved value
   --> packages/desktop/src/lib.rs:215:24
    |
115 |     event_loop.run(move |window_event, event_loop, control_flow| {
    |                          ------------ move occurs because `window_event` has type `wry::application::event::Event<'_, UserWindowEvent>`, which does not implement the `Copy` trait
...
118 |         let mut event = window_event.to_static().unwrap().map_nonuser_event::<()>();
    |                                      ----------- `window_event` moved due to this method call
...
215 |                 event, window_id, ..
    |                        ^^^^^^^^^ value used here after move

```

I think the issue is that inside the closure for `event_loop.run` it can't be static at all if I want to use it twice.

ghostnumber7:
Note that "rest of the code" uses the original event, not the casted one, so it explodes when it gets moved

amrbashir:
if you want to be able to handle the original event then you need to either:

- handle it before calling `map_nonuser_event`
```rs
	  event_loop.run(move |event, event_loop, control_flow| {
	          
	          match event {
	              ...
	          }
	  
	          let mut nonuser_event = event.map_nonuser_event::<()>();
```
- or turn it into static then clone it and then call `map_nonuser_event`
```rs
	    event_loop.run(move |event, event_loop, control_flow| {
		            
		            let mut event = event.to_static().unwrap();
		            let nonuser_events = event.clone().map_nonuser_event::<()>();
		            
		            match event {
		              ...
		    
```

ghostnumber7:
> if you want to be able to handle the original event then you need to either:
> 
> * handle it before calling `map_nonuser_event`
> 
> ```rust
> 	  event_loop.run(move |event, event_loop, control_flow| {
> 	          *control_flow = ControlFlow::Wait;
> 	          
> 	          match event {
> 	              ...
> 	          }
> 	  
> 	          let mut nonuser_event = event.map_nonuser_event::<()>();
> ```
> 
> * or turn it into static then clone it and then call `map_nonuser_event`
> 
> ```rust
> 	    event_loop.run(move |event, event_loop, control_flow| {
> 		            *control_flow = ControlFlow::Wait;
> 		            
> 		            let mut event = event.to_static().unwrap();
> 		            let nonuser_events = event.clone().map_nonuser_event::<()>();
> 		            
> 		            match event {
> 		              ...
> 		    
> ```

yep, that works.

amrbashir:
We are trying to minimize the differences between tao and winit so the transition back to winit can be easier.

On that note, I am gonna close this and the relevant issue.

ghostnumber7:
> We are trying to minimize the differences between tao and winit so the transition back to winit can be easier.
> 
> On that note, I am gonna close this and the relevant issue.

Thanks for the help :)

PR #487, opened 2022-07-19
fix(windows): subscribe to taskbar restart event, closes #476

### LINKED ISSUE #476: [bug] System tray icon on taskbar disappears when explorer restarts
### Describe the bug

On Windows when you create an app with Tauri with a system tray icon and run it, the icon is visible on the system tray in the taskbar. When you restart explorer, the system tray icon disappears. This is problematic when your app runs in the background with no visible window and where the window can only be made visible via the system tray menu. So when explorer restarts, the app is still running but you cannot access it anymore.

The problem is probably caused by Tauri not handling the TaskbarCreated event to register the icon again for the new task bar.

More info about the event can be found here:

https://docs.microsoft.com/en-us/windows/win32/shell/taskbar#taskbar-creation-notification

Another problem related to restarting explorer is when you have set skip_taskbar to true and then restart explorer, suddenly the Tauri app is visible on the taskbar.

On Windows, some applications restart explorer when an update is installed in the background. Therefore it is necessary that Tauri supports this and re-registers the system tray icon when this happens and also handles the skip_taskbar properly.

### Reproduction

Create a simple Tauri app with one window with a system tray icon and with a system tray menu where you can close the app and set skip_taskbar to true on the window.

Run the app and notice the system tray icon.

Now restart explorer.exe. You can do this via Task Manager. Run Task Manager, search for "Windows Explorer", select it, and press the restart button at the bottom right.

Notice that Windows Explorer restarts and there is no system tray icon of the Tauri app anymore. Also notice that the app is visible again on the task bar.

### Expected behavior

When you have a Tauri app with a system tray icon and where skip_taskbar is set to true, restarting explorer.exe should still show the system tray icon and also it should not make the app visible on the taskbar.

### Platform and versions

```shell
Environment
  › OS: Windows 10.0.19044 X64
  › Webview2: 103.0.1264.49
  › MSVC: 
      - Visual Studio Build Tools 2022
  › Node.js: 16.14.2
  › npm: 8.5.0
  › pnpm: Not installed!
  › yarn: 1.19.1
  › rustup: 1.24.3
  › rustc: 1.62.0
  › cargo: 1.62.0
  › Rust toolchain: stable-x86_64-pc-windows-msvc 

Packages
  › @tauri-apps/cli [NPM]: 1.0.4
  › @tauri-apps/api [NPM]: 1.0.2
  › tauri [RUST]: 1.0.4,
  › tauri-build [RUST]: 1.0.4,
  › tao [RUST]: 0.12.1,
  › wry [RUST]: 0.19.0,

App
  › build-type: bundle
  › CSP: default-src 'self'
  › distDir: ../dist
  › devPath: http://localhost:5173/
  › framework: Vue.js

App directory structure
  ├─ .idea
  ├─ .vscode
  ├─ dist
  ├─ node_modules
  ├─ src
  └─ src-tauri
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/platform_impl/windows/system_tray.rs
+++ b/src/platform_impl/windows/system_tray.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use super::{
+  event_loop::S_U_TASKBAR_RESTART,
   menu::{subclass_proc as menu_subclass_proc, Menu, MenuHandler},
   util, OsError,
 };
@@ -27,12 +28,15 @@ use windows::{
 
 const WM_USER_TRAYICON: u32 = 6001;
 const WM_USER_UPDATE_TRAYMENU: u32 = 6002;
-const TRAYICON_UID: u32 = 6003;
-const TRAY_SUBCLASS_ID: usize = 6004;
-const TRAY_MENU_SUBCLASS_ID: usize = 6005;
+const WM_USER_UPDATE_TRAYICON: u32 = 6003;
+const TRAYICON_UID: u32 = 6004;
+const TRAY_SUBCLASS_ID: usize = 6005;
+const TRAY_MENU_SUBCLASS_ID: usize = 6006;
 
 struct TrayLoopData {
+  hwnd: HWND,
   hmenu: Option<HMENU>,
+  icon: Icon,
   sender: Box<dyn Fn(Event<'static, ()>)>,
 }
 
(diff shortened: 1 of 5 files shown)


src/platform_impl/windows/system_tray.rs
```diff
@@ -217,6 +223,18 @@ unsafe extern "system" fn tray_subclass_proc(
     subclass_input.hmenu = Some(HMENU(wparam.0 as _));
   }
 
+  if msg == WM_USER_UPDATE_TRAYICON {
+    let icon = wparam.0 as *mut Icon;
+    subclass_input.icon = (*icon).clone();
```
amrbashir, line 228: @wusyong does this correctly free the memory of the original icon? given that `subclass_input` is a raw pointer and it is only dropped when the app exists.

WilfredC64:
I've tested the fix and the system tray icon is now always visible after Windows Explorer restarts. Thanks!

The skip_taskbar fix doesn't work for my application.

I do see my window icon shown on the task bar when I restart Windows Explorer while skip_taskbar is set to all my created windows. My application has 2 windows created in memory but none visible. I don't have a window defined in tauri.conf.json. I create the 2 windows programmatically via WindowBuilder with "visible" false and "skip_taskbar" true. When a CloseRequested event is received I hide the window and call api.prevent_close().

When I run my application, then restart explorer, the application icon appear on the task bar. When I then close the window via the taskbar (right mouse click and select "Close window"), the icon on the task bar is gone. I can restart Windows Explorer now over and over and it will not appear anymore on the task bar and system tray icon is still visible.

However, I still can show the 2 windows via the system tray icon menu. When I then try to close my application via the system tray icon, my application wants to close the 2 created windows and then I'm getting an error on the close() method of the window. This error I only get when I close the window via the taskbar after a Windows Explorer restart.

I hope my description is clear. If not, let me know and I will setup a repository to reproduce the problem.

amrbashir:
I think I know what you mean. the icon that appears after the taskbar has been restarted, is it one of your windows? if not, can you open it and take a screenshot of it? if you can't open it, then I think I know where is the problem

WilfredC64:
It's not one of the windows that I created in the app. I cannot open it but I can hover over it, and that looks weird. I made a screenshot of it.
taskbar

amrbashir:
yup, seems consistent with what I observed. Will try to fix it too.

amrbashir:
The fix for the ghost window is in https://github.com/tauri-apps/tao/pull/489

WilfredC64:
I've tested the PR #489 and it fixes the issue that I mentioned here. All the issues are fixed now related to restarting Windows Explorer. Thanks.

src/platform_impl/windows/system_tray.rs
amrbashir: from a quick test in rust playground it is dropped correctly.

amrbashir:
Thanks for testing, It will be merged and published soon.

wusyong (approved):

PR #411, opened 2022-05-30
Remove trivial tray features

wusyong:
### What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [ ] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary

PR #492, opened 2022-07-22
Update libappindicator dependency

nazar-pc:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This addresses my comment https://github.com/tauri-apps/tao/pull/411#issuecomment-1192208890, in that PR `ayatana` support was removed as "unnecessary", but it is only the case with `libappindicator >= 0.7.1`, which wasn't bumped, so in our project old version was used even after Tauri upgrade.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,7 +115,7 @@ gdk = { version = "0.15", features = [ "v3_22" ] }
 gdk-sys = "0.15"
 gdkx11-sys = "0.15"
 gdk-pixbuf = { version = "0.15", features = [ "v2_36_8" ] }
-libappindicator = { version = "0.7", optional = true }
+libappindicator = { version = "0.7.1", optional = true }
 dirs-next = { version = "2.0.0", optional = true }
 x11-dl = "2.19"
 uuid = { version = "1.1", features = [ "v4" ] }


FabianLars:
> so in our project old version was used even after Tauri upgrade.

Yeah you need to run `cargo update` so cargo picks up the changes. But you're right, we should've included this in 0.10, thanks!

amrbashir:
shouldn't `0.7` resolve to `0.7.1` also?

FabianLars:
Yeah, but only if you remove the lockfile (or don't have one), or run cargo update. if your project is a little older with a lockfile that was just dragged along, cargo won't update it

amrbashir:
I now vaguely remember that `7.0.0` was broken or something and we had to release a patch version, right?

amrbashir (approved):

amrbashir:
Thanks for the PR.

FabianLars:
Yeah, well libappindicator-sys was basically broken _and_ its 0.7.1 should have been 0.8. so basically everything was just about sys, but there was a compat flag added to non-sys and that's why 0.7.1 of non-sys was released.

And now that i look at libappindicator, this PR actually fixes _nothing_, because lipappindicator depends on libappindicator-sys `0.7` and not `0.7.3`...

amrbashir:
Oh well, it doesn't hurt to merge anyways. I guess `cargo update` should be enough for ppl to fix?

FabianLars:
Yep it's enough 👍

Issue #494, opened 2022-07-24
Resize event isn't called on time


morajabi:
In order to preserve window control buttons position, we need to re-apply the logic to update their position on resize, move, and exit/start full-screen. (taken from Electron) Although the move event works fine, on resize we get a flicker. 

![CleanShot 2022-07-24 at 13 11 27](https://user-images.githubusercontent.com/12202757/180640676-a4f3acec-0e03-4e20-8bbe-414cd8b4d8ef.gif)

Code that updates the titlebar:
```
   unsafe {
            let id = self.ns_window().unwrap() as cocoa::base::id;
            NSWindow::setTitlebarAppearsTransparent_(id, cocoa::base::YES);
            let mut style_mask = id.styleMask();
            style_mask.set(
                NSWindowStyleMask::NSFullSizeContentViewWindowMask,
                title_transparent,
            );

            if remove_tool_bar {
                style_mask.remove(
                    NSWindowStyleMask::NSClosableWindowMask
                        | NSWindowStyleMask::NSMiniaturizableWindowMask
                        | NSWindowStyleMask::NSResizableWindowMask,
                );
            }

            id.setStyleMask_(style_mask);

            id.setTitleVisibility_(if title_transparent {
                NSWindowTitleVisibility::NSWindowTitleHidden
            } else {
                NSWindowTitleVisibility::NSWindowTitleVisible
            });

            id.setTitlebarAppearsTransparent_(if title_transparent {
                cocoa::base::YES
            } else {
                cocoa::base::NO
            });

            let title_bar_x = -4.0;
            let title_bar_y = 26.0;

            // TAKEN FROM ELECTRON
            // https://github.com/patr0nus/DeskGap/blob/master/lib/src/platform/mac/BrowserWindow.mm#L100
            // also need to handle on full screen enter, exit, resize, etc

            let close = NSWindow::standardWindowButton_(id, NSWindowButton::NSWindowCloseButton);
            let miniaturize = id.standardWindowButton_(NSWindowButton::NSWindowMiniaturizeButton);
            let zoom = id.standardWindowButton_(NSWindowButton::NSWindowZoomButton);

            let title_bar_container_view = close.superview().superview();

            let button_height = NSView::frame(close).size.height;
            let title_bar_frame_height = button_height + title_bar_y;
            let mut title_bar_rect = NSView::frame(title_bar_container_view);
            title_bar_rect.size.height = title_bar_frame_height;
[...]

PR #513, opened 2022-08-03
Inset Mac OS traffic lights

haasal:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [x] New Binding issue tauri-apps/tauri#4789
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

**Important**: There still exist artifacts when resizing the window because the traffic lights are set just after the window has already been drawn.


morajabi:
Thanks for this awesome PR! I had tried doing this only to fail at the "resize" artifacts: #494 Reading your solution made it clear. I should close that issue then.

PhotonQuantum:
This feature is fantastic!

However, I noticed a strange behavior: the hover effect still shows when I move the cursor to the "original" position of traffic lights. The electron implementation doesn't seem like have this behavior. I dug into its code but failed to find a solution.

Update: strangely, the problem goes away if I resize the window.

tr3ysmith:
Hey what's the status of this? I really would like to be able to apply inset on the traffic lights without work arounds

haasal:
Hi, me too. But I know really little about Tao. If this gets merged then we still have to implement it somehow into tauri.
Could someone review this please?

wusyong:
The implementation seems fine to me. But I'm not sure if it's good to merge now.
@amrbashir Do you think winit can accept this?
Btw you can get the raw window handle to change it yourself if you can't wait for it any longer.

amrbashir:
there is a good chance they may not accept but lets go ahead and merge this and if they didn't accept it upstream, we can implement it directly in tauri

examples/macos_traffic_lights.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2019-2021 Tauri Programme within The Commons Conservancy
```
amrbashir: example not needed

src/platform/macos.rs
```diff
@@ -54,6 +54,8 @@ pub trait WindowExtMacOS {
 
   /// Sets whether or not the window has shadow.
   fn set_has_shadow(&self, has_shadow: bool);
+
+  fn set_traffic_light_inset(&self, inset: (f64, f64));
```
amrbashir: missing documentation and use `LogicalSize` instead of a tuple

src/platform/macos.rs
```diff
@@ -338,6 +344,7 @@ pub trait WindowBuilderExtMacOS {
   fn with_resize_increments(self, increments: LogicalSize<f64>) -> WindowBuilder;
   fn with_disallow_hidpi(self, disallow_hidpi: bool) -> WindowBuilder;
   fn with_has_shadow(self, has_shadow: bool) -> WindowBuilder;
+  fn with_traffic_light_inset(self, inset: (f64, f64)) -> WindowBuilder;
```
amrbashir: missing documentation and use `LogicalSize` instead of a tuple

src/platform_impl/macos/window.rs
```diff
@@ -134,6 +136,7 @@ unsafe fn create_view(
   })
 }
 
+// TODO: Modify this
```
amrbashir: why modify this?

src/platform_impl/macos/window.rs
```diff
@@ -262,6 +265,11 @@ fn create_window(
         menu::initialize(window_menu);
       }
 
+      if let Some((offset_x, offset_y)) = pl_attrs.traffic_light_inset {
+        println!("{} {}", offset_x, offset_y);
```
amrbashir:
```suggestion
```

src/platform_impl/macos/window.rs
```diff
@@ -1397,3 +1409,29 @@ unsafe fn set_max_inner_size<V: NSWindow + Copy>(window: V, mut max_size: Logica
     window.setFrame_display_(current_rect, NO)
   }
 }
+
+pub(crate) unsafe fn position_traffic_lights<V: NSWindow + Copy>(window: V, x: f64, y: f64) {
```
amrbashir:
```suggestion
unsafe fn position_traffic_lights<W: NSWindow + Copy>(window: W, x: f64, y: f64) {
```

src/platform_impl/macos/window.rs
```diff
@@ -1397,3 +1409,29 @@ unsafe fn set_max_inner_size<V: NSWindow + Copy>(window: V, mut max_size: Logica
...
+  let close = window.standardWindowButton_(NSWindowButton::NSWindowCloseButton);
+  let miniaturize = window.standardWindowButton_(NSWindowButton::NSWindowMiniaturizeButton);
+  let zoom = window.standardWindowButton_(NSWindowButton::NSWindowZoomButton);
+
+  let title_bar_container_view = close.superview().superview();
+
+  let close_rect: NSRect = msg_send![close, frame];
```
amrbashir:
```suggestion
  let close_rect = close.frame();
```

src/platform_impl/macos/window.rs
```diff
@@ -1397,3 +1409,29 @@ unsafe fn set_max_inner_size<V: NSWindow + Copy>(window: V, mut max_size: Logica
...
+  let button_height = close_rect.size.height;
+
+  let title_bar_frame_height = button_height + y;
+  let mut title_bar_rect = NSView::frame(title_bar_container_view);
```
amrbashir:
```suggestion
  let mut title_bar_rect = title_bar_container_view.frame();
```

src/platform_impl/macos/window.rs
```diff
@@ -1397,3 +1409,29 @@ unsafe fn set_max_inner_size<V: NSWindow + Copy>(window: V, mut max_size: Logica
...
+  title_bar_rect.size.height = title_bar_frame_height;
+  title_bar_rect.origin.y = NSWindow::frame(window).size.height - title_bar_frame_height;
```
amrbashir:
```suggestion
  title_bar_rect.origin.y = window.frame().size.height - title_bar_frame_height;
```

src/platform_impl/macos/window.rs
```diff
@@ -1397,3 +1409,29 @@ unsafe fn set_max_inner_size<V: NSWindow + Copy>(window: V, mut max_size: Logica
...
+  let _: () = msg_send![title_bar_container_view, setFrame: title_bar_rect];
+
+  let window_buttons = vec![close, miniaturize, zoom];
+  let space_between = NSView::frame(miniaturize).origin.x - NSView::frame(close).origin.x;
```
amrbashir:
```suggestion
  let space_between = miniaturize.frame().origin.x - close_rect.origin.x;
```

src/platform_impl/macos/window.rs
```diff
@@ -1397,3 +1409,29 @@ unsafe fn set_max_inner_size<V: NSWindow + Copy>(window: V, mut max_size: Logica
...
+  let close = window.standardWindowButton_(NSWindowButton::NSWindowCloseButton);
+  let miniaturize = window.standardWindowButton_(NSWindowButton::NSWindowMiniaturizeButton);
+  let zoom = window.standardWindowButton_(NSWindowButton::NSWindowZoomButton);
+
+  let title_bar_container_view = close.superview().superview();
+
+  let close_rect: NSRect = msg_send![close, frame];
+  let button_height = close_rect.size.height;
+
+  let title_bar_frame_height = button_height + y;
```
amrbashir:
```suggestion
  let title_bar_frame_height = close_rect.size.height + y;
```

src/platform_impl/macos/window.rs
```diff
@@ -1397,3 +1409,29 @@ unsafe fn set_max_inner_size<V: NSWindow + Copy>(window: V, mut max_size: Logica
...
+  let mut title_bar_rect = NSView::frame(title_bar_container_view);
+  title_bar_rect.size.height = title_bar_frame_height;
+  title_bar_rect.origin.y = NSWindow::frame(window).size.height - title_bar_frame_height;
+  let _: () = msg_send![title_bar_container_view, setFrame: title_bar_rect];
+
+  let window_buttons = vec![close, miniaturize, zoom];
+  let space_between = NSView::frame(miniaturize).origin.x - NSView::frame(close).origin.x;
+
+  for (i, button) in window_buttons.into_iter().enumerate() {
+    let mut rect: NSRect = NSView::frame(button);
```
amrbashir:
```suggestion
    let mut rect = button.frame();
```

amrbashir (changes requested):

src/platform_impl/macos/window.rs
haasal: Honestly... I have no clue. I worked on this a long time ago

haasal:
I'm sorry this has taken so long. I didn't really have time to work on this. I will document the function and delete the example these days. I hope it's not to late to merge?

amrbashir (commented):
Looks fine to me, however I'll hand this to @wusyong and @JonasKruckenberg (since you were interested in this)

amrbashir:
Please rebase the PR to fix the conflicts and add a change file

JonasKruckenberg:
I'll review this tomorrow, it's a super important imo and this or has been open wayy too long 😉 let's get this over the finish line

haasal:
Maybe it's important to add that I couldn't test the changes since my MacBook broke 🥺. If one of you could do that it would be amazing

JonasKruckenberg:
After testing this PR now, I feel `LogicalSize` makes no sense here, the way you currently call the method is as such

```rust
                                      // WTH means width here? and height?
.with_traffic_light_inset(LogicalSize { width: 10, height: 20 });
```

JonasKruckenberg:
Okay a few other observations:

1. While this does successfully inset the traffic lights, it does not increase the draggable area size
2. When resizing the window, it looses the inset

The second one really is a dealbreaker right now, I guess we need to implement a delegate and stuff just like electron does as well

haasal:
> Okay a few other observations:
> 
>     1. While this does successfully inset the traffic lights, it does not increase the draggable area size
> 
>     2. When resizing the window, it looses the inset
> 
> 
> The second one really is a dealbreaker right now, I guess we need to implement a delegate and stuff just like electron does as well

I don't know if you fixed it already in that commit, but when I tested it with my supplied example the inset was persistent during resizing of the window.

JonasKruckenberg:
> I don't know if you fixed it already in that commit, but when I tested it with my supplied example the inset was persistent during resizing of the window.

Okay yes my bad, calling the `set_traffic_light_inset` method after the RedrawRequested event got fired fixes the artifacts.

My only question @haasal is why the `with_traffic_light_inset` method exists if it's not really doing what the user expects. This feels like misleading api design

JonasKruckenberg:
@amrbashir I added back the example, since it's indeed needed to use the method correctly (it's not enough to just call it once but it needs to be called after `RedrawRequested`)

haasal:
Yeah that was the reason I made the example. I already tried implementing this in Tauri itself using the raw window handle but I couldn't get rid of these artifacts. Do you think this method can now be used in tauri without causing artifacts?

haasal:
> My only question @haasal is why the `with_traffic_light_inset` method exists if it's not really doing what the user expects. This feels like misleading api design

I don't know how to implement it instead. This whole thing gave me big headaches because I couldn't find a way to set the inset without having to call the method in some weird way. If you have any idea how to do this without having to call the method after `RedrawRequested` it would be great!

amrbashir:
I still don't see why the example is needed since tao should already handle that internally and not require users to keep calling it in `RedrawRequested`

JonasKruckenberg:
> I still don't see why the example is needed since tao should already handle that internally and not require users to keep calling it in `RedrawRequested`

then we should change it to behave that way

haasal:
I really would like to get this over with but as said earlier I don't really know how. If you know how to directly inject something into the "window draw pipeline" (I.e. how to automatically call my inset-method/function without requiring the user to do some weird setup) I could take a look at it

amrbashir:
@haasal you need add an inset field on the `ViewState` struct here https://github.com/tauri-apps/tao/blob/c5d606dffeb1733ab06fd8c43eb3b9e7b2f553fe/src/platform_impl/macos/view.rs#L54 and your new function `set_traffic_light_inset` sets this field, then in `draw_rect` https://github.com/tauri-apps/tao/blob/c5d606dffeb1733ab06fd8c43eb3b9e7b2f553fe/src/platform_impl/macos/view.rs#L382  you have access to the state where you can read it and re-position the tarffic lights

haasal:
But the `with_traffic_light_inset()` method should remain as the public API method for the user to call, right?
The user just shouldn't ever have to call `set_traffic_light_inset()` manually.

amrbashir:
@haasal No, both methods should be public, both should modify the newly added state and triggers the traffic light positioning

haasal:
Sorry for not contributing to this in the last weeks. In spite of what I said I noticed that I don't currently have the time to finish this. I would really appreciate if one of you could push this over the finish line.

haasal:
Got a new mac and fixed it. I removed the set_traffic_light_inset() method because it wasn't really necessary anymore and would add unnecessary complexity. The with_traffic_light_inset() method on WindowBuilder should now work as expected.

Edit: I added the `set_traffic_light_inset()` back in. I overestimated how complicated it would be to implement this

haasal:
Imho apart from minor refactoring (if you see required) this should be ready to merge. Could you maybe take a look at this @JonasKruckenberg @amrbashir?

src/platform_impl/macos/view.rs
```diff
@@ -414,13 +420,10 @@ extern "C" fn reset_cursor_rects(this: &Object, _sel: Sel) {
...
-    if !cursor.is_null() {
-      let _: () = msg_send![this,
-          addCursorRect:bounds
-          cursor:cursor
-      ];
-    }
+    let _: () = msg_send![this,
+        addCursorRect:bounds
+        cursor:cursor
+    ];
```
amrbashir: why was this check removed?

src/platform_impl/macos/window.rs
```diff
@@ -1569,6 +1577,16 @@ impl WindowExtMacOS for UnownedWindow {
       ns_string_to_rust(tabbing_identifier)
     }
   }
+
+  #[inline]
+  fn set_traffic_light_position(&self, position: (f64, f64)) {
+    // NOTE: Not sure if &self shouldn't be &mut self
```
amrbashir:
```suggestion
```

src/platform_impl/macos/window.rs
```diff
@@ -1569,6 +1577,16 @@ impl WindowExtMacOS for UnownedWindow {
       ns_string_to_rust(tabbing_identifier)
     }
   }
+
+  #[inline]
+  fn set_traffic_light_position(&self, position: (f64, f64)) {
```
amrbashir:
```suggestion
  fn set_traffic_light_position<P: Into<Position>>(&self, position: P) {
```

amrbashir (changes requested):

src/platform_impl/macos/view.rs
haasal: Must have been on accident when deleting my own stuff

src/platform_impl/macos/window.rs
haasal: I don't really understand this. Do you want me to change only the public interface (i.e. set_traffic_..., with_...) or should I make the WindowBuilder field and the View attributes (or how they are called) also into a `position`? How do I extract x and y from this?

The x and y for the inset are LogicalPosition coordinates right? Can't we just use LogicalPosition as the inset Type instead of Position? This can be called passed using `LogicalPosition {x, y}` and so Jonas' argument against `LogicalSize` is satisfied

haasal:
@amrbashir ok, I implemented your feedback. I hope the `LogicalPosition` is ok

src/platform_impl/macos/window.rs
amrbashir: This change should be added to all public APIs, and then you can convert them into logical position internally before applying the inset.

amrbashir: Taking `Into<Position>` instead of `Into<LogicalPosition>` allows more flexibility and also to be consistent with other public APIs.

haasal: The issue is that in the `set_traffic_light_position()` implementation I don't have access to the `self.window.scale_factor()` method for `WindowBuilder` which is needed to convert from `Position` to `LogicalPosition`

amrbashir: you can get the scale factor using `NSWindow::backingScaleFactor(ns_window)` like this https://github.com/tauri-apps/tao/blob/bd8bafeca8a063351022cd5d5791f3111e8761d0/src/platform_impl/macos/window.rs#L820

haasal: Resolved?

amrbashir (approved):

amrbashir:
Thank you

amrbashir:
Just waiting for @wusyong to test and approve as well.

amrbashir:
@haasal looks like you didn't sign your commits and I can't merge the PR unless they are signed. You need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928) for example.

haasal:
@amrbashir Do i also need to sign the partially verified commits? Also @JonasKruckenberg 's commits seem not to be signed either, right?

wusyong:
> Just waiting for @wusyong to test and approve as well.

I'm okay with the API but the title bar is expanded. Is this what you expect?

Screenshot 2023-10-26 at 6 41 42 PM

haasal:
Oh god... I messed up didn't I?
@wusyong yes this is what one should expect. This will only work (look) properly with the overlay titlebar type

wusyong:
You can squash all your commits into one and then merge with dev. This should be the easiest way to sign commit.

haasal:
@wusyong I didn't do anything like this before and I don't want to mess up even worse. Do you mean pulling from upstream onto my dev branch and then merging with all of _my_ contributions squashed into one commit? Wouldn't this negate all my commit messages?

amrbashir (approved):
I guess it is ready for merge now, @wusyong would you mind testing and merge?

PR #516, opened 2022-08-04
[Feat] Added the possibility to enable a toolbar for the macos window

haasal:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

I wanted to add the feature as a toolbar can have different styles in the future, makes the traffic light inset look more native and allows for thicker toolbars.

### DIFF
--- a/examples/macos_toolbar.rs
+++ b/examples/macos_toolbar.rs
@@ -1,0 +1,54 @@
+// Copyright 2019-2021 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+
+// This example heavily uses the window.rs example as a template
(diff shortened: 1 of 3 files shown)


amrbashir:
if I understand correctly, `NSToolbar` is a UI element so if we include this, we also need to return its handle so it can be used afterwards. It also possible to add the toolbar in your code without having it directly in tao, so I don't think this should be added directly in tao.

haasal:
Alright. I'll close this.

Issue #470, opened 2022-07-10
Tracking issue for different implementation between `winit`


wusyong:
This is an issue to track which PR or commit we add that's not in winit yet.
We should either add to winit too or adjust the behavior in tao, so they are consistent.
- [X] [ALL] system tray - [tauri-apps/tray-icon](https://github.com/tauri-apps/tray-icon)
- [X] [ALL] menus - [tauri-apps/muda](https://github.com/tauri-apps/muda)
- [X] [ALL] global-hotkey - [tauri-apps/global-hotkey](https://github.com/tauri-apps/global-hotkey)
- [X] [Windows] ~~Transparent Windows don't work unless we draw to the window surface https://github.com/rust-windowing/winit/issues/2502~~, use `softbuffer` to clear the window surface
- [ ] [macOS] https://github.com/tauri-apps/tao/commit/1359fccfe5b95de9a28afb92e9ac3adfc331fb3c - Default cursor will not set to arrow cursor only
- [ ] ~~[ALL] https://github.com/tauri-apps/tao/commit/c0d5a5307cd6625622989ef529f3189967f3a76f - `monitor_from_point`~~ **REJECTED:** use `available_monitors()` and compare bounds with the point.
- [ ] [ALL] https://github.com/tauri-apps/tao/commit/5d8bf51d7ea2ea89feb4e8b003e583fed69bf300 - `cursor_position` - https://github.com/rust-windowing/winit/pull/2648
- [ ] [macOS] https://github.com/tauri-apps/tao/commit/045b768e30b4dc261edcaba4b8ed8ec9fee8305e - call `nextResponder` in `key_down`
- [ ] [iOS] https://github.com/tauri-apps/tao/commit/651137ce9ec5bf37593e6641d8f6ab79fc9d6f3c - iOS proxy Send
- [ ] [Windows] https://github.com/tauri-apps/tao/commit/62db431338636dd0adafca15d636c08f68941984 - focus hack
- [ ] [Windows] https://github.com/tauri-apps/tao/commit/c5d606dffeb1733ab06fd8c43eb3b9e7b2f553fe - auto-hide taskbar with maximized undecorated window
- [ ] [macOS] https://github.com/tauri-apps/tao/commit/813290130ea255b2cb45a66234a422519d13f667 - fix the unexpected shifting of the window when dragging after closing the share dialog
[...]

PR #517, opened 2022-08-04
feat(macos): add Reopen event, closes #218

### LINKED ISSUE #218: Support `activate` event on macOS
On macOS clicking the applications dock icon does a few things depending on the context:
- launch the application
- focus windows
- if all windows are closed create a new one
Afaik all these events are communicated through the `appkit::NSEventSubtype::NSApplicationActivatedEventType` event. Can we expose this even to consumers somehow? 

Add an `ActivationEvent` member to the `tao::event::Event` enum. This would mirror the `appkit::NSEventSubtype::NSApplicationActivatedEventType` event.

Some of this can be implemented through the use of `WindowEvent::Focus`, but that will only be fired on a per-window basis, not when all windows are closed.

**Would you assign yourself to implement this feature?**
- [x] Yes
- [ ] No

**Additional context**

BillGoldenWater:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
in macOS the app usually doesn't exit when all window closed, and will use click on dock icon for reopen window.
https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428638-applicationshouldhandlereopen?language=objc

---

this is a recreate of pr #515, due to it's not signed because misconfiguration in git

### DIFF
--- a/src/platform_impl/macos/app_delegate.rs
+++ b/src/platform_impl/macos/app_delegate.rs
@@ -4,8 +4,7 @@
 
 use crate::{platform::macos::ActivationPolicy, platform_impl::platform::app_state::AppState};
 
-use cocoa::base::id;
-use cocoa::foundation::NSString;
+use cocoa::{base::id, foundation::NSString};
 use objc::{
   declare::ClassDecl,
   runtime::{Class, Object, Sel, BOOL},
@@ -15,8 +14,7 @@ use std::{
   os::raw::c_void,
 };
 
-use cocoa::foundation::NSArray;
-use cocoa::foundation::NSURL;
+use cocoa::foundation::{NSArray, NSURL};
 use std::ffi::CStr;
 
 static AUX_DELEGATE_STATE_NAME: &str = "auxState";
@@ -53,6 +51,10 @@ lazy_static! {
     decl.add_method(
       sel!(application:openURLs:),
       application_open_urls as extern "C" fn(&Object, Sel, id, id),
+    );
+    decl.add_method(
+      sel!(applicationShouldHandleReopen:hasVisibleWindows:),
+      application_should_handle_reopen as extern "C" fn(&Object, Sel, id, BOOL) -> BOOL,
     );
     decl.add_method(
       sel!(applicationSupportsSecureRestorableState:),
@@ -123,6 +125,18 @@ extern "C" fn application_open_urls(_: &Object, _: Sel, _: id, urls: id) -> () {
   trace!("Get `application:openURLs:` URLs: {:?}", urls);
   AppState::open_urls(urls);
   trace!("Completed `application:openURLs:`");
+}
+
+extern "C" fn application_should_handle_reopen(
+  _: &Object,
+  _: Sel,
+  _: id,
+  has_visible_windows: BOOL,
+) -> BOOL {
+  trace!("Triggered `applicationShouldHandleReopen`");
+  AppState::reopen(has_visible_windows);
+  trace!("Completed `applicationShouldHandleReopen`");
+  has_visible_windows
 }
 
 extern "C" fn application_supports_secure_restorable_state(_: &Object, _: Sel, _: id) -> BOOL {
--- a/src/platform_impl/macos/app_delegate.rs
+++ b/src/platform_impl/macos/app_delegate.rs
@@ -4,7 +4,10 @@
 
 use crate::{platform::macos::ActivationPolicy, platform_impl::platform::app_state::AppState};
 
-use cocoa::{base::id, foundation::NSString};
+use cocoa::{
+  base::{id, NO},
+  foundation::NSString,
+};
 use objc::{
   declare::ClassDecl,
   runtime::{Class, Object, Sel, BOOL},
@@ -134,7 +137,7 @@ extern "C" fn application_should_handle_reopen(
   has_visible_windows: BOOL,
 ) -> BOOL {
   trace!("Triggered `applicationShouldHandleReopen`");
-  AppState::reopen(has_visible_windows);
+  AppState::reopen(has_visible_windows != NO);
   trace!("Completed `applicationShouldHandleReopen`");
   has_visible_windows
 }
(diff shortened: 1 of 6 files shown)


alchaplinsky:
Are there any plans on merging this?

wusyong:
We plan to switch to winit as #470 described. So winit also needs to implement this one.
@amrbashir Do you think we should merge this?

amrbashir:
last time I checked, the winit team has not decided on how they want to introduce this so I think we should keep this open until we make the switch.

alchaplinsky:
@wusyong By switching you mean that `winit` will be used in `tauri` as a window management library instead of `tao`?

wusyong:
> @wusyong By switching you mean that `winit` will be used in `tauri` as a window management library instead of `tao`?

We plan to do so in v2.0. If it's not that urgent, I suppose open a PR on winit is a good start.

alchaplinsky:
@wusyong Ok got it. Makes sense. But is there a way currently in tauri to handle this issue https://github.com/tauri-apps/tauri/issues/3084 (opening a window on dock icon click)?

wusyong:
I don't think so until this PR is merged in winit.

lencx:
Use minimized windows instead of hide.

```rs
tauri::Builder::default()
    .on_window_event(|event| {
        if let tauri::WindowEvent::CloseRequested { api, .. } = event.event() {
            // event.window().hide().unwrap();
            event.window().minimize().unwrap();
            api.prevent_close();
        }
    })
```

tance77:
Any news with this? I would love to be able to hide the window and show it rather than minimize it.

tance77:
Would love to see this in the next version of Tauri. Missing this functionality. Great work.

src/platform_impl/macos/app_delegate.rs
```diff
@@ -125,6 +130,18 @@ extern "C" fn application_open_urls(_: &Object, _: Sel, _: id, urls: id) -> () {
...
+extern "C" fn application_should_handle_reopen(
+  _: &Object,
+  _: Sel,
+  _: id,
+  has_visible_windows: BOOL,
+) -> BOOL {
+  trace!("Triggered `applicationShouldHandleReopen`");
+  AppState::reopen(has_visible_windows != NO);
+  trace!("Completed `applicationShouldHandleReopen`");
+  has_visible_windows
```
FabianLars, line 142: I'm not sure if using has_visible_windows as the return type is a good idea. 🤔 

It's required to make event_loop.hide_application() to keep working out of the box but it may still cause incorrect behavior somewhere (or even with hide_application if the user doesn't want the default show behavior).

I'm not sure how we could best expose this to tauri consumers etc so maybeee just leave it like this for now (but adding a comment acknowledging that the logic is technically incorrect) and see where this will get us 🤔 

What do you think @pewsheen ?

BillGoldenWater: some context: https://github.com/electron/electron/blob/main/shell/browser/mac/electron_application_delegate.mm#L108-L113

(I tried some simple/cheap method to pass a thing to the user, but all of them failed :(
(is passing a wrapper struct around a shared reference to mutex/atomic and expose a method to modify it acceptable?

FabianLars: > some context: https://github.com/electron/electron/blob/main/shell/browser/mac/electron_application_delegate.mm#L108-L113

Interesting that electron does it too, then i guess it won't be too bad and just validates my statement above to just try this as is :)

> is passing a wrapper struct around a shared reference to mutex/atomic and expose a method to modify it acceptable?

Sometimes yes. In wry we also do this in some places with a static OnceCell or Lazy. In some places also use objc ffi to set a function on the object/delegate for later use but i personally prefer the full rust solution (the static vars) over that shitshow.

BillGoldenWater: https://github.com/BillGoldenWater/tao/compare/dev...BillGoldenWater:tao:feat/reopen_ret_test

implemented a test for expose this to user, could you take a look?

pronebird: I have tested this. Basically if `hasVisibleWindows = false` and you return `true` from `applicationShouldHandleReopen()` then the system calls [`applicationOpenUntitledFile()`](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428491-applicationopenuntitledfile). Probably does some other similar things for document based apps. So this is the default behavior that we disable here by returning `false`. 

I think Electron actually do it kinda right.  But since we don't handle `applicationOpenUntitledFile()` then it really doesn't matter all that much.

pewsheen: I think we could go this way first and see if anyone has the need to control the return value. If so, we can add something like https://github.com/tauri-apps/wry/pull/1196 to accept a handler from caller.

FabianLars: Yeah i agree, maybe by then winit also has more ideas for their platform specific events 😂 

If electron did this for 10 years i think it's good enough for us too x) At least i didn't see any complains about it in electron-land.

FabianLars: cc @amrbashir because 🕺

amrbashir: this is fine for me as is for now, let's ship this and worry about other stuff later when requestred

amrbashir (approved):

amrbashir:
Thanks @BillGoldenWater for your work and sorry for the very long delay, do you plan to work on adding this event to tauri?

BillGoldenWater:
> do you plan to work on adding this event to tauri?

yes

amrbashir:
Thank you 🙏

pronebird:
Really looking forward that!

ayangweb:
@BillGoldenWater @amrbashir 

image

Why don't I have a `Reopen` event?

ayangweb:
@BillGoldenWater @amrbashir How does it work in tauri 1?

ayangweb:
@FabianLars How does it work in tauri v1?

PR #522, opened 2022-08-06
Always on bottom

### LINKED ISSUE #113: feat: Add always below other windows
reason1: for completeness
use case: since it is possible to make the window transparent and not show borders/decorations i thought it would be nice to make a new multiplatform conky alternative which would be nicer to configure since its all html/css. As far as i can see, this is the only feature really missing for this.

henry40408:
closes #113

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

- `Linux` I'd like to implement `is_always_below_bottom` and `is_always_on_top` as well but there is a long-living upstream bug in GTK https://gitlab.gnome.org/GNOME/gtk/-/issues/682
- `Windows` might help for implementation https://stackoverflow.com/a/65052538

### DIFF
--- a/src/platform_impl/linux/window.rs
+++ b/src/platform_impl/linux/window.rs
@@ -215,7 +215,12 @@ impl Window {
     window.set_visible(attributes.visible);
     window.set_decorated(attributes.decorations);
 
-    window.set_keep_above(attributes.always_on_top);
+    if attributes.always_below_bottom && !attributes.always_on_top {
+      window.set_keep_below(attributes.always_below_bottom);
+    }
+    if attributes.always_on_top && !attributes.always_below_bottom {
+      window.set_keep_above(attributes.always_on_top);
+    }
     if let Some(icon) = attributes.window_icon {
       window.set_icon(Some(&icon.inner.into()));
     }
@@ -552,6 +557,15 @@ impl Window {
     }
   }
 
+  pub fn set_always_below_bottom(&self, always_below_bottom: bool) {
+    if let Err(e) = self.window_requests_tx.send((
+      self.window_id,
+      WindowRequest::AlwaysBelowBottom(always_below_bottom),
+    )) {
+      log::warn!("Fail to send always on bottom request: {}", e);
+    }
+  }
+
   pub fn set_always_on_top(&self, always_on_top: bool) {
     if let Err(e) = self
       .window_requests_tx
@@ -757,6 +771,7 @@ pub enum WindowRequest {
   DragWindow,
   Fullscreen(Option<Fullscreen>),
   Decorations(bool),
+  AlwaysBelowBottom(bool),
   AlwaysOnTop(bool),
   WindowIcon(Option<Icon>),
   UserAttention(Option<UserAttentionType>),
--- a/src/platform_impl/linux/window.rs
+++ b/src/platform_impl/linux/window.rs
@@ -215,12 +215,15 @@ impl Window {
     window.set_visible(attributes.visible);
     window.set_decorated(attributes.decorations);
 
-    if attributes.always_below_bottom && !attributes.always_on_top {
-      window.set_keep_below(attributes.always_below_bottom);
-    }
-    if attributes.always_on_top && !attributes.always_below_bottom {
+    if attributes.always_on_bottom && attributes.always_on_top {
+      log::warn!(
+        "Always on top and always on bottom, are both specified, you should only set one."
+      );
+    } else if attributes.always_on_bottom || attributes.always_on_top {
+      window.set_keep_below(attributes.always_on_bottom);
       window.set_keep_above(attributes.always_on_top);
     }
+
     if let Some(icon) = attributes.window_icon {
       window.set_icon(Some(&icon.inner.into()));
     }
@@ -557,10 +560,10 @@ impl Window {
     }
   }
 
-  pub fn set_always_below_bottom(&self, always_below_bottom: bool) {
+  pub fn set_always_on_bottom(&self, always_on_bottom: bool) {
     if let Err(e) = self.window_requests_tx.send((
       self.window_id,
-      WindowRequest::AlwaysBelowBottom(always_below_bottom),
+      WindowRequest::AlwaysOnBottom(always_on_bottom),
     )) {
       log::warn!("Fail to send always on bottom request: {}", e);
     }
@@ -771,7 +774,7 @@ pub enum WindowRequest {
   DragWindow,
   Fullscreen(Option<Fullscreen>),
   Decorations(bool),
-  AlwaysBelowBottom(bool),
+  AlwaysOnBottom(bool),
   AlwaysOnTop(bool),
   WindowIcon(Option<Icon>),
   UserAttention(Option<UserAttentionType>),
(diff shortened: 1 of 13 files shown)


src/platform_impl/linux/window.rs
```diff
@@ -215,7 +215,12 @@ impl Window {
     window.set_visible(attributes.visible);
     window.set_decorated(attributes.decorations);
 
-    window.set_keep_above(attributes.always_on_top);
+    if attributes.always_below_bottom && !attributes.always_on_top {
+      window.set_keep_below(attributes.always_below_bottom);
+    }
+    if attributes.always_on_top && !attributes.always_below_bottom {
+      window.set_keep_above(attributes.always_on_top);
+    }
```
amrbashir: I think we should return an Err if both are specified
```rs
if attributes.always_on_top && attributes.always_below_bottom {
	return Err("Always on top and always on bottom, are both specified, you should only set one.");
}
```
Either that, or in `WindowBuilder` we can set the other one false depending on which one was called last
```rs
pub fn with_always_on_top(mut self, always_on_top: bool) -> Self {
    self.window.always_on_top = always_on_top;
    self.window.always_below_bottom = false;
    self
}
pub fn with_always_below_bottom(mut self, always_below_bottom: bool) -> Self {
    self.window.always_below_bottom = always_below_bottom;
    self.window.always_on_top = false;
    self
}
```

cc @wusyong

src/window.rs
```diff
@@ -190,6 +190,11 @@ pub struct WindowAttributes {
   /// The default is `false`.
   pub always_on_top: bool,
 
+  /// Whether the window should always be on bottom of other windows.
+  ///
+  /// The default is `false`.
+  pub always_below_bottom: bool,
```
amrbashir: Should be renamed to 
```suggestion
  pub always_on_bottom: bool,
```
and all related methods should be renamed too

src/window.rs
```diff
@@ -784,6 +801,16 @@ impl Window {
     self.window.set_decorations(decorations)
   }
 
+  /// Change whether or not the window will always be below other windows.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **iOS / Android:** Unsupported.
```
amrbashir, line 811:
```suggestion
  /// - **Windows**: There is no guarantee that the window will be the bottom most but it will try to be.
  /// - **iOS / Android:** Unsupported.
```

amrbashir (changes requested):
Thank you, the implementation for Linux and Windows LGTM but I will leave macOS for @wusyong to approve. There is a couple of nit-picks but this should be mostly good to go.

src/platform_impl/linux/window.rs
wusyong: lgtm. Just tested on macOS too and it works fine.

src/window.rs
henry40408: 3f4feba

src/window.rs
henry40408: 3f4feba

src/platform_impl/linux/window.rs
henry40408: 3f4feba

henry40408: I cannot choose which error should be returned when "always on top" and "always on bottom" are both set, so I print warning instead.

src/platform_impl/linux/window.rs
```diff
@@ -215,7 +215,15 @@ impl Window {
...
-    window.set_keep_above(attributes.always_on_top);
+    if attributes.always_on_bottom && attributes.always_on_top {
+      log::warn!(
+        "Always on top and always on bottom, are both specified, you should only set one."
+      );
+    } else if attributes.always_on_bottom || attributes.always_on_top {
+      window.set_keep_below(attributes.always_on_bottom);
+      window.set_keep_above(attributes.always_on_top);
+    }
+
```
amrbashir: The warning is not needed anymore, since in your latest commit, the builder will disable one of them
```suggestion
    if attributes.always_on_bottom {
      window.set_keep_below(attributes.always_on_bottom);
    }

    if attributes.always_on_top {
      window.set_keep_above(attributes.always_on_top);
    }

```

amrbashir (changes requested):

src/platform_impl/linux/window.rs
henry40408: a563d89

I also [set `ALWAYS_ON_BOTTOM` flag](https://github.com/tauri-apps/tao/pull/522/files#diff-edacb93792c7cf2d4cfe6f4ff2259781ed3c13e060fcc381e9cbb30ce481a336R858) when attribute is set on Windows.

amrbashir (approved):
Thank you

jabza:
@henry40408  This is great. Although I'm noticing on Windows when clicking the 'Show Desktop' button, the window still gets hidden annoyingly. Any ideas?

henry40408:
hi @jabza I just tested and confirmed that window is always hidden whichever I set it "always on top" or "always on bottom", so it might not relate to "always on bottom" or "always on top". Would you kindly open another issue?

amrbashir:
@jabza this is a normal behavior of Windows. Nothing we can control. I think what you want is to make your tauri window behave like a desktop wallpaper? in this case check out my comment here https://github.com/tauri-apps/tauri/issues/4261#issuecomment-1146767667

jabza:
@henry40408 @amrbashir I see. For now I guess the wallpaper/progman child approach seems like the only way to remain visible upon 'Show Desktop'.

Windows 11 is re-introducing the concept of Window 'widgets' which will be interesting to see how that changes things.

I also tried OR'ing in the NOZORDER flag and pretty much everything else I could find on the topic. It seems like perhaps this used to be possible but not any more.

amrbashir:
@jabza Windows has no concept of "always on bottom", but you can change the z-order of the window to be the bottom most but that can be changed later due to many factors. The only way you can be truly on bottom is by using wallpaper/progman approach. Also the "Show Desktop" button hides all windows to show the destkop and since an "always on bottom" window is still a window, it makes sense for it to be hidden.

jabza:
@amrbashir understood, but it is possible for windows to 'survive' "Show Desktop", assuming they're "always on top". I guess that's the one exception Windows makes. Thanks for the alternative solution, i'll go ahead and attempt a plugin!

amrbashir:
even "Always on top" can't survive "Show desktop" because "Always on top" guarantees it is on top of everything else but it doesn't guarantee that it can't be hidden or minimized.

jabza:
I've tested it locally with Always on Top set and it does survive the event for me

amrbashir:
That's interesting behavior that I didn't know about it but unfortunately Windows doesn't guarantee the same for "always on bottom".

jabza:
All good, your alternative solution looks to be the ticket :) Hopefully come Windows 11 update a cleaner way will present itself!

amrbashir:
What you want to do is probably similar to Rainmeter, so you could try looking into their source code and see if you can find anything related there.

jabza:
Wow this is interesting... https://github.com/rainmeter/rainmeter/blob/78a24859aa2aafd69e819619eb9bbf44b77d2bd3/Library/System.cpp they seem to be using a 'dummy window' to detect the 'Show Desktop' event... there's some gems in here I'm sure...

amrbashir:
yeah Rainmeter has been around for a while and I bet they have a lot of experience/knowledge in this area.

PR #533, opened 2022-08-16
Window::is_focused

### LINKED ISSUE #111: feat: add `is_focused()`
In windows 10 the isVisible() always return true if the app is minimized or not focus.

henry40408:
closes #111 

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

N/A

### DIFF
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -575,6 +575,14 @@ impl UnownedWindow {
     }
   }
 
+  #[inline]
+  pub fn is_focused(&self) -> bool {
+    unsafe {
+      let is_key_window: BOOL = msg_send![*self.ns_window, isKeyWindow];
+      is_key_window == YES
+    }
+  }
+
   pub fn request_redraw(&self) {
     AppState::queue_redraw(RootWindowId(self.id()));
   }
--- a/.changes/is-focused.md
+++ b/.changes/is-focused.md
@@ -1,0 +1,5 @@
+---
+"tao": "patch"
+---
+
+Add `Window::is_focused`.
--- a/examples/window_debug.rs
+++ b/examples/window_debug.rs
@@ -127,6 +127,12 @@ fn main() {
         window_id,
         ..
       } if window_id == window.id() => *control_flow = ControlFlow::Exit,
+      Event::WindowEvent {
+        event: WindowEvent::Focused(focused),
+        ..
+      } => {
+        dbg!(focused, window.is_focused());
+      }
       _ => (),
     }
   });
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -598,6 +598,11 @@ impl Window {
     warn!("set_focus not yet implemented on Android");
   }
 
+  pub fn is_focused(&self) -> bool {
+    log::warn!("`Window::is_focused` is ignored on Android");
+    false
+  }
+
   pub fn set_resizable(&self, _resizeable: bool) {}
 
   pub fn set_minimized(&self, _minimized: bool) {}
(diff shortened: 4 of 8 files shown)


src/platform_impl/macos/window.rs
```diff
@@ -575,6 +575,14 @@ impl UnownedWindow {
     }
   }
 
+  #[inline]
+  pub fn is_focused(&self) -> bool {
+    unsafe {
+      let is_key_window: BOOL = msg_send![*self.ns_window, isKeyWindow];
+      is_key_window == YES
+    }
+  }
```
amrbashir, line 584: I am not sure if this enough, `Window::set_focus` calls `makeKeyAndOrderFront` and you are only checking for the key part, can we also check for the front order? I am not a macOS dev so I will let @wusyong approve this one

amrbashir (approved):

src/platform_impl/macos/window.rs
wusyong: Will probably need to check the order too. But I fear it will become a little bit too difficult. It's good to merge until we see any edge case I guess.

amrbashir: @henry40408 , do you think you can check for order too in this PR? or should I merge?

henry40408: @amrbashir I'd like to try checking order as well. Please do not merge.

henry40408: According to [Window Types and Behavior](https://www.oreilly.com/library/view/building-cocoa-applications/0596002351/ch01s05.html), I should use [`isMainWindow`](https://developer.apple.com/documentation/appkit/nswindow/1419130-ismainwindow) instead. I copy the paragraph here in case the source is missing:

> The key window is the window or dialog that will respond to the Mac keyboard. The main window is the document window that corresponds to the **active document** (e.g., a document window in a word processor or image window in a graphics editor).

I try to search for definitions of key window and main window in Apple documentation but no luck, so I provided alternative source here to clarify why I use `isMainWindow` instead.

amrbashir: from the link you provided for `isMainWindow`:
> The value of this property is YES when the window is the main window for the application; otherwise, NO.

So `isMainWindow` will only return true if it is the main window of the application, so I don't think that's it. 

Also we should keep the check for `keyWindow`, we need to have two checks:
1. is this window a `keyWindow`?
2. is this key the active window (probably the top most foreground window)?

henry40408: I modified `is_focused` so it will only return true when the following conditions are true:

1. is this window a key window ➡️ `isKeyWindow`
2. is this key the active window ➡️ `isMainWindow`

amrbashir: we can't depend on `isMainWindow` because it doesn't take other applications in consideration and it is not necessary that the mainWindow is the active window

henry40408: So we can just check `isKeyWindow` only?

amrbashir: we also want to check if this is top most window, if there is no API for that, we can just isKeyWindow only

henry40408: I did a quick search in [`NSWindow`](https://developer.apple.com/documentation/appkit/nswindow), there is no API to check whether the window is front-most or not. If we all agree we can just check `isKeyWindow` for now, then we can merge this pull request.

By the way, I have found [`orderedIndex`](https://developer.apple.com/documentation/appkit/nswindow/1449577-orderedindex) but whether the window has focus or not, it always returns `1`, so maybe the meaning of "order" in `orderedIndex` differs from the "order" in `makeKeyAndOrderFront`.

amrbashir: I guess we can use `isKeyWindow` only for now, thanks for looking into it

amrbashir (approved):

PR #544, opened 2022-08-31
Update windows to 0.39.0

wravery:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

The `alloc` feature has been removed, which means Win32 APIs no longer accept Rust `String` or `&str` parameters and implicitly convert them to `PWSTR` or `PSTR`.

For string literals, that feature was replaced with `s!()` and `w!()` macros which null terminate the string literal at compile time and convert to UTF-16 if necessary. The `s!()` macro is fine, however the `w!()` macro uses `HSTRING` types from WinRT for maximum compatibility with WinRT types. Since Tao only uses Win32 APIs (or dynamically loaded DLL entry-points), it still works on Win7, but I'm worried about taking any dependencies on `HSTRING`. Instead of the `w!()` macro, this change relies on the internal `util::encode_wide` function to convert to a `Vec<u16>` with the Rust standard library.

### DIFF
--- a/.changes/windows-0.39.0.md
+++ b/.changes/windows-0.39.0.md
@@ -6,4 +6,4 @@ Update `windows-rs` to the latest 0.39.0 release.
 
 The `alloc` feature has been removed, which means it no longer accepts Rust `String` or `&str` parameters and implicitly converts them to `PWSTR` or `PSTR`.
 
-For string literals, that feature was replaced with `s!()` and `w!()` macros which null terminate the string literal at compile time and convert to UTF-16 if necessary.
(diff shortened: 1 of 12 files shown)


wravery:
This should take care of #443 and #444.

amrbashir (approved):

amrbashir:
Thank you

PR #552, opened 2022-09-04
fix(deps): Use less specific version for once_cell

FabianLars:
Reason: One of tauri's dependencies lock it to 1.13.

Edit: This dependency is specific to the android target which i don't have set up right now. I didn't see anything in the source code that needs a newer version than 1.0 but if we want to be absolutely sure we need someone with a working android setup who can lock it to 1.0.2 or something locally and try it out.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ jni = "0.19"
 ndk = "0.6"
 ndk-sys = "0.3"
 ndk-context = "0.1"
-once_cell = "1.14"
+once_cell = "1"
 paste = "1.0"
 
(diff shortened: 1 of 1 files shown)


lucasfernog (approved):

lucasfernog:
Gotta do the same in wry and tauri right?

FabianLars:
I didn't look for it in the other repos because this was enough to make the API example compile again :sweat_smile:

PR #565, opened 2022-09-23
feat(macos): support to change tray icon aspect ratio, close #564

### LINKED ISSUE #564: [feat] Supports to change tray icon aspect ratio (macOS)
image

Hi, I try to achieve something like this on the macOS tray, but the tray icon size has been fixed.

https://github.com/tauri-apps/tao/blob/93082da8c98fe7fb2f14e1245a693cbe0372f057/src/platform_impl/macos/system_tray.rs#L169-L170

zzzgydi:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/macos-tray-icon-aspect-ratio.md
+++ b/.changes/macos-tray-icon-aspect-ratio.md
@@ -1,5 +1,5 @@
 ---
-"tao": "minor"
+"tao": "patch"
 ---
 
-Added `icon_aspect_ratio` to `SystemTray` to change the tray icon aspect ratio on macOS.
+Scale the tray icon according to its aspect ratio on macOS.
--- a/.changes/macos-tray-icon-aspect-ratio.md
+++ b/.changes/macos-tray-icon-aspect-ratio.md
@@ -1,0 +1,5 @@
+---
+"tao": "minor"
+---
+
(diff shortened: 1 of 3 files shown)


amrbashir (changes requested):
I think we should just remove the hard-coded values and use the passed image height and width directly without modifications to match other platforms.

zzzgydi:
Will there be a breaking change if we use the icon width and height directly?

amrbashir:
No, since other platforms already have this behavior. We just need to confirm that the `system_tray` example looks the same after/before the change. I mean the icon is not blurry and is not bigger than other tray icons.

zzzgydi:
image

It will bigger than other tray icons on macOS, I think the height should be 18.

amrbashir:
maybe try keeping the height fixed to 18.0 and calculate the width dynamically based on ratio of the original height to the fixed height so it would be something like this

```rs
let height = 18.0;
let width = icon.width / (icon.height / height);
```

This way it will keep the original behavior and allow you to specify a different width for your icon.

So if you have an icon that is 32Hx120W, you get a 18Hx67W in tray

amrbashir (approved):

PR #570, opened 2022-09-25
fix(menus): add app name to native Quit and Hide items, closes #536

### LINKED ISSUE #536: `Quit` and `Hide` native menu items should include app name
Just like the [`About`](https://docs.rs/tauri/1.0.5/tauri/enum.MenuItem.html#variant.About) menu item, the [`Quit`](https://docs.rs/tauri/1.0.5/tauri/enum.MenuItem.html#variant.Quit) and [`Hide`](https://docs.rs/tauri/1.0.5/tauri/enum.MenuItem.html#variant.Hide) items should also include the app name on macOS to follow platform conventions.

Here's an example of the correct behaviour:
Screenshot 2022-08-17 at 22 14 38

And here's a Tauri app:
Screenshot 2022-08-17 at 22 23 03

caesar:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Implemented by getting the name of the current app from `NSRunningApplication` and adding it to the menu items labels.

closes #536

### DIFF
--- a/.changes/macos-menus-appname.md
+++ b/.changes/macos-menus-appname.md
@@ -1,5 +1,5 @@
 ---
-"tao": minor
+"tao": patch
 ---
 
-Add the application name to the "Quit" and "Hide" native menu items.
+Add the application name to the "Quit" and "Hide" native menu items on macOS.
--- a/.changes/macos-menus-appname.md
+++ b/.changes/macos-menus-appname.md
@@ -1,0 +1,5 @@
+---
+"tao": minor
+---
+
+Add the application name to the "Quit" and "Hide" native menu items.
--- a/src/platform_impl/macos/menu.rs
+++ b/src/platform_impl/macos/menu.rs
@@ -210,7 +210,7 @@ impl Menu {
             menu_type,
           ),
         ))
-      },
+      }
       MenuItem::Hide => {
         let label = match get_app_name() {
           Some(name) => format!("Hide {}", name),
@@ -226,7 +226,7 @@ impl Menu {
             menu_type,
           ),
         ))
-      },
+      }
       MenuItem::HideOthers => Some((
         None,
         make_menu_item(
--- a/src/platform_impl/macos/menu.rs
+++ b/src/platform_impl/macos/menu.rs
@@ -26,7 +26,7 @@ use crate::{
 use super::{
   app_state::AppState,
   event::EventWrapper,
-  util::{app_name, ns_string_to_rust},
+  util::{app_name_string, ns_string_to_rust},
   window::get_window_id,
 };
 
@@ -197,40 +197,24 @@ impl Menu {
           menu_type,
         ),
       )),
-      MenuItem::Quit => {
-        let label = unsafe {
-          match app_name() {
-            Some(app_name) => format!("Quit {}", ns_string_to_rust(app_name)),
-            None => "Quit".to_string(),
-          }
-        };
-        Some((
-          None,
-          make_menu_item(
-            label.as_str(),
-            Some(selector("terminate:")),
-            Some(Accelerator::new(RawMods::Meta, KeyCode::KeyQ)),
-            menu_type,
-          ),
-        ))
-      }
-      MenuItem::Hide => {
-        let label = unsafe {
-          match app_name() {
-            Some(app_name) => format!("Hide {}", ns_string_to_rust(app_name)),
-            None => "Hide".to_string(),
-          }
-        };
-        Some((
-          None,
-          make_menu_item(
-            label.as_str(),
-            Some(selector("hide:")),
-            Some(Accelerator::new(RawMods::Meta, KeyCode::KeyH)),
-            menu_type,
-          ),
-        ))
-      }
+      MenuItem::Quit => Some((
+        None,
+        make_menu_item(
+          format!("Quit {}", unsafe { app_name_string() }.unwrap_or_default()).trim(),
+          Some(selector("terminate:")),
+          Some(Accelerator::new(RawMods::Meta, KeyCode::KeyQ)),
+          menu_type,
+        ),
+      )),
+      MenuItem::Hide => Some((
+        None,
+        make_menu_item(
+          format!("Hide {}", unsafe { app_name_string() }.unwrap_or_default()).trim(),
+          Some(selector("hide:")),
+          Some(Accelerator::new(RawMods::Meta, KeyCode::KeyH)),
+          menu_type,
+        ),
+      )),
       MenuItem::HideOthers => Some((
         None,
         make_menu_item(
--- a/src/platform_impl/macos/menu.rs
+++ b/src/platform_impl/macos/menu.rs
@@ -11,8 +11,7 @@ use objc::{
   declare::ClassDecl,
   runtime::{Class, Object, Sel, NO, YES},
 };
-
-use std::{str, sync::Once};
+use std::sync::Once;
 
 use crate::{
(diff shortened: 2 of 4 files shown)


caesar:
Sorry, forgot `cargo fmt` again… fixed now

src/platform_impl/macos/menu.rs
```diff
@@ -607,3 +622,18 @@ impl Accelerator {
     flags
   }
 }
+
+fn get_app_name() -> Option<&'static str> {
```
amrbashir: there is a similar utility function in `src/plagrom_impl/macos/util/mod.rs`

src/platform_impl/macos/menu.rs
```diff
@@ -607,3 +622,18 @@ impl Accelerator {
...
+  unsafe {
+    let app_class = class!(NSRunningApplication);
+    let app: id = msg_send![app_class, currentApplication];
+    let name: *const Object = msg_send![app, localizedName];
+    let size: usize = msg_send![name, lengthOfBytesUsingEncoding:4];
+    let bytes: *const c_char = msg_send!(name, UTF8String);
+    let c_str = str::from_utf8(slice::from_raw_parts(bytes as *const u8, size));
```
amrbashir: you can use `util::ns_string_to_rust`

amrbashir (changes requested):

src/platform_impl/macos/menu.rs
caesar: Huh, I looked and didn't find anything equivalent, but I'll have another look later

src/platform_impl/macos/menu.rs
caesar: Oh perfect!

src/platform_impl/macos/menu.rs
caesar: I found the function you're referring to but it doesn't seem to work, at least in the example apps. I think the problem is it's only looking in the `Info.plist` file whereas the API I used also checks other sources.

Since that function isn't used anywhere in Tao, can I modify the function signature? If possible I'd like to change the return type to `Option<&str>` and also make it safe. It [doesn't seem to be exported publicly](https://docs.rs/tao/latest/tao/?search=app_name) so I guess it won't be a breaking change – right?

caesar:
This feels to me like a much cleaner implementation now, thanks for the pointers @amrbashir.

Could you just confirm that you're happy with [these changes to the `app_name` function](https://github.com/tauri-apps/tao/pull/570/files#diff-66f1e62f74c201cb5d003969827b239e85d71af03288ecdb3e55fab5348a0baaL134)?

FabianLars:
I think that the Option should be reintroduced. You can't really expect that CFBundleName is set in all scenarios

src/platform_impl/macos/util/mod.rs
```diff
@@ -131,16 +131,13 @@ pub unsafe fn ns_string_to_rust(ns_string: id) -> String {
...
-    Some(app_name)
-  } else {
-    None
+/// Gets the app's name from the `localizedName` property of `NSRunningApplication`
+pub fn app_name() -> String {
+  unsafe {
+    let app_class = class!(NSRunningApplication);
+    let app: id = msg_send![app_class, currentApplication];
+    let name: id = msg_send![app, localizedName];
+    ns_string_to_rust(name)
```
amrbashir: just took a glance at the relevant docs and seems like `localizedName`  could be undefned and also could be a localized name. I don't use mac so you'd have to cnofirm whether native Quit men item changes the app name to a localized version or keeps the same no matter what system language is used.

Also this function should return an Option like it used to be

you can also keep the current implementatio  as is and add a new function that calls this and converts the id to a string cause I think it might be need to have access to the pointer at some point.

src/platform_impl/macos/menu.rs
```diff
@@ -11,7 +11,8 @@ use objc::{
   declare::ClassDecl,
   runtime::{Class, Object, Sel, NO, YES},
 };
-use std::sync::Once;
+
```
amrbashir: nit-pick: remove extra line

.changes/macos-menus-appname.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": minor
+---
+
+Add the application name to the "Quit" and "Hide" native menu items.
```
amrbashir:
```suggestion
Add the application name to the "Quit" and "Hide" native menu items on macOS.
```

.changes/macos-menus-appname.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": minor
```
amrbashir:
```suggestion
"tao": patch
```
tao hasn't published a major releaae yet so only breaking changes are considerd minor

amrbashir (changes requested):

src/platform_impl/macos/util/mod.rs
caesar: > just took a glance at the relevant docs and seems like localizedName could be undefned

Oh, my bad, I'll look into that. I didn't think it could be undefined but I'll check it out.

> I don't use mac so you'd have to cnofirm whether native Quit men item changes the app name to a localized version or keeps the same no matter what system language is used.

Normal native behaviour is to display the localized name.

> Also this function should return an Option like it used to be

I changed it because the current implementation always returns a value, it can never be `None`, and I was changing the signature anyway… but if you're right that `localizedName` could be undefined it will indeed need to be an `Option`.

> you can also keep the current implementatio as is and add a new function that calls this and converts the id to a string cause I think it might be need to have access to the pointer at some point.

Ok, I can do that if you prefer.

src/platform_impl/macos/util/mod.rs
```diff
@@ -131,12 +131,11 @@ pub unsafe fn ns_string_to_rust(ns_string: id) -> String {
   string.to_owned()
 }
 
-#[allow(dead_code)] // In case we want to use this function in the future
+/// Gets the app's name from the `localizedName` property of `NSRunningApplication`
 pub unsafe fn app_name() -> Option<id> {
```
amrbashir, line 135: I meant you should add another utility function

```rs
unsafe fn app_name_str() -> Option<String> {
  app_name.map(ns_string_to_rust)
}
```

src/platform_impl/macos/menu.rs
```diff
@@ -194,24 +197,40 @@ impl Menu {
...
+          ),
+        ))
+      }
+      MenuItem::Hide => {
+        let label = unsafe {
+          match app_name() {
+            Some(app_name) => format!("Hide {}", ns_string_to_rust(app_name)),
+            None => "Hide".to_string(),
+          }
+        };
```
amrbashir:
```suggestion
        let label = format!("Quit {}", unsafe { app_name_str() }.unwrap_or_default());
```

amrbashir (changes requested):

src/platform_impl/macos/util/mod.rs
caesar: Ah, then I misunderstood your comment in Discord. Makes sense though.

caesar:
Thanks for all the reviews @amrbashir 🙂

amrbashir (approved):
Thank you

PR #573, opened 2022-09-26
feat(menus): add support for Plus key in accelerators, closes #227

### LINKED ISSUE #227: Accelerator support for virtual keys like `+`
A common keyboard shortcut is `CmdOrCtrl+Plus`, which tao currently doesn't support (at least on macOS, not sure about other platforms)

Returning `"+"` from `Accelerator::key_equivalent` on macOS and it gave me `⌘+` in the menu bar. On my keyboard, both `⌘=` and `⌘⇧+` trigger that shortcut (which is expected).

caesar:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Adds `KeyCode::Plus` which can also be used (eg in Tauri) with the string `"Plus"` such as in `"CmdOrCtrl+Plus"`.
As [noted in the added docs](https://github.com/tauri-apps/tao/pull/573/files#diff-792b8d0bfd0c836c48cce2af6cc7d60ec434486ba56de63ffbaa1a6d09d17246R326), behaviour is somewhat platform-dependent and (at least for US and similar keyboards) only works in menu accelerators.

A solution which also works with `Accelerator::matches` for app-wide keyboard shortcuts would be ideal, but the implementation would be very complex (if it's even feasible at all in a cross-platform way).

I'm not sure if this is a full resolution for #227 but using `+` in menus is the only case specifically discussed there. Any other missing values will have to be implemented on a case-by-case basis I think.

### DIFF
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -323,6 +323,8 @@ pub enum KeyCode {
   KeyZ,
   /// <kbd>-</kbd> on a US keyboard.
   Minus,
+  /// <kbd>+</kbd> on a US keyboard.
+  Plus,
   /// <kbd>.</kbd> on a US keyboard.
   Period,
   /// <kbd>'</kbd> on a US keyboard.
@@ -698,6 +700,7 @@ impl FromStr for KeyCode {
       "NUM9" | "NUMPAD9" => KeyCode::Numpad9,
       "=" => KeyCode::Equal,
       "-" => KeyCode::Minus,
+      "PLUS" => KeyCode::Plus,
       "." | "PERIOD" => KeyCode::Period,
       "'" | "QUOTE" => KeyCode::Quote,
       "\\" => KeyCode::IntlBackslash,
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -323,7 +323,14 @@ pub enum KeyCode {
   KeyZ,
   /// <kbd>-</kbd> on a US keyboard.
   Minus,
-  /// <kbd>+</kbd> on a US keyboard.
+  /// <kbd>Shift</kbd>+<kbd>=</kbd> on a US keyboard.
+  ///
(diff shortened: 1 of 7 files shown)


src/keyboard.rs
```diff
@@ -323,6 +323,15 @@ pub enum KeyCode {
...
   /// <kbd>-</kbd> on a US keyboard.
   Minus,
+  /// <kbd>Shift</kbd>+<kbd>=</kbd> on a US keyboard.
+  ///
+  /// When used as a menu accelerator this is displayed as <kbd>+</kbd>, and on macOS
+  /// and Windows the accelerator can be used without pressing the <kbd>Shift</kbd> key.
+  /// On Linux the <kbd>Shift</kbd> key is still required with a US keyboard layout.
+  ///
+  /// `Plus` does not work as a value for keyboard accelerators outside menus, on
+  /// US keyboards or others where <kbd>+</kbd> requires a modifier key.
```
amrbashir:
```suggestion
  ///
  /// `Plus` does not work as a value for keyboard accelerators outside menus where <kbd>+</kbd> requires a modifier key.
```

amrbashir (changes requested):

src/keyboard.rs
caesar: I'm not sure that quite captures the intent of what I meant to say – how about this?

```suggestion
  /// `Plus` does not work as a value for keyboard accelerators outside menus on
  /// keyboards where <kbd>+</kbd> requires a modifier key.
```

amrbashir: sounds good

amrbashir (approved):
Thanks

PR #574, opened 2022-09-29
feat: options to disable individual window controls, closes #116

### LINKED ISSUE #116: feat: add `minimizable` and `maximizable` options
Feature required where when alwaysOnTop is set to true, then even clicking on the app icon on the OS's task bar should not minimize the application and it stays always on top unless exited by the user.
Thanks, 
Donnie

caesar:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

#### Platform support
| API | macOS | Windows | Linux |
|-----|------|------|-----|
|`Window::set_closable`| ✔️ | ✔️ | ✔️ |
|`Window::is_closable`| ✔️ | ✔️ | ✔️ |
|`WindowBuilder::with_closable`| ✔️ | ✔️ | ✔️ |
|`Window::set_minimizable`| ✔️ | ✔️ | ❌ |
|`Window::is_minimizable`| ✔️ | ✔️ | ❌ |
|`WindowBuilder::with_minimizable`| ✔️ | ✔️ | ❌ |
|`Window::set_maximizable`| ✔️ | ✔️ | ❌ |
|`Window::is_maximizable`| ✔️ | ✔️ | ❌ |
|`WindowBuilder::with_maximizable`| ✔️ | ✔️ | ❌ |

I am not able to find any GTK APIs which would make it possible to implement the `minimizable` and `maximizable` APIs on Linux.

See #116 and https://github.com/tauri-apps/tauri/issues/2353, https://github.com/tauri-apps/tauri/issues/4286.


caesar:
> Maximizable / Zoomable / Fullscreenable could also be considered.

It should be noted that disabling maximize was requested in each of the linked issues.
However, a few thoughts…:

- Disabling resize already disables maximize. Would it ever make sense to disable maximize when the window can be resized to that size anyway?
- On macOS, maximize isn't (really) a thing. We would want to disable "zoom" which is *sometimes* equivalent to maximize but not always (app-dependent). In which case: *should a hypothetical `set_maximizable` also enable/disable zoom on macOS, or would a separate `set_zoomable` method be better?*
- On macOS, zoom and fullscreen share the same button. I *think* the only way to disable either of them is to directly disable the button, something like this:
  ```obj-c
  NSButton *button = [window standardWindowButton:NSWindowZoomButton];
  [button setEnabled: NO];
  ```

amrbashir:
> > Maximizable / Zoomable / Fullscreenable could also be considered.
> 
> It should be noted that disabling maximize was requested in each of the linked issues. However, a few thoughts…:
> 
> * Disabling resize already disables maximize. Would it ever make sense to disable maximize when the window can be resized to that size anyway?

Yeah, they can be separate from each other.

> * On macOS, maximize isn't (really) a thing. We would want to disable "zoom" which is _sometimes_ equivalent to maximize but not always (app-dependent). In which case: _should a hypothetical `set_maximizable` also enable/disable zoom on macOS, or would a separate `set_zoomable` method be better?_

`set_maximizable` is better to fit the terminology of other platforms and other methods like `set_maximized`

> * On macOS, zoom and fullscreen share the same button. I _think_ the only way to disable either of them is to directly disable the button, something like this:
>   ```objc
>   NSButton *button = [window standardWindowButton:NSWindowZoomButton];
>   [button setEnabled: NO];
>   ```

Sure, whatever works for macOS.

caesar:
> `set_maximizable` is better to fit the terminology of other platforms and other methods like `set_maximized`

Agreed, that probably makes the most sense, especially since Tao seems to implement "zoom" as if it was maximize anyway.

caesar:
Note: I added an [example for `closable`](https://github.com/tauri-apps/tao/pull/574/files#diff-a2e708bfc6e26b9e55c394363983e0bd42d3734e007016c34156ff8e052d2bda) and modified the [example for `minimize`](https://github.com/tauri-apps/tao/pull/574/files#diff-c3f2b5cf73dcb9e10f062c8254f27743f5166e42217bae25e26d4fc0aba074de) to also cover `minimizable`.
I hope that's ok.
I thought it unnecessary to also add an example for `maximizable` since it works identically to `minimizable`.

examples/closable.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2014-2021 The winit contributors
```
amrbashir: No need for a separate example, just use `window_debug.rs` example.

.changes/window-controls-disable.md
```diff
@@ -0,0 +1,9 @@
+---
+"tao": patch
+---
+
+Add APIs for disabling the individual window controls on desktop platforms:
+`Window::set_closable`, `Window::is_closable`, `WindowBuilder::with_closable`,
+`Window::set_minimizable`, `Window::is_minimizable`, `WindowBuilder::with_minimizable`,
+`Window::set_maximizable`, `Window::is_maximizable`, `WindowBuilder::with_maximizable`.
+See the docs for platform-specific notes, especially regarding Linux.
```
amrbashir:
```suggestion
Add APIs for disabling the individual window controls on desktop platforms, `Window::set_closable`, `Window::is_closable`, `WindowBuilder::with_closable`, `Window::set_minimizable`, `Window::is_minimizable`, `WindowBuilder::with_minimizable`, `Window::set_maximizable`, `Window::is_maximizable`, `WindowBuilder::with_maximizable`. See the docs for platform-specific notes, especially regarding Linux.
```

examples/minimize.rs
```diff
@@ -5,7 +5,7 @@
 extern crate tao;
```
amrbashir: use `window_debug.rs` example instead of modifying this one.

src/platform_impl/windows/window_state.rs
```diff
@@ -219,11 +223,17 @@ impl WindowFlags {
...
     let (mut style, mut style_ex) = (Default::default(), Default::default());
-    style |= WS_CLIPSIBLINGS | WS_CLIPCHILDREN | WS_SYSMENU | WS_CAPTION | WS_MINIMIZEBOX;
+    style |= WS_CLIPSIBLINGS | WS_CLIPCHILDREN | WS_SYSMENU | WS_CAPTION;
     style_ex |= WS_EX_ACCEPTFILES;
 
     if self.contains(WindowFlags::RESIZABLE) {
-      style |= WS_SIZEBOX | WS_MAXIMIZEBOX;
+      style |= WS_SIZEBOX;
+    }
+    if self.contains(WindowFlags::RESIZABLE | WindowFlags::MAXIMIZABLE) {
```
amrbashir, line 234:
```suggestion
    if self.contains(WindowFlags::MAXIMIZABLE) {
```

src/platform_impl/windows/window.rs
```diff
@@ -879,6 +939,8 @@ unsafe fn init<T: 'static>(
   window_flags.set(WindowFlags::TRANSPARENT, attributes.transparent);
   // WindowFlags::VISIBLE and MAXIMIZED are set down below after the window has been configured.
   window_flags.set(WindowFlags::RESIZABLE, attributes.resizable);
+  window_flags.set(WindowFlags::MINIMIZABLE, attributes.minimizable);
+  window_flags.set(WindowFlags::MAXIMIZABLE, attributes.maximizable);
```
amrbashir, line 943:
```suggestion
  window_flags.set(WindowFlags::MAXIMIZABLE, attributes.maximizable);
  window_flags.set(WindowFlags::CLOSABLE, attributes.closeable);
```

src/platform_impl/windows/window.rs
```diff
@@ -309,6 +309,44 @@ impl Window {
...
+    let mut enable = MF_BYCOMMAND;
+    if closable {
+      enable |= MF_ENABLED;
+    } else {
+      enable |= MF_DISABLED | MF_GRAYED;
+    }
+    unsafe {
+      let window_menu = GetSystemMenu(self.hwnd(), false);
+      EnableMenuItem(window_menu, SC_CLOSE, enable);
+    }
```
amrbashir: move this logic inside `window_state.rs` in `apply_diff`

src/platform_impl/windows/window.rs
```diff
@@ -480,6 +518,28 @@ impl Window {
...
+  }
+
+  #[inline]
+  pub fn is_closable(&self) -> bool {
+    let menu_state;
+    unsafe {
+      let window_menu = GetSystemMenu(self.hwnd(), false);
+      menu_state = GetMenuState(window_menu, SC_CLOSE, MF_BYCOMMAND);
+    }
+    (menu_state & (MF_DISABLED | MF_GRAYED).0) == 0
```
amrbashir: after moving the `set_closable` logic inside window_state. just check here for the `CLOSABLE` window flag.

src/window.rs
```diff
@@ -156,6 +156,21 @@ pub struct WindowAttributes {
...
+
+  /// Whether the window is maximizable or not.
+  ///
+  /// The default is `true`.
+  pub maximizable: bool,
+
+  /// Whether the window is closable or not.
+  ///
+  /// The default is `true`.
+  pub closable: bool,
```
amrbashir: Add `platform-specific` notes here as well or link to `set_closable` ...etc

src/window.rs
```diff
@@ -694,6 +745,41 @@ impl Window {
     self.window.set_resizable(resizable)
   }
 
+  /// Sets whether the window is minimizable or not.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** Unsupported.
+  /// - **iOS / Android:** Unsupported.
```
amrbashir:
```suggestion
  /// - **Linux / iOS / Android:** Unsupported.
```

src/window.rs
```diff
@@ -694,6 +745,41 @@ impl Window {
...
+    self.window.set_minimizable(minimizable)
+  }
+
+  /// Sets whether the window is maximizable or not.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **macOS:** Disables the "zoom" button in the window titlebar, which is also used to enter fullscreen mode.
+  /// - **Linux:** Unsupported.
+  /// - **iOS / Android:** Unsupported.
```
amrbashir:
```suggestion
  /// - **Linux / iOS / Android:** Unsupported.
```

src/window.rs
```diff
@@ -754,6 +840,38 @@ impl Window {
     self.window.is_resizable()
   }
 
+  /// Gets the window's current minimizable state.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** Unsupported.
+  /// - **iOS / Android:** Unsupported.
```
amrbashir:
```suggestion
  /// - **Linux / iOS / Android:** Unsupported.
```

src/window.rs
```diff
@@ -754,6 +840,38 @@ impl Window {
...
+  pub fn is_minimizable(&self) -> bool {
+    self.window.is_minimizable()
+  }
+
+  /// Gets the window's current maximizable state.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** Unsupported.
+  /// - **iOS / Android:** Unsupported.
```
amrbashir:
```suggestion
  /// - **Linux / iOS / Android:** Unsupported.
```

amrbashir (changes requested):
Great work! Thanks

src/window.rs
caesar: I didn't add these because none of the others have them – eg `set_resizable` has some quite extensive platform specific notes which aren't linked from here. But probably better to link anyway I guess 👍

examples/closable.rs
caesar: Oh that's much better! I hadn't noticed that one existed 😄

src/platform_impl/windows/window_state.rs
caesar: I don't think this is right – or at any rate, it's a change in behaviour from before. Previously, if `resizable` was disabled, the maximize button was also disabled. (On other platforms this is handled automatically by the window manager, but not on Windows). I have maintained that previous behaviour.

With your change, if `resazable` is disabled the mazimize button will still work unless `maximizable` is also disabled.
Is that what you intended?

caesar:
Thanks for the review @amrbashir. All the changes made a lot of sense except https://github.com/tauri-apps/tao/pull/574#discussion_r991188760 – please see my comment there. Is the change in behaviour intended? If so it will be different behaviour from other platforms.

src/platform_impl/windows/window_state.rs
amrbashir: yeah, previously, I coupled both together because we didn't have the option to disable maximize alone but now that we do, it is fine to separate them

amrbashir: lets keep it as is for now and maybe change it in the future if needed

caesar: Okay, yeah, I can see an argument both ways but I agree with keeping the existing behaviour for now.
(Also I guess changing it would technically be a breaking change – so `minor` instead of `patch`)

src/platform_impl/windows/window_state.rs
```diff
@@ -351,7 +361,17 @@ impl WindowFlags {
...
+      if new.contains(WindowFlags::CLOSABLE) {
+        close_item_flags |= MF_ENABLED;
+      } else {
+        close_item_flags |= MF_DISABLED | MF_GRAYED;
+      }
+
       unsafe {
+        let window_menu = GetSystemMenu(window, false);
+        EnableMenuItem(window_menu, SC_CLOSE, close_item_flags);
+
```
amrbashir:
```suggestion
    if diff.contains(WindowFlags::CLOSABLE) {
      unsafe {
        let system_menu = GetSystemMenu(window, false);
        EnableMenuItem(
          system_menu,
          SC_CLOSE,
          MF_BYCOMMAND
            | if new.contains(WindowFlags::CLOSABLE) {
              MF_ENABLED
            } else {
              MF_DISABLED | MF_GRAYED
            },
        );
      }
    }
```

And move it above L355

examples/window_debug.rs
```diff
@@ -139,6 +142,21 @@ fn main() {
           content_protection = !content_protection;
           window.set_content_protection(content_protection);
         }
+        "M" => {
+          let minimizable = !window.is_minimizable();
+          println!("Minimizable: {}", minimizable);
```
amrbashir: remove

examples/window_debug.rs
```diff
@@ -139,6 +142,21 @@ fn main() {
...
+          window.set_minimizable(minimizable);
+        }
+        "X" => {
+          let maximizable = !window.is_maximizable();
+          println!("Maximizable: {}", maximizable);
```
amrbashir: remove

examples/window_debug.rs
```diff
@@ -139,6 +142,21 @@ fn main() {
...
+          window.set_maximizable(maximizable);
+        }
+        "Q" => {
+          let closable = !window.is_closable();
+          println!("Closable: {}", closable);
```
amrbashir: remove

amrbashir (changes requested):
Last few nit-pick

src/platform_impl/windows/window_state.rs
caesar: I tried this but it doesn't work on window creation (with `WindowBuilder::with_closable(false)`) except by removing the `if diff.contains(WindowFlags::CLOSABLE)`.

examples/window_debug.rs
caesar: I'm confused, I thought you asked me to move the examples of usage of these new APIs to here. Do you want me to remove all the examples altogether, or do you want them somewhere else?

amrbashir: no just remove this println! call

caesar: Ah, got it! Ok, will do.

src/platform_impl/windows/window_state.rs
amrbashir: my bad, that's probably because of line 938 in `windows/window.rs`
it should be removed and instead call `window.set_closable()` later after creating the window at L1041

caesar: Ah, could be… I can't remember if I tried that or not. I'll try it now.

caesar: Still doesn't work properly… I'll play with it a bit more tomorrow

caesar: It works if I do this:
```rust
    if self == WindowFlags::empty() || diff.contains(WindowFlags::CLOSABLE) {
```
Would that be okay?

amrbashir: I pushed a fix, it needed to have a default so the diffing would work.

src/platform_impl/linux/window.rs
```diff
@@ -473,6 +474,23 @@ impl Window {
     }
   }
 
+  pub fn set_minimizable(&self, _minimizable: bool) {
+    warn!("`Window::set_minimizable` is ignored on Linux")
+  }
+
+  pub fn set_maximizable(&self, _maximizable: bool) {
+    warn!("`Window::set_maximizable` is ignored on Linux")
+  }
```
```suggestion
  pub fn set_minimizable(&self, _minimizable: bool) {}

  pub fn set_maximizable(&self, _maximizable: bool) {}
```

src/platform_impl/linux/window.rs
```diff
@@ -503,6 +521,20 @@ impl Window {
...
 
+  pub fn is_minimizable(&self) -> bool {
+    warn!("`Window::is_minimizable` is ignored on Linux");
+    false
+  }
+
+  pub fn is_maximizable(&self) -> bool {
+    warn!("`Window::is_maximizable` is ignored on Linux");
+    false
+  }
```
```suggestion
  pub fn is_minimizable(&self) -> bool {
    true
  }

  pub fn is_maximizable(&self) -> bool {
    true
  }
```

src/window.rs
```diff
@@ -156,6 +156,27 @@ pub struct WindowAttributes {
...
+  ///
+  /// See notes at [`Window::set_maximizable`].
+  pub maximizable: bool,
+
+  /// Whether the window is closable or not.
+  ///
+  /// The default is `true`.
+  ///
+  /// See notes at [`Window::set_closable`].
+  pub closable: bool,
```
amrbashir:
```suggestion
  /// See [`Window::set_minimizable`] for details.
  pub minimizable: bool,

  /// Whether the window is maximizable or not.
  ///
  /// The default is `true`.
  ///
  /// See [`Window::set_maximizable`] for details.
  pub maximizable: bool,

  /// Whether the window is closable or not.
  ///
  /// The default is `true`.
  ///
  /// See [`Window::set_closable`] for details.
  pub closable: bool,
```

amrbashir (changes requested):

amrbashir (approved):

caesar:
Thanks @amrbashir, that seems to work.

PR #586, opened 2022-10-11
feat(macos): add `allows_automatic_window_tabbing` APIs

caesar:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Adds the following APIs on macOS:
- `WindowExtMacOS::set_allows_automatic_window_tabbing`
- `WindowExtMacOS::allows_automatic_window_tabbing`
- `WindowBuilderExtMacOS::with_automatic_window_tabbing`

This allows disabling macOS's "automatic window tabbing" feature, whereby an undesired tab bar is added to the app and all the windows are merged into one.

See https://github.com/tauri-apps/tauri/issues/3912.

### DIFF
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -91,6 +91,7 @@ pub struct PlatformSpecificWindowBuilderAttributes {
   pub resize_increments: Option<LogicalSize<f64>>,
   pub disallow_hidpi: bool,
   pub has_shadow: bool,
+  pub automatic_tabbing: bool,
 }
 
 impl Default for PlatformSpecificWindowBuilderAttributes {
@@ -107,6 +108,7 @@ impl Default for PlatformSpecificWindowBuilderAttributes {
       resize_increments: None,
       disallow_hidpi: false,
       has_shadow: true,
+      automatic_tabbing: true,
     }
   }
 }
@@ -270,6 +272,10 @@ fn create_window(
 
       if let Parent::ChildOf(parent) = pl_attrs.parent {
         let _: () = msg_send![parent as id, addChildWindow: *ns_window ordered: NSWindowOrderingMode::NSWindowAbove];
+      }
+
+      if !pl_attrs.automatic_tabbing {
+        NSWindow::setAllowsAutomaticWindowTabbing_(*ns_window, NO);
       }
 
       if !pl_attrs.has_shadow {
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -1472,6 +1472,21 @@ impl WindowExtMacOS for UnownedWindow {
       is_document_edited == YES
     }
   }
+
+  #[inline]
+  fn set_allows_automatic_window_tabbing(&self, enabled: bool) {
+    unsafe {
+      NSWindow::setAllowsAutomaticWindowTabbing_(*self.ns_window, if enabled { YES } else { NO })
+    }
+  }
+
+  #[inline]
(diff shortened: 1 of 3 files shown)


src/platform_impl/macos/window.rs
```diff
@@ -272,6 +274,10 @@ fn create_window(
         let _: () = msg_send![parent as id, addChildWindow: *ns_window ordered: NSWindowOrderingMode::NSWindowAbove];
       }
 
+      if !pl_attrs.automatic_tabbing {
+        NSWindow::setAllowsAutomaticWindowTabbing_(*ns_window, NO);
```
amrbashir, line 278:
```suggestion
        ns_window.setAllowsAutomaticWindowTabbing_(NO);
```

src/platform_impl/macos/window.rs
```diff
@@ -1472,6 +1478,21 @@ impl WindowExtMacOS for UnownedWindow {
       is_document_edited == YES
     }
   }
+
+  #[inline]
+  fn set_allows_automatic_window_tabbing(&self, enabled: bool) {
+    unsafe {
+      NSWindow::setAllowsAutomaticWindowTabbing_(*self.ns_window, if enabled { YES } else { NO })
```
amrbashir, line 1485:
```suggestion
      self.ns_window.setAllowsAutomaticWindowTabbing_(if enabled { YES } else { NO })
```

src/platform_impl/macos/window.rs
```diff
@@ -1472,6 +1478,21 @@ impl WindowExtMacOS for UnownedWindow {
...
+    }
+  }
+
+  #[inline]
+  fn allows_automatic_window_tabbing(&self) -> bool {
+    unsafe {
+      let allows_tabbing: BOOL = NSWindow::allowsAutomaticWindowTabbing(*self.ns_window);
```
amrbashir, line 1492:
```suggestion
      let allows_tabbing: BOOL = self.ns_window.allowsAutomaticWindowTabbing();
```

amrbashir (changes requested):

src/platform_impl/macos/window.rs
caesar: This doesn't work, `setAllowsAutomaticWindowTabbing_` only works when called as a static method, not an instance method.

src/platform_impl/macos/window.rs
caesar: This doesn't work, `setAllowsAutomaticWindowTabbing_` only works when called as a static method, not an instance method.

src/platform_impl/macos/window.rs
caesar: This doesn't work, `allowsAutomaticWindowTabbing` only works when called as a static method, not an instance method.

src/platform_impl/macos/window.rs
amrbashir: yeah I was confused when I reviewed the PR, didn't think that it would require `Self` which seems like it might be a bug in `cocoa` but shouldn't concern this PR.

amrbashir (approved):
thank you

src/platform_impl/macos/window.rs
caesar: Yes I also thought it was a strange API, not sure why it wants `Self`since it doesn't use it...

PR #588, opened 2022-10-14
refactor(linux): reduce channel redirect

wusyong:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Reduce channel redirection. Sending user events and redraw request should be faster.
This can make dragging egui window very smoothly now.

### DIFF
--- a/src/platform_impl/linux/event_loop.rs
+++ b/src/platform_impl/linux/event_loop.rs
@@ -8,11 +8,11 @@ use std::{
   error::Error,
   process,
   rc::Rc,
-  sync::mpsc::SendError,
   time::Instant,
 };
 
 use cairo::{RectangleInt, Region};
+use crossbeam_channel::SendError;
 use gdk::{Cursor, CursorType, EventKey, EventMask, ScrollDirection, WindowEdge, WindowState};
 use gio::{prelude::*, Cancellable};
 use glib::{source::Priority, Continue, MainContext};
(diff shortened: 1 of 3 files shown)


src/platform_impl/linux/event_loop.rs
```diff
@@ -760,13 +755,6 @@ impl<T: 'static> EventLoop<T> {
               Inhibit(false)
             });
           }
-          WindowRequest::Redraw => {
-            if let Err(e) = draw_tx.send(id) {
-              log::warn!("Failed to send redraw event to event channel: {}", e);
-            }
-
-            window.queue_draw();
```
amrbashir, line 768: this is seems to be removed completely, is that desired?

wusyong: This is unintentional. But since we now understand `queue_draw` is more like a signal for gtk widgets to draw, I would like to think about if we still need this. `RequestRedraw` event is more for users to draw on GL library themselves IMHO.
I'll give my conclusion soon.

wusyong: Alright, I really think `queue_draw` could be removed. Calling `queue_draw` will invoke `connect_draw` and hence send another draw event. It doesn't make sense to send two draw events when calling `request_redraw`.
And for regular gtk widgets, they should handle all drawing processes properly.

amrbashir: alright, sounds good

amrbashir (approved):

PR #613, opened 2022-11-02
On macOS, add `WindowExtMacOS::reset_cursor_rect`

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/reset-cursor.md
+++ b/.changes/reset-cursor.md
@@ -1,5 +1,5 @@
 ---
-"tao"
+"tao": patch
 ---
 
 On macOS, add `WindowExtMacOS::reset_cursor_rect`.
--- a/.changes/reset-cursor.md
+++ b/.changes/reset-cursor.md
@@ -1,0 +1,5 @@
+---
+"tao"
+---
+
+On macOS, add `WindowExtMacOS::reset_cursor_rect`.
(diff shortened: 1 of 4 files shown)


amrbashir (commented):
I am not sure if this fix will be accepted upstream, how would this be used in WRY btw?

wusyong:
@amrbashir wry will call `window.reset_cursor_rect(false)` internally when creating webview.
I don't know if upstream would accept either. But this is the only I can think of without replacing TaoView.
Replacing TaoView will lose lost of WindowEvent including resize event.

amrbashir:
calling `window.reset_cursor_rect(false)` will effectively disable `window.set_cursor_icon()`?

wusyong:
Yes, this is `resetCursorRects:` is used for loading custom cursor in TaoView.
Wen `set_cursor_icon` is called, it will invoke `invalidateCursorRectsForView:` which will trigger `resetCursorRects:`

amrbashir:
I don't think users will appreciate disabling the `window.set_cursor_icon`, maybe `window.set_cursor_icon` could be changed to take `Option<CusrsorIcon>` and calling it with `None` variant will make tao/winit fallback to the system behavior.

wusyong:
Closed this in favor of #614

PR #615, opened 2022-11-07
feat(linux): use dirs instead of dirs-next

light4:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

deps: use [dirs](https://github.com/dirs-dev/dirs-rs) instead of [dirs-next](https://github.com/xdg-rs/dirs), seems like latter is unmaintained

- [x] No

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ targets = [
 [features]
 default = [ ]
 dox = [ "gtk/dox" ]
-tray = [ "libappindicator", "dirs-next" ]
+tray = [ "libappindicator", "dirs" ]
 
 [build-dependencies]
 cc = "1"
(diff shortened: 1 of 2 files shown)


amrbashir:
We use `dirs-next` because `dirs` was unmaintained, did the situation turn around? cc @lucasfernog

FabianLars:
It did indeed change a bit. But both projects seem to be pretty inactive nowadays imo :(

amrbashir:
Thanks for your contribution but after discussing it with Lucas, both crates seem to be on the same level of maintenance and so it is fine to keep using `dirs-next` until we absolutely have to change.

PR #644, opened 2022-12-13
Add `with_cursor_moved` unix extension method

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -76,7 +76,7 @@ pub trait WindowBuilderExtUnix {
   fn with_app_paintable(self, app_paintable: bool) -> WindowBuilder;
 
   /// Whether to set cursor moved event. Cursor event is suited for native GUI frameworks and
-  /// games. But it can block gtk's own pipeline ocationally. Turn this off can help Gtk looks
+  /// games. But it can block gtk's own pipeline occasionally. Turn this off can help Gtk looks
   /// smoother.
   ///
   /// Default is `true`.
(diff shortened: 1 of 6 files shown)


src/platform/unix.rs
```diff
@@ -74,6 +74,13 @@ pub trait WindowBuilderExtUnix {
   ///
   /// Default is `false` but is always `true` if [`WindowAttributes::transparent`](crate::window::WindowAttributes::transparent) is `true`
   fn with_app_paintable(self, app_paintable: bool) -> WindowBuilder;
+
+  /// Whether to set cursor moved event. Cursor event is suited for native GUI frameworks and
+  /// games. But it can block gtk's own pipeline ocationally. Turn this off can help Gtk looks
```
```suggestion
  /// games. But it can block gtk's own pipeline occasionally. Turn this off can help Gtk looks
```

src/platform_impl/linux/event_loop.rs
```diff
@@ -351,7 +351,10 @@ impl<T: 'static> EventLoop<T> {
...
             };
           }
-          WindowRequest::WireUpEvents { transparent } => {
+          WindowRequest::WireUpEvents {
+            transparent,
+            cursor_moved,
+          } => {
             window.add_events(
               EventMask::POINTER_MOTION_MASK
                 | EventMask::BUTTON1_MOTION_MASK
```
amrbashir, line 360: We could probably remove these Event masks instead and the event shouldn't be triggered at all

wusyong: Hmm I still got the event despite if I disable it

wusyong: Oh, nevermind I only disabled `BUTTON1_MOTION_MASK`. But there are other events like holding mouse button and then dragging can still trigger lots of events. I prefer we check it in connect signal callback for all.

amrbashir: sure

amrbashir (approved):

PR #654, opened 2022-12-23
feat: switch android_fn! to a proc_macro

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This is rewrite was due to happen soon and it allows us to:
1. handle `_` in package name and escape it properly, instead of relying on users to pass it escaped.
2. support variadic arguments to functions (encountered while implementing https://github.com/tauri-apps/wry/issues/804)

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -78,6 +78,10 @@
     "tao": {
       "path": "./",
       "manager": "rust"
+    },
+    "tao-macros": {
+      "path": "./tao-macros",
+      "manager": "rust"
     }
   }
 }
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -6,8 +6,8 @@ on:
     - cron: '0 0 * * *'
   push:
     paths:
-      - "Cargo.lock"
-      - "Cargo.toml"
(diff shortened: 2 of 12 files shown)


wusyong (changes requested):
LGTM, but you should add doc on those macro if we want to expose this.
And also add tao-macro to workspace and our CI config

wusyong:
Or we could make this into separate repo and crate then.

amrbashir:
> LGTM, but you should add doc on those macro if we want to expose this.
> And also add tao-macro to workspace and our CI config

will do, I was just looking for a first review for the macro code.

> Or we could make this into separate repo and crate then.

I'd rather keep this in the repo.

wusyong:
> I'd rather keep this in the repo.

Alright! We keep int this repo then. I'm not concerned about the implementation since it's fairly short but I treat proc macro as a different DSL. A comprehensive documentation is required.

wusyong (approved):

PR #701, opened 2023-02-15
fix(windows): backport default undecorated shadow to false (#690)

### LINKED ISSUE #690: Add PR patches to version 0.16
Tauri version 1.x will stick to tao v0.16.
There are some patches we would like to merge into it as well. Here are the patches we need:

### LINKED ISSUE #890: Expose the collection behavior on mac (always on top hides in mission control)
Expose the nswindowbehavior managed on macos.

Also, would be nice if "set_always_on_top" also set this flag so the window wouldn't disappear from mission control when you make it sticky. I would expect the behavior for "always on top" to match that of sticky notes on mac, which both float on top and participate in window management.

More info here:
https://developer.apple.com/documentation/appkit/nswindowcollectionbehavior/nswindowcollectionbehaviormanaged

Here's the code:

```rust
        // https://developer.apple.com/documentation/appkit/nswindowcollectionbehavior/nswindowcollectionbehaviormanaged
        #[cfg(target_os = "macos")]
        {
            use cocoa::appkit::NSWindowCollectionBehavior;
            use cocoa::base::id;
            use objc::{msg_send, sel, sel_impl};
            use tao::platform::macos::WindowExtMacOS;

            unsafe {
                let window: id = window.ns_window() as id;
                let _: () = msg_send![window, setCollectionBehavior: NSWindowCollectionBehavior::NSWindowCollectionBehaviorManaged];
            }
        }
```

WolfspiritM:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #690

### Other information
This is a backport of an already fixed issue to tao 0.16. The reason this is needed is that currently (per tauri dev branch) transparent doesn't work on windows cause it seems like shadows are breaking transparent similar to decorations.

(I tried cherry picking but that won't really work with commit signing)

### DIFF
--- a/.changes/shadow.md
+++ b/.changes/shadow.md
@@ -1,0 +1,5 @@
+---
+"tao": "patch"
+---
+
+Fix undecorated window shadow enabled by default on Windows.
--- a/src/platform_impl/windows/mod.rs
+++ b/src/platform_impl/windows/mod.rs
@@ -63,7 +63,7 @@ impl Default for PlatformSpecificWindowBuilderAttributes {
       drag_and_drop: true,
       preferred_theme: None,
       skip_taskbar: false,
-      decoration_shadow: true,
+      decoration_shadow: false,
     }
   }
 }


amrbashir (approved):

amrbashir:
Thank you

PR #705, opened 2023-02-21
fix: don't panic if primary monitor not discoverable.

### LINKED ISSUE #704: `EventLoopWindowTarget::primary_monitor` can panic on linux
**Describe the bug**
`EventLoopWindowTarget::primary_monitor` is documented as returning None if it can't find the primary monitor.  In fact all platforms (?) return `Some()`, but linux does so by calling `unwrap()` on the underlying call to `gdk::Display::primary_monitor()`, which *can* return None.  Thus rather than returning None, tao panics.

**Steps To Reproduce**
I stumbled across this on hyprland; `get_primary_monitor` succeeds under weston.  So I think the cases this fails in must be pretty rare, but there's still the workaround of using `available_monitors` and picking one.

**Expected behavior**
should return `Some()` or `None`

OS: Linux
Rustc: rustc 1.69.0-nightly (7aa413d59 2023-02-19)

- [x] Yes

**Additional context**

2e0byo:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #704 

### DIFF
--- a/.changes/linux-primary-monitor.md
+++ b/.changes/linux-primary-monitor.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Attempt to get primary monitor on linux will now return None rather than panicking if monitor not found.
--- a/src/platform_impl/linux/event_loop.rs
+++ b/src/platform_impl/linux/event_loop.rs
@@ -79,9 +79,11 @@ impl<T> EventLoopWindowTarget<T> {
 
   #[inline]
   pub fn primary_monitor(&self) -> Option<RootMonitorHandle> {
-    let monitor = self.display.primary_monitor().unwrap();
-    let handle = MonitorHandle { monitor };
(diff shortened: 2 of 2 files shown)


amrbashir (approved):

amrbashir (changes requested):
Could you also add a change file?

see https://github.com/tauri-apps/tao/blob/dev/.changes/linux-arrow-cursor.md for an example.

wusyong (approved):

2e0byo:
I confess I don't understand the covector ci fail.  What should be in the changes file?

FabianLars:
it's missing some `-`. There is only `--` while it must be `---` like this
```
---
"tao": patch
---

```

Issue #708, opened 2023-03-02
`window.cursor_position()` is actually `LogicalPosition`


2e0byo:
**Describe the bug**
`Window.cursor_position()` returns a `PhysicalPosition` which has to be multiplied by the scale factor to get the same value as the `CursorMoved` event.  Thus `Window.cursor_position()` is a different value from `CursorMoved` if scaling is in use.

This occurs at least on Hyprland on linux.

`CursorMoved` explicitly handles the conversion to PhysicalPosition [here](https://github.com/tauri-apps/tao/blob/28b53f80c49bbf2ae8902b98a2e28f6451a5a8f1/src/platform_impl/linux/event_loop.rs#L580).

**Steps to reproduce**
I don't have a MWE but could probably get one together.  This was discovered in EmacsNG (too large to compile just to reproduce).  Credit to @declantsien for working out what the problem was.

**Expected behavior**
Positions should be identical.

**Platform and Versions (please complete the following information):**
OS: Linux 6.2.1-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 26 Feb 2023 03:39:23 +0000 x86_64 GNU/Linux
Rustc: rustc 1.69.0-nightly (7aa413d59 2023-02-19)

**Would you want to assign yourself to resolve this bug?**
- [x] Yes
- [ ] No

**Additional context**
I don't know if this occurs on other platforms. Windows makes an [os call](https://github.com/tauri-apps/tao/blob/28b53f80c49bbf2ae8902b98a2e28f6451a5a8f1/src/platform_impl/windows/util.rs#L235-L244) as does [Mac](https://github.com/tauri-apps/tao/blob/28b53f80c49bbf2ae8902b98a2e28f6451a5a8f1/src/platform_impl/macos/util/mod.rs#L127-L132).  But linux goes via 
the [gdk Display](https://github.com/tauri-apps/tao/blob/28b53f80c49bbf2ae8902b98a2e28f6451a5a8f1/src/platform_impl/macos/util/mod.rs#L127-L132).  I'm guessing the seat returns a scaled position as I can't find anything in the gdk docs.

wusyong:
I think I need an MWE to see if that's actually happen in tao. I thought GDK should handles that already.

wusyong:
Okay, I tested in #711 and looks like they are wrong indeed. They are actually logical positions.

2e0byo:
great, thanks!  Sorry, I'm snowed under and hadn't got round to this.

PR #711, opened 2023-03-07
fix(macOS): `cursor_position` returns incorrect position

### LINKED ISSUE #707: `monitor::cursor_position()` not displaying correct cursor position
**Describe the bug**
Hi, I have two monitors stacked vertically on macOS.

```
🖥️   <-- cursor here where should be -1440 < y < 0
🖥️
```

However, when I have the cursor on the top monitor, I'm seeing a positive value which seems to indicate a distance from the bottom edge of the bottom monitor instead of from where the monitor starting axis is `(0, 0)`.

**Steps To Reproduce**

```rs
let cursor = PhysicalPosition { x: 397.55078125, y: 2035.25 }; // Cursor here is positioned above the primary monitor, and y should be negative
assert_eq!(cursor, window.cursor_position());
assert_eq!(window.monitor_from_point(cursor.x, cursor.y), None);
```

Then, I'm not able to find `window.monitor_from_point(cursor.x, cursor.y)` because here are my monitors:

```rs
let monitors = [
  MonitorHandle { inner: MonitorHandle { name: Some("Monitor #..."), native_identifier: 2, size: PhysicalSize { width: 2560, height: 1440 }, position: PhysicalPosition { x: 0, y: 0 }, scale_factor: 1.0 } },
  MonitorHandle { inner: MonitorHandle { name: Some("Monitor #..."), native_identifier: 5, size: PhysicalSize { width: 2560, height: 1440 }, position: PhysicalPosition { x: 0, y: -1440 }, scale_factor: 1.0 } },
];
assert_eq!(monitors, window.available_monitors());
```

**Expected behavior**
In my example, the cursor `y` position should be `-1440 < y < 0`. The cursor position appears to use a different `PhysicalPosition` axis than the position calculation for monitors.

OS: `macOS 13.2 (22D49)`
Rustc: `rustc 1.61.0 (fe5b13d68 2022-05-18)`
tao: `"0.18"`

**Workaround**

[This snippet from `rustdesk`](https://github.com/rustdesk/rustdesk/blob/8b3c4d3d642df5e7dd346294ac058a79753e06dc/src/platform/macos.rs#L242-L257) appears to be a decent workaround.

wusyong:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #707

### DIFF
--- a/.changes/cursor.md
+++ b/.changes/cursor.md
@@ -1,0 +1,6 @@
+---
+"tao": patch
+---
+
+On macOS, Fix `cursor_position` return incorrect position.
+
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -20,10 +20,9 @@ fn main() {
       .unwrap(),
   );
 
-  event_loop.run(move |event, e, control_flow| {
+  event_loop.run(move |event, _, control_flow| {
     *control_flow = ControlFlow::Wait;
-    let cursor = e.cursor_position().unwrap();
-    println!("{:?}", cursor);
+    println!("{:?}", event);
 
     match event {
       Event::WindowEvent {
@@ -43,9 +42,7 @@ fn main() {
       }
       Event::MainEventsCleared => {
         if let Some(w) = &window {
-          println!("{:?}", w.cursor_position());
-          println!("{:?}", w.monitor_from_point(cursor.x, cursor.y));
-          // w.request_redraw();
+          w.request_redraw();
         }
       }
       _ => (),
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -20,9 +20,10 @@ fn main() {
       .unwrap(),
   );
 
-  event_loop.run(move |event, _, control_flow| {
+  event_loop.run(move |event, e, control_flow| {
     *control_flow = ControlFlow::Wait;
-    println!("{:?}", event);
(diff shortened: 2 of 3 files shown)


wusyong:
@amrbashir I've test it and the coordinates should be correct now.
But I found #708 is true, all platforms return logical positions. We'll need to discuss how to fix this.

amrbashir:
what do you mean by logical exactly? not relative to the window top-left corner? because `CursorMoved` is relative to the window top-left corner, while `window.cursor_position()` and `event_loop.cursor_position()` get the cursor position in the global space.

wusyong:
I mean `window.cursor_position()` and `event_loop.cursor_position()` will return logical position. I have a monitor with physical size 2880x1800 and with scale factor 2. Both `current_position()` will return `1440x900` if I point to the bottom right of the screen. I can test further on Linux and Windows later, but they don't convert with scale factor when returning either.

Anyway, this discussion should move to #708. This PR should fix correct coordinate in second monitor (excluding scale factor).

amrbashir (approved):

PR #712, opened 2023-03-07
fix: scale cursor_position

### LINKED ISSUE #708: `window.cursor_position()` is actually `LogicalPosition`

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [X] This PR will resolve #708
- [X] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tao/blob/dev/.changes/readme.md).
- [X] It can be built on all targets and pass CI/CD.

### DIFF
--- a/.changes/cursor-position-scale.md
+++ b/.changes/cursor-position-scale.md
@@ -1,0 +1,5 @@
+---
+"tao": "patch"
+---
+
+Fix `Window::cursor_position` and `EventLoopWindowTarget::cursor_position` scale on Linux and macOS.
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -248,7 +248,7 @@ impl<T> EventLoopWindowTarget<T> {
   ///
   /// ## Platform-specific
   ///
-  /// - **iOS / Android**: Unsupported.
+  /// - **iOS / Android / Linux(Wayland)**: Unsupported, returns `0,0`.
   #[inline]
   pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, ExternalError> {
     self.p.cursor_position()
--- a/src/platform_impl/linux/event_loop.rs
+++ b/src/platform_impl/linux/event_loop.rs
@@ -108,7 +108,7 @@ impl<T> EventLoopWindowTarget<T> {
 
   #[inline]
   pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, ExternalError> {
-    util::cursor_position()
+    util::cursor_position(self.is_wayland())
   }
 }
 
--- a/src/platform_impl/linux/util.rs
+++ b/src/platform_impl/linux/util.rs
@@ -6,20 +6,24 @@ use crate::{
 };
 
 #[inline]
-pub fn cursor_position() -> Result<PhysicalPosition<f64>, ExternalError> {
-  Display::default()
-    .map(|d| {
-      (
-        d.default_seat().and_then(|s| s.pointer()),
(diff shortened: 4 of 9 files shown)


amrbashir:
@wusyong I fixed the linux impl, feel free to take macOS. Windows should be working fine iirc but I will test later today once I switch back to windows.

wusyong:
@amrbashir I tested with `event_loop.cursor_position()` and it'll return error.
And I feel like this is tricky when two monitors have different scale factors. Not sure if this could be the case.

Edit: Linux should be easier because I believe it's all one display one screen now. They should share same scale factor.

amrbashir:
> I tested with `event_loop.cursor_position()` and it'll return error.

I haven't been able to get it to error at all, using one screen or two.

> And I feel like this is tricky when two monitors have different scale factors. Not sure if this could be the case.
> Edit: Linux should be easier because I believe it's all one display one screen now. They should share same scale factor.

yeah they share the same scale_factor at least on my KDE.

amrbashir:
just tested Windows and it works as expected with single and multi-monitor setup

wusyong:
macOS added

amrbashir:
can't approve my own PR so feel free to approve and merge

wusyong:
but where's Windows?

amrbashir:
Windows works fine

amrbashir:
Tested it yesterday

wusyong (approved):
I also tested on Windows myself and you are right.

PR #719, opened 2023-04-03
fix(windows): Use SetForegroundWindow before focus hack

FabianLars:
This fixes an issue we had with interactive notifications and "deep links", see https://github.com/FabianLars/tauri-plugin-deep-link/issues/26#issuecomment-1493671121 / https://github.com/FabianLars/tauri-plugin-deep-link/pull/27#issuecomment-1493784405

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### Other information
Small discord conversation: https://discord.com/channels/616186924390023171/807549941936816148/1092384046912380978

### DIFF
--- a/.changes/fix-windows-focus.md
+++ b/.changes/fix-windows-focus.md
@@ -1,0 +1,5 @@
+---
+'tao': 'patch'
+---
+
+Fix set_focus not working on Windows in some situations like interactive notifications.
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -1266,10 +1266,13 @@ unsafe fn taskbar_mark_fullscreen(handle: HWND, fullscreen: bool) {
 }
 
 unsafe fn force_window_active(handle: HWND) {
-  // In some situation, calling SetForegroundWindow could not bring up the window,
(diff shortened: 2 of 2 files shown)


amrbashir (approved):

RandomEngy:
Thanks for getting the fix in here!

I think that original focus hack should probably go away. It seems like Windows has already patched it. Trying to steal foreground focus is always going to be a dangerous and brittle game. Windows is going to always be fighting against apps that try to take focus without permission, so we should push devs to go with the `AllowSetForegroundWindow` approach instead in single-instance scenarios.

https://devblogs.microsoft.com/oldnewthing/20090220-00/?p=19083

FabianLars:
I agree we should keep an eye on it, but removing it prematurely may be the wrong call right now. The hack was not added with single instance applications in mind so it likely has use-cases we're not thinking about right now.

RandomEngy:
Fair enough, though maybe we should pay attention to the output of the first `SetForegroundWindow` and skip the hack code if it succeeds.

PR #722, opened 2023-04-08
fux: avoid running the focus hack if the first focus attempt succeedes

RandomEngy:
We don't need to send the keystrokes or call `SetForegroundWindow` a second time if the first try succeeded.

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Optimization

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -1267,7 +1267,10 @@ unsafe fn taskbar_mark_fullscreen(handle: HWND, fullscreen: bool) {
 
 unsafe fn force_window_active(handle: HWND) {
   // Try to focus the window without the hack first.
-  SetForegroundWindow(handle);
+  let initial_try_succeeded = SetForegroundWindow(handle).as_bool();
+  if initial_try_succeeded {
+    return;
+  }
 
   // In some situations, calling SetForegroundWindow could not bring up the window,
   // This is a little hack which can "steal" the foreground window permission.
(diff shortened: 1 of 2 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -1267,7 +1267,10 @@ unsafe fn taskbar_mark_fullscreen(handle: HWND, fullscreen: bool) {
 
 unsafe fn force_window_active(handle: HWND) {
   // Try to focus the window without the hack first.
-  SetForegroundWindow(handle);
+  let initial_try_succeeded = SetForegroundWindow(handle).as_bool();
+  if initial_try_succeeded {
```
```suggestion
  if SetForegroundWindow(handle).as_bool() {
```

amrbashir (changes requested):

FabianLars:
I wonder if windows will falsely report success in some cases 😅 Wouldn't surprise me with its track record so far but we'll see.

RandomEngy:
Windows has been getting more stringent in when it allows apps to take foreground focus. I expect that the effectiveness of that hack code is going to dwindle over time; so even if `SetForegroundWindow` is reporting incorrectly it won't make a difference.

This is the conclusion I came to after trying 7 different variants of focus hacks posted on the internet; Windows had patched them all. You have to go through official channels to get focus.

amrbashir (approved):

PR #732, opened 2023-05-09
feat: enable shadows by default for undecorated window on Windows

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

### Checklist

### DIFF
--- a/.changes/windows-undecorated-shadow.md
+++ b/.changes/windows-undecorated-shadow.md
@@ -1,0 +1,5 @@
+---
+"tao": "minor"
+---
+
+Enable shadows by default for undecorated window on Windows.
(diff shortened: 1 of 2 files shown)


wusyong (approved):

.changes/windows-undecorated-shadow.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": "minor"
```
wusyong, line 2: I think we can treat this as bug and just be a patch.

amrbashir: it is not exactly a bug, it was the normal/intended behavior on Windows for so long and so I'd consider this a breaking change and so we need to make it `minor`

PR #742, opened 2023-05-30
feat: add support for progress bar state, closes #418

### LINKED ISSUE #418: [feat] Window Progress Bar
> This issue has been upstreamed from https://github.com/tauri-apps/tauri/issues/4208 

### Describe the problem

It's a shame that Tauri doesn't have an API similar to [Electron's Progress Bar](https://www.electronjs.org/docs/latest/tutorial/progress-bar), it's very useful for showing progress in minimized windows.

An API similar to the Electron one, or even a Tao method that we could invoke in Rust code to set the progress quickly.

### Alternatives considered

_No response_

### Additional context

_No response_

ahqsoftwares:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will fix #418 

### Additional Context
- The Linux implementation might not work on some distros considering the diverse desktop environments and distros (eg, linux mint).


ahqsoftwares:
image
image
image

Tested to work on the following platforms
 - Windows 7
 - Windows 10
 - Windows 11
 - Ubuntu 22.04

Exact Repo: https://github.com/ahqsoftwares/tauri-test

wusyong:
Since this is a windows only feature, could you move it to platform-specific trait [`WindowExtWindows`](https://docs.rs/tao/latest/i686-pc-windows-msvc/tao/platform/windows/trait.WindowExtWindows.html)?

ahqsoftwares:
> Since this is a windows only feature, could you move it to platform-specific trait [`WindowExtWindows`](https://docs.rs/tao/latest/i686-pc-windows-msvc/tao/platform/windows/trait.WindowExtWindows.html)?

I am actually forking raw window handle to extend its support for linux progressbars

wusyong:
Alright, let me know if you want to get reviewed again.

ahqsoftwares:
Researched a while and found that for linux like ubuntu, need to setup unity path protocol & progressbars won't work until they are installed or appImaged. Any way tao can get the unity protocol thing?

ahqsoftwares:
image
Partly Working...

ahqsoftwares:
@wusyong 
Done ✅
It has been successfully implemented for Windows & Linux (including Unity)

ahqsoftwares:
# Following are the implementations (Outdated Impl Ahead as of 18-Jun-2023)

## Windows
```rs
  let win: &tao::window::Window = window.as_ref().unwrap();

  win.set_taskbar_progress_state(
    TaskbarProgressState::Normal
  );
  win.set_taskbar_progress(50, 100);
```

## Linux
```rs
  let win: &tao::window::Window = window.as_ref().unwrap();

  win.set_taskbar_progress_state(
    TaskbarProgressState::Normal,
    Some("application://tao-test.desktop")
  ); //Some(x), where x is the deb .desktop file name

  win.set_taskbar_progress(50, 100, Some("application://tao-test.desktop")); //Some(x), where x is the deb .desktop file name
```

Cargo.toml
```diff
@@ -120,7 +120,10 @@ gdkwayland-sys = "0.16.0"
 gdk-pixbuf = "0.16"
 libappindicator = { version = "0.8.0", optional = true }
 dirs-next = { version = "2.0.0", optional = true }
+lazy_static = "1"
```
amrbashir: lazy_static is already a dependency

src/window.rs
```diff
@@ -838,6 +838,63 @@ impl Window {
     self.window.set_maximizable(maximizable)
   }
 
+  /// Set the taskbar progress
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** May work, might not work
```
amrbashir: you should rather explain why it won't work? like what DE's are supported

src/window.rs
```diff
@@ -838,6 +838,63 @@ impl Window {
...
+      _ => None,
+    };
+    self.window.set_taskbar_progress(current, total, uri);
+  }
+
+  /// Set the taskbar progress
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** May work, might not work
```
amrbashir: ditto

src/window.rs
```diff
@@ -838,6 +838,63 @@ impl Window {
...
+  /// - **macOS/Android/iOS:** Unimplemented
+  #[cfg(windows)]
+  pub fn set_taskbar_progress(&self, current: u64, total: u64) {
+    self.window.set_taskbar_progress(current, total);
+  }
+
+  #[cfg(target_os = "linux")]
+  pub fn set_taskbar_progress(&self, current: u64, total: u64, unity_uri: Option<&str>) {
```
amrbashir: you should not have methods with different signature based on the platform here

src/window.rs
```diff
@@ -838,6 +838,63 @@ impl Window {
...
+        TaskbarProgressState::Normal => 2,
+        TaskbarProgressState::Paused => 4,
+      }
+    };
+
+    self.window.set_taskbar_progress_state(taskbar_state);
+  }
+
+  #[cfg(target_os = "linux")]
+  pub fn set_taskbar_progress_state(&self, state: TaskbarProgressState, unity_uri: Option<&str>) {
```
amrbashir: ditto

src/window.rs
```diff
@@ -838,6 +838,63 @@ impl Window {
     self.window.set_maximizable(maximizable)
   }
 
+  /// Set the taskbar progress
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** May work, might not work
+  /// - **macOS/Android/iOS:** Unimplemented
```
amrbashir:
```suggestion
  /// - **macOS:** Unimplemented
  /// - **Android / iOS:** Unsupported
```

src/window.rs
```diff
@@ -838,6 +838,63 @@ impl Window {
...
+    };
+    self.window.set_taskbar_progress(current, total, uri);
+  }
+
+  /// Set the taskbar progress
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** May work, might not work
+  /// - **macOS/Android/iOS:** Unimplemented
```
amrbashir:
```suggestion
  /// - **macOS:** Unimplemented
  /// - **Android / iOS:** Unsupported
```

src/platform_impl/linux/event_loop.rs
```diff
@@ -39,10 +39,14 @@ use crate::{
...
   keyboard,
   monitor::{self, MonitorHandle},
-  util,
+  taskbar, util,
   window::{WindowId, WindowRequest},
 };
 
+use taskbar::TaskbarIndicator;
+
+static mut TASKBAR: Option<TaskbarIndicator> = None;
```
amrbashir: I think this could be stored on `EventLoop` type instead of a static variable

src/platform_impl/linux/taskbar/unity.rs
```diff
@@ -0,0 +1,113 @@
+use dbus::{
```
amrbashir: I don't quite like this implementation to be honest since it depends on dbus and I think we can get the same effect using dlopen/dlsym similar to electron implementation here https://github.com/electron/electron/blob/905aad9cb6007c2e0ff4897ec64f0c9967f63ee2/shell/browser/linux/unity_service.cc#L134

Cargo.toml
```diff
@@ -120,7 +120,10 @@ gdkwayland-sys = "0.16.0"
...
 x11-dl = "2.21"
+x11rb = { version = "0.12.0",  features = ["allow-unsafe-code", "dl-libxcb"] }
```
amrbashir: this dependency is not needed, x11-dl is enough

src/platform_impl/linux/taskbar/xapps.rs
```diff
@@ -0,0 +1,162 @@
...
+pub struct Manager {
+  handle: (RawWindowHandle, RawDisplayHandle),
+  progress: u32,
+  progress_visible: bool,
+  pulse: bool,
+}
+
+impl Manager {
+  pub fn new(handle: RawWindowHandle, d_handle: RawDisplayHandle) -> Option<Self> {
+    if !matches!(handle, RawWindowHandle::Xlib(_) | RawWindowHandle::Xcb(_)) {
```
amrbashir: you should check for xlib handle only since that is the only one we create, we never create xcb handle

src/platform_impl/linux/taskbar/xapps.rs
```diff
@@ -0,0 +1,162 @@
...
+      progress_visible: false,
+      pulse: false,
+    })
+  }
+
+  pub fn update_property(&self, name: &CStr, value: u32) {
+    match self.handle {
+      (
+        RawWindowHandle::Xlib(XlibWindowHandle { window, .. }),
+        RawDisplayHandle::Xlib(XlibDisplayHandle { display, .. }),
```
amrbashir: ditto, xlib only

src/platform_impl/linux/taskbar/xapps.rs
```diff
@@ -0,0 +1,162 @@
...
+      }
+      _ => unreachable!(),
+    }
+  }
+
+  pub fn delete_property(&self, name: &CStr) {
+    match self.handle {
+      (
+        RawWindowHandle::Xlib(XlibWindowHandle { window, .. }),
+        RawDisplayHandle::Xlib(XlibDisplayHandle { display, .. }),
```
amrbashir: ditto, xlib only

src/platform_impl/linux/taskbar/xapps.rs
```diff
@@ -0,0 +1,162 @@
...
+};
+use x11_dl::xlib::{self, PropModeReplace, Xlib};
+use x11rb::{
+  protocol::xproto::{AtomEnum, ConnectionExt, PropMode},
+  wrapper::ConnectionExt as WrapperConnectionExt,
+  xcb_ffi::XCBConnection,
+};
+
+lazy_static! {
+  static ref XLIB: Xlib = Xlib::open().unwrap();
```
amrbashir: don't unwrap here, instead, make this type an Option or a Result

src/platform_impl/linux/taskbar/xapps.rs
```diff
@@ -0,0 +1,162 @@
+use std::ffi::CStr;
```
amrbashir: is this implementation tested? which distro can I test this on?

amrbashir (changes requested):

ahqsoftwares:
Implementing the others when I reach home

src/window.rs
ahqsoftwares: ?

amrbashir: you should not have methods with different signature based on the platform here

ahqsoftwares: You mean to add it in **WindowExtWindows** and the other one for the linux one?

amrbashir: yup

src/platform_impl/linux/taskbar/xapps.rs
ahqsoftwares: Not tested but tested to have no panics due to this. It is basically the leftover code when I thought that xapps is what all distros use instead of unity protocol (used by many distros). But should work if some distro uses the xapps protocol (I need to research on that as I am a windows 10 & 11 pc user only)

I just left it as is in case some distro use it.
My knowledge in linux is pretty limited

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: hmm...
Any specific reason for not preferring dbus actually?

amrbashir: mainly trying to avoid another dependency but if we must use dbus, we should use `zbus` crate

ahqsoftwares: I am actually asking because to be honest, I am a full newbie at linux apis

amrbashir: I will see if I can find sometime to switch to my linux and refine the Linux implementation.

ahqsoftwares:
Thanks

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: Thanks, I shall also try to port it to zbus or dlopen or dlsym using my wsl and ubuntu vm

src/platform_impl/linux/event_loop.rs
ahqsoftwares: But then I guess the window events listener cant read it, though can make a eventloop method maybe?

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: @amrbashir Shall I change it to zbus?

amrbashir: yeah sure, go ahead

src/platform_impl/linux/event_loop.rs
amrbashir: which event listener?

ahqsoftwares: This part 
https://github.com/ahqsoftwares/tao/blob/e232c00378659c3c9a0e7b88b9b5cd4baf47f31f/src/platform_impl/linux/event_loop.rs#L315-L378

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: Actually it seems like dbus is smaller than zbus (about 43kb bigger)
Shall we keep it dbus? or shall I shift to zbus?

ahqsoftwares: @amrbashir Shall it use dbus?

MegaSa1nt:
Hi! How can i use it in my project?
Every time i try to add it to dependencies, it fails to build with this error:
```
error: failed to select a version for `gdk-sys`.
    ... required by package `webkit2gtk v0.18.2`
    ... which satisfies dependency `webkit2gtk = "^0.18.2"` (locked to 0.18.2) of package `tauri v1.3.0`
    ... which satisfies dependency `tauri = "^1.3"` (locked to 1.3.0) of package `gcs v1.3.3 (E:\GreenCatsServer\sa1nt.gcs\gcs\src-tauri)`
versions that meet the requirements `^0.15.0` (locked to 0.15.1) are: 0.15.1

the package `gdk-sys` links to the native library `gdk-3`, but it conflicts with a previous package which links to `gdk-3` as well:
package `gdk-sys v0.16.0`
    ... which satisfies dependency `gdk = "^0.16"` of package `gdkwayland-sys v0.16.0`
    ... which satisfies dependency `gdkwayland-sys = "^0.16.0"` of package `tao v0.20.0 (https://github.com/ahqsoftwares/tao.git?branch=dev#5d2a6163)`
    ... which satisfies git dependency `tao` of package `gcs v1.3.3 (E:\GreenCatsServer\sa1nt.gcs\gcs\src-tauri)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the links ='gdk-sys' value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `gdk-sys` which could resolve this conflict
       Error failed to build app: failed to build app
```

If i try to add it to [patch.crates-io], it just ignores it and doesnt work:
```
warning: Patch `tao v0.20.0 (https://github.com/ahqsoftwares/tao.git?branch=dev#5d2a6163)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

ahqsoftwares:
This version of tao should not be compatible with tauri unless implemented
into the crate directly

ahqsoftwares:
@MegaSa1nt Tauri wont support a different tao due to cargo lock restrictions, using tao directly only would work rn

MegaSa1nt:
> @MegaSa1nt Tauri wont support a different tao due to cargo lock restrictions, using tao directly only would work rn

How can i use it directly? *Sorry, i'm new to Tauri*

ahqsoftwares:
> > @MegaSa1nt Tauri wont support a different tao due to cargo lock restrictions, using tao directly only would work rn
> 
> How can i use it directly? *Sorry, i'm new to Tauri*

You will have to fork wry and tauri and change the git url?
(I am not much sure about using custom repos in packages)

ahqsoftwares:
Also, u can implement using the code that I had used here without changing anything

wusyong:
I feel like we should drop Xlib implementation.
We simply couldn't cover all cases. Even we support xlib, there will be wayland left behind.
I hope we can use libunity only and we mention this as requirement in documentation.

pewsheen:
Btw, I have made a demo app for Unix in case someone wants to try it.

https://github.com/pewsheen/taotao

ahqsoftwares:
> I feel like we should drop Xlib implementation. We simply couldn't cover all cases. Even we support xlib, there will be wayland left behind. I hope we can use libunity only and we mention this as requirement in documentation.

True, idk myself which distro uses XLib & Wayland

src/platform_impl/linux/event_loop.rs
amrbashir: you can store it in `Rc<RefCell>>` or `Rc<Mutex>` and pass it to the handler

ahqsoftwares: I shall try today night

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: @amrbashir

amrbashir: I'd like to use zbus anyways, because tauri uses `rfd` crate which uses `zbus` too and then there is `tauri-plugin-single-instance` which uses `zbus` as well so If we were to go with the dbus approach, we should be using `zbus`

ahqsoftwares: I see, it saves in the long run, when all crates build

MegaSa1nt:
> You will have to fork wry and tauri and change the git url? (I am not much sure about using custom repos in packages)

image
Like this?

MegaSa1nt:
if i try to add "webkit2gtk-sys" manually, it still shows me this error:
```
    Updating crates.io index
      Adding webkit2gtk-sys v1.1.0 to dependencies.
             Features:
             - dox
             - v2_10
             - v2_12
             - v2_14
             - v2_16
             - v2_18
             - v2_20
             - v2_22
             - v2_24
             - v2_26
             - v2_28
             - v2_30
             - v2_32
             - v2_34
             - v2_36
             - v2_38
             - v2_40
             - v2_6
             - v2_8
error: failed to select a version for `webkit2gtk-sys`.
    ... required by package `webkit2gtk v0.18.2`
    ... which satisfies dependency `webkit2gtk = "^0.18.2"` of package `tauri v1.3.0`
    ... which satisfies dependency `tauri = "^1.3"` of package `gcs v1.3.3 (E:\GreenCatsServer\sa1nt.gcs\gcs\src-tauri)`
versions that meet the requirements `^0.18` are: 0.18.0

the package `webkit2gtk-sys` links to the native library `web_kit2`, but it conflicts with a previous package which links to `web_kit2` as well:
package `webkit2gtk-sys v1.1.0`
    ... which satisfies dependency `webkit2gtk-sys = "^1.1.0"` of package `gcs v1.3.3 (E:\GreenCatsServer\sa1nt.gcs\gcs\src-tauri)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the links ='webkit2gtk-sys' value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `webkit2gtk-sys` which could resolve this conflict
```

FabianLars:
@MegaSa1nt You'd have to use Tauri's next branch, and probably patch that to use wry from git too

amrbashir:
Please keep the discussion here related to the PR itself and use discord or some other chat platform to triage and help testing the PR.

if you want to test the PR, see https://github.com/tauri-apps/tao/pull/742#issuecomment-1588968729 and follow the instructions in that repo

src/platform_impl/linux/taskbar/xapps.rs
ahqsoftwares: Update
- Implementation will be dropped in next commit to the pr

src/platform_impl/linux/event_loop.rs
ahqsoftwares: Trying if I can implement it without Rc or Arc

ahqsoftwares: WSL Green Signal
Just left testing the outputs

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: Implemented something with zbus with wsl green signal. Idk if it'll work.
Will check that out tomorrow in my linux vm

src/platform_impl/linux/taskbar/unity.rs
```diff
@@ -0,0 +1,131 @@
...
+  dirty_count_visible: bool,
+  dirty_urgent: bool,
+}
+
+enum DataTypes<'a> {
+  Str(&'a str),
+  String(&'a String),
+  Bool(&'a bool),
+  Number(&'a i32),
+  Float(&'a f64),
```
amrbashir: No need for `Str` and `String` at the same time, also no need to take a reference for `bool`, `i32` and `f64` because they all live on the stack and derive `Copy`
```suggestion
  Str(&'a str),
  Bool(bool),
  Number(i32),
  Float(f64),
```

src/platform_impl/linux/taskbar/unity.rs
```diff
@@ -0,0 +1,131 @@
...
+    if !properties.is_empty() {
+      let mapped = properties
+        .iter()
+        .map(|x| match x {
+          DataTypes::Str(string) => string.clone().to_string(),
+          DataTypes::String(string) => string.clone().to_owned(),
+          DataTypes::Bool(val) => val.to_string(),
+          DataTypes::Number(val) => val.to_string(),
+          DataTypes::Float(val) => val.to_string(),
+        })
```
amrbashir: instead of this, implement [`ToString`](https://doc.rust-lang.org/std/string/trait.ToString.html) trait for `DataType` then call `.map(ToString::to_string)`

src/platform_impl/linux/taskbar.rs
```diff
@@ -0,0 +1,66 @@
...
+pub struct TaskbarIndicator {
+  unity: Option<unity::Manager>,
+  progress: f64,
+  progress_visible: bool,
+  needs_attention: bool,
+}
+
+#[allow(dead_code)]
+impl TaskbarIndicator {
+  pub fn new() -> Result<Self, Box<dyn std::error::Error>> {
```
amrbashir: No need to return a result here, it doesn't contain any fallible operation

```suggestion
  pub fn new() -> Self {
```

src/platform_impl/linux/event_loop.rs
```diff
@@ -312,6 +316,66 @@ impl<T: 'static> EventLoop<T> {
...
+              }
+            }
+
+            if &taskbar.is_some() == &true {
+              taskbar
+                .as_mut()
+                .unwrap()
+                .set_progress(progress)
+                .unwrap_or(());
+            }
```
amrbashir:
```suggestion
            if taskbar.is_none() {
            	taskbar.replace(TaskbarIndicator::new());
            }
            
            if let Some(taskbar) = &mut taskbar {
            	if let Some(unity_uri) = &unity_uri {
            		if let Err(e) = taskbar.set_unity_app_uri(unity_uri) {
            			log::warn!("Failed to set unity app uri {}", e);
            		}
            	}
            	
            	if let Err(e) = taskbar.set_progress(progress) {
            		log::warn!("Failed to set progres: {}", e);
            	}
            }
```
and do the same for `WindowRequest::TaskbarProgressState` branch

src/platform/windows.rs
```diff
@@ -117,6 +117,12 @@ pub trait WindowExtWindows {
   /// Whether to show the window icon in the taskbar or not.
   fn set_skip_taskbar(&self, skip: bool);
 
+  /// Sets the Taskbar Progress State
+  fn set_taskbar_progress_state(&self, state: i32);
+
+  /// Sets the Taskbar Progress
+  fn set_taskbar_progress(&self, current: u64, total: u64);
+
```
amrbashir: lets combine these two methods into one method and move it back to `src/window.rs`:
```suggestion
  fn set_progress_bar(&self, state: ProgressBarState);
```
and

```rs
pub enum ProgressState{
  None,
  Normal,
  Intermediate,
  Error,
  Paused,
}

pub struct ProgressBarState {
	pub state: ProgressState,
	pub total: u32,
	pub current: u32,
	/// **Linux Only**
	pub unity_uri: Option<String>,
}
```

Cargo.toml
```diff
@@ -121,6 +121,7 @@ gdk-pixbuf = "0.16"
 libappindicator = { version = "0.8.0", optional = true }
 dirs-next = { version = "2.0.0", optional = true }
 x11-dl = "2.21"
+zbus = "3.13.1"
```
amrbashir:
```suggestion
zbus = "3.13"
```

amrbashir (changes requested):

src/platform_impl/linux/taskbar/unity.rs
lucasfernog: Prefer [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) instead

amrbashir: damn, I always forget which one has a blanket impl for the other

ahqsoftwares:
Implementation of unity not tested. I shall check that before implementing the requested changes

src/platform/windows.rs
ahqsoftwares: I think keeping them separate is better? Because You can change the progress without changing the state (can save in computation?).

ahqsoftwares: Also, It'll affect the windows version too

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: But the compiler was just complaining about that (when I was debugging in kali-wsl)

ahqsoftwares:
@amrbashir It seems like the zbus implementation is failing under the hood. Need to triage into that

ahqsoftwares:
> @amrbashir It seems like the zbus implementation is failing under the hood. Need to triage into that

Update
Will triage after some time (I surely suck at linux crates)

src/platform/windows.rs
amrbashir: > Because You can change the progress without changing the state

the `state` field could be an `Option<ProgressState>` if it is `None`, we keep the old progress state and update only the progress value.

Actually, lets also remove `total` field and change `current` to `progress` and document that progress should be between `0` and `100` and have the total internally fixed to `100` so it would be like this
```rs
pub struct ProgressBarState {
	pub state: Option<ProgressState>,
	pub progress: u32,
	/// **Linux Only**
	pub unity_uri: Option<String>,
}
```

> Also, It'll affect the windows version too

That's fine

src/platform_impl/linux/taskbar/unity.rs
amrbashir: If you have a reference to a `Copy` value, you can just dereference it:
```rs
let x = 21; // x == 21
let y = &x; // y == &21
let d = *y; // d == 21
```

src/platform/windows.rs
ahqsoftwares: ✅

ahqsoftwares: Shall Implement after I triage the zbus impl not working

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: Referencing as `*` is safe? I mean like can be act as a dangling pointer causing a panic?

amrbashir:
Why is it failing? is the message never sent or? do you have any stack traces or errors?

ahqsoftwares:
> Why is it failing? is the message never sent or? do you have any stack traces or errors?

No error does occur, but it doesn't work again..
Checking it as a separate app
(my wsl is always slower than my main cpu by 8x, I guess will take a lot of time.....)

amrbashir:
Let me know if you get stuck and I will switch over to my linux and test

src/platform_impl/linux/taskbar/unity.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+    if self.dirty_urgent {
+      self.dirty_urgent = false;
+
+      properties.push(DataTypes::Str("urgent"));
+      properties.push(DataTypes::Bool(&self.urgent));
+    }
+
+    if !properties.is_empty() {
+      let mapped = properties
```
pewsheen: I guess `mapped` is the reason why this zbus implementation is not working.

The type of `mapped`'s properties is mapping to `string` only, so the server side may not find the method to call since the signature is different.

Maybe you could try using `zvarient` provided by zbus and let them do the Type transformation.
The `Dictionary` strcut in this [FAQ](https://dbus2.github.io/zbus/faq.html) seems fit the need of `properties`

pewsheen: Something like this

```rust
#[derive(Default, SerializeDict, DeserializeDict, Type, PartialEq, Debug)]
#[zvariant(signature = "dict")]
struct UpdateProperties {
  progress: Option<f64>,
  #[zvariant(rename = "progress-visible")]
  progress_visible: Option<bool>,
  count: Option<i32>,
  #[zvariant(rename = "count-visible")]
  count_visible: Option<bool>,
  urgent: Option<bool>
}

let mut properties = UpdateProperties::default();

properties.progress = Some(self.progress);

let body = (self.app_uri.to_owned(), properties);
let signal =
      MessageBuilder::signal("/", "com.canonical.Unity.LauncherEntry", "Update")?.build(&body)?;
```

ahqsoftwares: I also found another varient using HashMap
I am kinda researching.

Will work on an impl today

ahqsoftwares: Main Problem I am facing currently is that my area lies in the storm region of India. And Electricity is behaving like anything.. (and I own a desktop)

ahqsoftwares:
Thank to @pewsheen 
The ZBus Implementation is working as expected

Slowly gonna proceed in the reviewed changes

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: Already Resolved

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: Already Resolved

src/platform/windows.rs
ahqsoftwares: I think let's keep the total, that makes it better

amrbashir: I don't see a use-case for `total` field tbh, we can just assume the total to be `100` and `current` should be between `0` and `100`

ahqsoftwares: Actually, I have a use case
Like the user is implementing a file downloader, then it'll it easier for supplying `current, total` instead of converting that into `current * 100 / total` for that person (considering he's starting at rust and doesn't know a lot)

amrbashir: I'd rather keep the calculation on the user side of things instead of in the library code. There is use-cases when the total is not only downloading the file and might be related to installing it as well. In any use-case, the calculation is super simple that I don't see any value adding it inside the library.

ahqsoftwares: That works, though there will be an internal counter that converts anything > 100 as 100

src/platform_impl/linux/event_loop.rs
ahqsoftwares: Resolved in the next commit

src/platform_impl/linux/taskbar/unity.rs
```diff
@@ -0,0 +1,92 @@
...
+    Ok(m)
+  }
+
+  fn update(&mut self) -> Result<()> {
+    let mut properties = Progress::default();
+
+    if self.dirty_progress {
+      self.dirty_progress = false;
+
+      properties.progress = Some(*&self.progress);
```
pewsheen: I think `*&` won't be necessary?

ahqsoftwares: The compiler complains if not used

amrbashir: it shouldn't, `self.progress` is of type `f64` which is a `Copy` type and `*&self.progress` is the same as `self.progress` because you're dereferencing by using `*` the reference you created using `&` which is just the same as doing `self.progress``

src/lib.rs
```diff
@@ -170,6 +170,22 @@ extern crate bitflags;
 #[macro_use]
 extern crate objc;
 
+/// Taskbar Progress State
+pub enum TaskbarProgressState {
```
amrbashir:
```suggestion
pub enum ProgressBarState {
```

src/lib.rs
```diff
@@ -170,6 +170,22 @@ extern crate bitflags;
...
+  None,
+  Normal,
+  Intermediate,
```
amrbashir: Windows supports 5 states, `Error` variant is missing

src/lib.rs
```diff
@@ -170,6 +170,22 @@ extern crate bitflags;
...
+  Paused,
+}
+
+pub struct ProgressBarState {
+  pub state: Option<TaskbarProgressState>,
+  /// **Treated as Percentage**
+  pub progress: Option<f64>,
+  /// **Linux Only**
+  pub unity_uri: Option<String>,
+}
```
amrbashir:
```suggestion
/// A struct describing the state of the progress bar
pub struct ProgressBarState {
  /// The progress bar state
  pub state: Option<TaskbarProgressState>,
  /// The progress bar progress. This can be a value ranging from `0` to `100`
  pub progress: Option<f64>,
  /// An identifier for your app to communicate with the Unity desktop window manager **Linux Only**
  pub unity_uri: Option<String>,
}
```

src/platform/unix.rs
```diff
@@ -32,6 +33,11 @@ pub trait WindowExtUnix {
 
   /// Whether to show the window icon in the taskbar or not.
   fn set_skip_taskbar(&self, skip: bool);
+
+  /// Sets the taskbar progress state./
+  /// ## NOTE
+  /// - On **Linux:** App must be installed using .deb binary, Might not work on some distros like Linux Mint (Cinnamon)
+  fn set_progress_bar(&self, state: ProgressBarState);
```
amrbashir: remove from here and move it to `src/window.rs`

src/platform/unix.rs
```diff
@@ -42,6 +48,28 @@ impl WindowExtUnix for Window {
...
+        .set_taskbar_progress_state(taskbar_state, progress.unity_uri.clone());
+    }
+    if let Some(value) = progress.progress {
+      let progress_num = if value > 100 { 100f64 } else { value };
+
+      self
+        .window
+        .set_taskbar_progress(progress_num, progress.unity_uri.clone());
+    }
+  }
```
amrbashir: remove this logic, create a method `set_progress_bar` in `src/platform_impl/linux/window.rs` and move this logic there

src/platform/windows.rs
```diff
@@ -117,6 +118,9 @@ pub trait WindowExtWindows {
   /// Whether to show the window icon in the taskbar or not.
   fn set_skip_taskbar(&self, skip: bool);
 
+  /// Sets the taskbar progress state./
+  fn set_progress_bar(&self, progress: ProgressBarState);
```
amrbashir: same as above, remove and use a unified method in `src/window.rs`

src/platform/windows.rs
```diff
@@ -166,6 +170,27 @@ impl WindowExtWindows for Window {
...
+      };
+
+      self.window.set_taskbar_progress_state(taskbar_state);
+    }
+    if let Some(value) = progress.progress {
+      let progress = if value > 100 { 100f64 } else { value };
+
+      self.window.set_taskbar_progress(progress);
+    }
+  }
```
amrbashir: same as above, move this implementation to `src/platform_impl/windows/window.rs`

src/platform_impl/windows/window.rs
```diff
@@ -881,6 +882,32 @@ impl Window {
...
+    let handle = self.window.0;
+
+    unsafe {
+      let taskbar_list: ITaskbarList = CoCreateInstance(&TaskbarList, None, CLSCTX_SERVER).unwrap();
+
+      taskbar_list
+        .SetProgressValue(handle, current as u64, 100)
+        .unwrap_or(());
+    }
+  }
```
amrbashir: combine theses into one function `set_progress_bar` that takes `ProgressBarState` and conditionally check for the `Some` fields and set it

src/platform_impl/linux/window.rs
```diff
@@ -911,6 +931,8 @@ pub enum WindowRequest {
   SetMenu((Option<menu::Menu>, AccelGroup, gtk::MenuBar)),
   GlobalHotKey(u16),
   SetVisibleOnAllWorkspaces(bool),
+  TaskbarProgressState(bool, Option<String>),
+  TaskbarProgress(f64, Option<String>),
```
amrbashir: lets combine these two int one variant that takes `ProgressBarState` strcut
```suggestion
  PrgoressBarState(ProgressBarState),
```

src/platform_impl/linux/window.rs
```diff
@@ -859,6 +859,26 @@ impl Window {
...
+
+  #[inline]
+  pub fn set_taskbar_progress(&self, current: f64, unity_uri: Option<String>) {
+    if let Err(e) = self.window_requests_tx.send((
+      self.window_id,
+      WindowRequest::TaskbarProgress(current / 100.0, unity_uri),
+    )) {
+      log::warn!("Fail to send skip taskbar request: {}", e);
+    }
+  }
```
amrbashir: and combine these two into one method `set_progress_bar`

src/platform_impl/linux/event_loop.rs
```diff
@@ -312,6 +316,40 @@ impl<T: 'static> EventLoop<T> {
...
+                if let Err(e) = taskbar.set_unity_app_uri(unity_uri) {
+                  log::warn!("Failed to set unity app uri {}", e);
+                }
+              }
+
+              if let Err(e) = taskbar.set_progress_state(state) {
+                log::warn!("Failed to set progres: {}", e);
+              }
+            }
+          }
```
amrbashir: and combine these as well

src/platform_impl/linux/taskbar.rs
```diff
@@ -0,0 +1,66 @@
...
+    &mut self,
+    needs_attention: bool,
+  ) -> Result<(), Box<dyn std::error::Error>> {
+    self.needs_attention = needs_attention;
+
+    if let Some(ref mut unity) = self.unity {
+      unity.needs_attention(needs_attention)?;
+    }
+    Ok(())
+  }
```
amrbashir: lets combine all these into one method called `update` that takes `ProgressBarState` as inupt, so if `progress` and `state` fields are set at the same time, we make one dbus call instead of two

src/platform_impl/linux/taskbar/unity.rs
```diff
@@ -0,0 +1,92 @@
...
+    self.update()?;
+    Ok(())
+  }
+
+  pub fn needs_attention(&mut self, needs_attention: bool) -> Result<()> {
+    self.urgent = needs_attention;
+    self.dirty_urgent = true;
+    self.update()?;
+    Ok(())
+  }
```
amrbashir: same as above, merge these all into one `update` method

src/platform_impl/linux/taskbar.rs
```diff
@@ -0,0 +1,66 @@
+mod unity;
+
+pub struct TaskbarIndicator {
+  unity: Option<unity::Manager>,
+  progress: f64,
+  progress_visible: bool,
+  needs_attention: bool,
```
amrbashir: no need to keep track of these state, just remove them

src/platform_impl/linux/taskbar/unity.rs
```diff
@@ -0,0 +1,92 @@
...
+
+pub struct Manager {
+  conn: Connection,
+  app_uri: String,
+  progress: f64,
+  progress_visible: bool,
+  urgent: bool,
+  dirty_progress: bool,
+  dirty_progress_visible: bool,
+  dirty_urgent: bool,
```
amrbashir: same as above, no need to keep track of these values

amrbashir (changes requested):

src/lib.rs
ahqsoftwares: The Error Variant seems to do nothing with windows 11 at least

src/platform_impl/windows/window.rs
ahqsoftwares: Pretty Sure I missed to change that

ahqsoftwares:
Strange, my wsl & native windows didn't complain about value > 100

src/lib.rs
amrbashir: There is nothing in the documentation stating that but anyways we still need to add it for older versions.

ahqsoftwares: Got it, tauri will support windows 7 (even after it's EOS) & probably 10 too (after 2025)

amrbashir: For now, we support windwos 7 and higher, that might change in the future but for now we take windows 7 into considerations

src/lib.rs
ahqsoftwares: Conflicts with the struct ProgressBarState
Using ProgressState instead

src/platform_impl/linux/taskbar.rs
ahqsoftwares: Yeah

src/lib.rs
amrbashir: right, just call it `ProgressState` then

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: Done in next commit

src/lib.rs
ahqsoftwares: Adding in next commit

src/platform/windows.rs
ahqsoftwares: Adding in next commit

src/platform/unix.rs
ahqsoftwares: ✅

src/platform_impl/linux/taskbar/unity.rs
ahqsoftwares: Next commit will try to address that

src/platform_impl/linux/taskbar.rs
```diff
@@ -0,0 +1,59 @@
...
+      let progress = if progress > 100 { 100 } else { progress };
+
+      properties.progress = Some(progress as f64 / 100.0);
+    }
+
+    if let Some(state) = progress.state {
+      match state {
+        ProgressState::None => properties.progress_visible = Some(false),
+        _ => properties.progress_visible = Some(true),
+      }
```
amrbashir:
```suggestion
      properties.progress_visible = matches!(state, ProgressState::None);
```

src/lib.rs
```diff
@@ -170,6 +170,27 @@ extern crate bitflags;
...
+
+pub struct ProgressBarState {
+  /// The progress bar state.
+  pub state: Option<ProgressState>,
+  /// The progress bar progress. This can be a value ranging from `0` to `100`
+  pub progress: Option<u64>,
+  /// The identifier for your app to communicate with the Unity desktop window manager **Linux Only**
+  pub unity_uri: Option<String>,
+}
```
amrbashir: lets move these into `src/window.rs`

src/window.rs
```diff
@@ -1042,6 +1042,17 @@ impl Window {
...
   }
 
+  /// Sets the taskbar progress state.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **iOS / Android / macOS:** Unsupported.
+  #[inline]
+  pub fn set_progress_bar(&self, progress: ProgressBarState) {
+    #[cfg(not(target_os = "macos"))]
```
amrbashir:
```suggestion
    #[cfg(any(
    	windows,
        target_os = "linux",
  		target_os = "dragonfly",
 		target_os = "freebsd",
  		target_os = "netbsd",
  		target_os = "openbsd"
    ))]
```

src/window.rs
```diff
@@ -1042,6 +1042,17 @@ impl Window {
     self.window.set_ime_position(position.into())
   }
 
+  /// Sets the taskbar progress state.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **iOS / Android / macOS:** Unsupported.
```
amrbashir:
```suggestion
  /// - **Linux**: Progress bar is app-wide and not specific to this window.
  /// - **macOS**: Unimplemented.
  /// - **iOS / Android / macOS:** Unsupported.
```

src/window.rs
```diff
@@ -1042,6 +1042,17 @@ impl Window {
...
+  #[inline]
+  pub fn set_progress_bar(&self, progress: ProgressBarState) {
```
amrbashir, line 1074: we need to also provide this function on `EventLoopWindowTarget` struct but for Linux only (and macOS in the future)

src/platform_impl/linux/taskbar.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    let signal = MessageBuilder::signal("/", "com.canonical.Unity.LauncherEntry", "Update")?
+      .build(&(self.app_uri.clone(), properties))?;
+
+    self.conn.send_message(signal).unwrap();
```
amrbashir: we shouldn't unwrap here
```suggestion
    self.conn.send_message(signal)?;
```

src/platform_impl/linux/event_loop.rs
```diff
@@ -312,6 +316,19 @@ impl<T: 'static> EventLoop<T> {
...
+          WindowRequest::PrgoressBarState(state) => {
+            if taskbar.is_none() {
+              if let Ok(indicator) = TaskbarIndicator::new() {
+                taskbar.replace(indicator);
+              }
+            }
+
+            if let Some(taskbar) = &mut taskbar {
+              if let Err(e) = taskbar.update(state) {
+                log::warn!("Failed to set taskbar progress {}", e);
```
amrbashir:
```suggestion
                log::warn!("Failed to update taskbar progress {}", e);
```

src/platform_impl/linux/taskbar.rs
```diff
@@ -0,0 +1,59 @@
...
+struct Progress {
+  progress: Option<f64>,
+  #[zvariant(rename = "progress-visible")]
+  progress_visible: Option<bool>,
+  urgent: Option<bool>,
+}
+
+impl TaskbarIndicator {
+  pub fn new() -> Result<Self> {
+    let conn = Connection::session()?;
```
amrbashir, line 25: we need to check if we are running inside `Unity` and avoid creating a dbus connection on othr desktop environments

src/platform_impl/linux/taskbar.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    Ok(Self {
+      conn,
+      app_uri: String::new(),
+    })
+  }
+
+  pub fn update(&mut self, progress: ProgressBarState) -> Result<()> {
```
amrbashir, line 33: same here, avoid updating if we are not inside `Unity` desktop manager.

amrbashir (changes requested):

src/platform_impl/linux/taskbar.rs
ahqsoftwares: any way?

src/window.rs
ahqsoftwares: with logic or just the fn?

amrbashir: it will be something like this
```rs
#[inline]
pub fn set_progress_bar(&self, progress: ProgressBarState) {
	self.p.set_progress_bar(progress)
}
```

and then in `src/platfomr_impl/linux/event_loop.rs` you will have the implementation use  this field to send the request
https://github.com/tauri-apps/tao/blob/9c8613c3ba3cd64dbaa2d9ddbdbe75bae7035d44/src/platform_impl/linux/event_loop.rs#L57

and then you can receive the request here https://github.com/tauri-apps/tao/blob/9c8613c3ba3cd64dbaa2d9ddbdbe75bae7035d44/src/platform_impl/linux/event_loop.rs#L905 and do the same logic

src/platform_impl/linux/taskbar.rs
amrbashir: maybe this https://stackoverflow.com/questions/8742750/how-can-i-detect-when-im-on-a-system-running-unity can help

ahqsoftwares:
Severe rain currently, checking after 1-3days

ahqsoftwares:
Pretty sure will be active tomorrow

src/lib.rs
ahqsoftwares: Resolved in this commit

src/platform_impl/linux/taskbar.rs
ahqsoftwares: Done

src/platform_impl/linux/taskbar.rs
ahqsoftwares: Done

src/window.rs
ahqsoftwares: Done

src/window.rs
```diff
@@ -18,6 +18,27 @@ use crate::{
...
+  /// **Treated as Normal in linux**
+  Paused,
+  /// **Treated as Normal in linux**
+  Error,
+}
+
+pub struct ProgressBarState {
+  /// The progress bar state.
+  pub state: Option<ProgressState>,
+  /// The progress bar progress. This can be a value ranging from `0` to `100`
```
pewsheen, line 36: How about using `f64` and ranging the progress from 0 to 1?

Since most use cases are only a part of something, we don't need to multiply by 100 every time we calculate the progress.

ahqsoftwares: Then too you'll need to multiply it by 100, because the windows impl takes u64

ahqsoftwares: and float isn't a great type after all I think (precision issues with floats)

pewsheen: Multiply the value first is a bit unnecessary because each platform takes different input like:
- Linux takes a `double` value from `0.0` to `1.0`
- Windows takes `0` to `100`
- macOS I think it defaults to `0` to `100` too, but it can be set to `0.0` to `1.0` as well

So either way we need to do the math on the `platform_impl` side. But yes, letting the range from `0` to `100` is much more intuitive when you want to set progress directly.

wusyong: There are two platforms using integer. I think it's better to stick to 0-100.
float causes more operation overhead than multiply imho.

ahqsoftwares: Yeah totally agree

Cargo.toml
```diff
@@ -121,6 +121,7 @@ gdk-pixbuf = "0.16"
 libappindicator = { version = "0.8.0", optional = true }
 dirs-next = { version = "2.0.0", optional = true }
 x11-dl = "2.21"
+zbus = "3.13"
```
pewsheen: btw, I think the latest zbus release (3.13.1) will cause dependency conflict with `tauri:next` branch due to
https://github.com/dbus2/zbus/commit/dfbfd21ab42eb46d73d83e4ab6ec0ea86f957609 is not released yet.

ref: https://github.com/tauri-apps/tauri/pull/6845

FabianLars: we've unlocked the winnow dependency in tauri again so it shouldn't be that much of an issue anymore, _but_ for this to really work we'll have to lower the minor version specified here **or just go with `"3"`**.

ahqsoftwares: Well, idk what to say

ahqsoftwares: Though at max I guess it will cause dependency cloning?

FabianLars: > Though at max I guess it will cause dependency cloning?

Only if the major versions would be different. For 2x the same major version cargo will complain that it couldn't find a fitting version.

ahqsoftwares: Well, changing zbus to just `3` will help?

FabianLars: yes, then can cargo go as low as it needs and users with MSRV requirements can pin it to the minor version they need themselves.

ahqsoftwares: ✅

pewsheen: oh okay, I see no problem.

src/event_loop.rs
```diff
@@ -253,6 +253,20 @@ impl<T> EventLoopWindowTarget<T> {
...
+  /// 
+  /// ## Platform-specific
+  /// 
+  /// - **Windows:** Use the Progress Bar Function Available in Window (Windows can have different progress bars for different window)
+  /// - **macOS:** Unimplemented.
+  /// - **iOS / Android:** Unsupported.
+  #[inline]
+  #[allow(unused)]
+  pub fn set_progress_bar(&self, progress: ProgressBarState) {
+    #[cfg(target_os = "macos")]
```
pewsheen: Should this be Windows and Linux only?

ahqsoftwares: Yes if u like to impl macos one using dock
Or else impl it as an alias to macos dock impl

ahqsoftwares:
Anything left to refactor?

src/platform_impl/linux/taskbar.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    if let Some(progress) = progress.progress {
+      let progress = if progress > 100 { 100 } else { progress };
+
+      properties.progress = Some(progress as f64 / 100.0);
+    }
+
+    if let Some(state) = progress.state {
+      properties.progress_visible = Some(matches!(state, ProgressState::None));
```
pewsheen:
```suggestion
      properties.progress_visible = Some(!matches!(state, ProgressState::None));
```

Should be visible if it's not `ProgressState::None`

src/platform_impl/linux/util.rs
```diff
@@ -27,3 +28,24 @@ pub fn cursor_position(is_wayland: bool) -> Result<PhysicalPosition<f64>, Extern
       .ok_or(ExternalError::Os(os_error!(super::OsError)))?
   }
 }
+
+pub fn is_unity() -> bool {
+  if let Ok(child) = Command::new("echo")
+    .arg("$XDG_CURRENT_DESKTOP")
+    .stdout(Stdio::piped())
+    .spawn()
```
pewsheen:
```suggestion
  if let Ok(child) = Command::new("sh")
    .arg("-c")
    .arg("echo -n $XDG_CURRENT_DESKTOP")
    .stdout(Stdio::piped())
    .spawn()
```

`echo` is built-in the `sh` so you need to call it from `sh`

`echo -n` to trim the newline that return from the `echo` otherwise the return will be like `Unity\n`

src/platform_impl/linux/util.rs
```diff
@@ -27,3 +28,24 @@ pub fn cursor_position(is_wayland: bool) -> Result<PhysicalPosition<f64>, Extern
...
+        string == "unity" || string == "gnome"
+      } else {
+        false
+      }
+    } else {
+      false
+    }
+  } else {
+    false
+  }
```
pewsheen: I think those `false` can be merged into one and return `true` when the string matched

```rust
pub fn is_unity() -> bool {
  if let Ok(child) = Command::new("sh")
    .arg("-c")
    .arg("echo -n $XDG_CURRENT_DESKTOP")
    .stdout(Stdio::piped())
    .spawn()
  {
    if let Ok(output) = child.wait_with_output() {
      if let Ok(string) = String::from_utf8(output.stdout) {
        let string = string.as_str().to_lowercase();
        return string == "unity" || string == "gnome";
      }
    }
  }
  false
}
```

ahqsoftwares: Well, idk why it didn't came to my mind...

ahqsoftwares:
@amrbashir Review requested

src/event_loop.rs
```diff
@@ -253,6 +253,20 @@ impl<T> EventLoopWindowTarget<T> {
...
+
+  /// Sets the progress bar state
+  /// 
+  /// ## Platform-specific
+  /// 
+  /// - **Windows:** Use the Progress Bar Function Available in Window (Windows can have different progress bars for different window)
+  /// - **macOS:** Unimplemented.
+  /// - **iOS / Android:** Unsupported.
+  #[inline]
+  #[allow(unused)]
```
pewsheen:
```suggestion
  /// - **Windows:** Use the Progress Bar Function Available in Window (Windows can have different progress bars for different window)
  /// - **Linux:** Only supported desktop environments with `libunity` (e.g. GNOME).
  /// - **macOS:** Unimplemented.
  /// - **iOS / Android:** Unsupported.
  #[inline]
```

src/window.rs
```diff
@@ -1042,6 +1063,26 @@ impl Window {
     self.window.set_ime_position(position.into())
   }
 
+  /// Sets the taskbar progress state.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux**: Progress bar is app-wide and not specific to this window.
+  /// - **macOS**: Unimplemented.
+  /// - **iOS / Android / macOS:** Unsupported.
```
pewsheen:
```suggestion
  /// Sets the taskbar progress state.
  ///
  /// ## Platform-specific
  ///
  /// - **Linux**: Progress bar is app-wide and not specific to this window. Only supported desktop environments with `libunity` (e.g. GNOME).
  /// - **macOS**: Unimplemented.
  /// - **iOS / Android:** Unsupported.
```

pewsheen (changes requested):
Just some documentation needs to be added then it's good to go!

pewsheen (approved):

src/event_loop.rs
```diff
@@ -253,6 +253,20 @@ impl<T> EventLoopWindowTarget<T> {
...
+  /// ## Platform-specific
+  /// 
+  /// - **Windows:** Use the Progress Bar Function Available in Window (Windows can have different progress bars for different window)
+  /// - **Linux:** Only supported desktop environments with `libunity` (e.g. GNOME).
+  /// - **macOS:** Unimplemented.
+  /// - **iOS / Android:** Unsupported.
+  #[inline]
+  pub fn set_progress_bar(&self, progress: ProgressBarState) {
+    #[cfg(target_os = "linux")]
+    self.p.set_progress_bar(progress)
```
pewsheen:
```suggestion
  pub fn set_progress_bar(&self, _progress: ProgressBarState) {
    #[cfg(target_os = "linux")]
    self.p.set_progress_bar(_progress)
```

Oops, this should have a `_` prefix, my fault 😿

src/event_loop.rs
```diff
@@ -253,6 +254,20 @@ impl<T> EventLoopWindowTarget<T> {
   pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, ExternalError> {
     self.p.cursor_position()
   }
+
+  /// Sets the progress bar state
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows:** Use the Progress Bar Function Available in Window (Windows can have different progress bars for different window)
```
amrbashir:
```suggestion
  /// - **Windows:** Unsupported. Use the Progress Bar Function Available in Window (Windows can have different progress bars for different window)
```

src/platform_impl/linux/event_loop.rs
```diff
@@ -119,6 +121,21 @@ impl<T> EventLoopWindowTarget<T> {
...
+    let window = window.iter().next();
+
+    if let Some(window) = window {
+      if let Err(e) = self
+        .window_requests_tx
+        .send((*window, WindowRequest::ProgressBarState(progress)))
+      {
+        log::warn!("Fail to send update progress bar request: {}", e);
+      }
+    }
```
amrbashir:
```suggestion
    if let Err(e) = self
      .window_requests_tx
      .send((WindowId::dummy(), WindowRequest::ProgressBarState(progress)))
    {
      log::warn!("Fail to send update progress bar request: {}", e);
    }
```

src/platform_impl/linux/util.rs
```diff
@@ -27,3 +28,20 @@ pub fn cursor_position(is_wayland: bool) -> Result<PhysicalPosition<f64>, Extern
...
+    .spawn()
+  {
+    if let Ok(output) = child.wait_with_output() {
+      if let Ok(string) = String::from_utf8(output.stdout) {
+        let string = string.as_str().to_lowercase();
+        return string == "unity" || string == "gnome";
+      }
+    }
+  }
+  false
```
amrbashir:
```suggestion
  std::env::var("XDG_CURRENT_DESKTOP").map(|d| {
  	let d = d.to_lowercase();
  	d == "unity" || d == "gnome"
  }).unwrap_or(false)
```

src/platform_impl/linux/window.rs
```diff
@@ -859,6 +860,15 @@ impl Window {
     }
   }
 
+  pub fn set_progress_bar(&self, progress: ProgressBarState) {
+    if let Err(e) = self
+      .window_requests_tx
+      .send((self.window_id, WindowRequest::ProgressBarState(progress)))
```
```suggestion
      .send((WindowId::dummy(), WindowRequest::ProgressBarState(progress)))
```

src/platform_impl/linux/event_loop.rs
```diff
@@ -312,6 +332,21 @@ impl<T: 'static> EventLoop<T> {
             window.set_skip_taskbar_hint(skip);
             window.set_skip_pager_hint(skip)
           }
+          WindowRequest::ProgressBarState(state) => {
```
amrbashir: and move this logic down below next to global shortcut and menu handling

amrbashir (changes requested):
Just some final tweaks

src/platform_impl/linux/event_loop.rs
ahqsoftwares: ?

ahqsoftwares: This? https://github.com/tauri-apps/tao/pull/742/commits/16db1576d701903801b4c07e5ca809ed918ef07c

amrbashir: yeah

amrbashir:
@pewsheen @FabianLars could you give this PR a final test on your gnome installations?

ahqsoftwares:
I am too testing it on my ubuntu 20.04

ahqsoftwares:
Linux One Building

Till Then

Windows:
image
image

ahqsoftwares:
@amrbashir The linux impl doesn't seem to be working

.changes/implementTaskbarProgress.md
```diff
@@ -0,0 +1,7 @@
+---
+"tao": minor
+---
+
+Added a taskbar feature (to windows & linux) allowing the following to the tao crate
+- Taskbar Progress Bar (Window wise bar for windows & os wise bar for linux)
+- Taskbar Progress State Management (⬆️)
```
amrbashir:
```suggestion
Added APIs for setting progress bars for the application icon on Linux (Unity only) and for specific window on Windows.
```

pewsheen:
> @pewsheen @FabianLars could you give this PR a final test on your gnome installations?

Sadly no 😢 

https://github.com/tauri-apps/tao/pull/742/files#diff-e9c7e647c61cc58c5c7838936f4c17a860eb288e376301312c6a12ae66436a45R866
`WindowId.dummy()` somehow broke it

amrbashir (approved):

ahqsoftwares:
I got the error though
Ubuntu returns ubuntu:gnome

in short changing == to includes would do the job

ahqsoftwares:
image
Pretty sure that's related?

amrbashir:
@pewsheen how did `WindowId::dummy` break it?

ahqsoftwares:
> @pewsheen how did `WindowId::dummy` break it?

Nah it's not related to WindowId::dummy, the terminal output has changed. On ubuntu it returns ubuntu:GNOME
i.e. a string.includes is enough

amrbashir:
@pewsheen could you give it another test with the latest commit?

src/platform_impl/linux/util.rs
```diff
@@ -4,6 +4,7 @@ use crate::{
   dpi::{LogicalPosition, PhysicalPosition},
   error::ExternalError,
 };
+use std::process::{Command, Stdio};
```
amrbashir: unneeded imports

src/platform_impl/linux/util.rs
```diff
@@ -27,3 +28,12 @@ pub fn cursor_position(is_wayland: bool) -> Result<PhysicalPosition<f64>, Extern
       .ok_or(ExternalError::Os(os_error!(super::OsError)))?
   }
 }
+
+pub fn is_unity() -> bool {
+  std::env::var("XDG_CURRENT_DESKTOP")
+    .map(|d| {
+      let d = d.to_lowercase();
+      d.includes("unity") || d.includes("gnome")
```
amrbashir:
```suggestion
      d.contains("unity") || d.contains("gnome")
```

amrbashir (changes requested):

pewsheen:
It didn't receive `WindowRequest::ProgressBarState(state)` event in `linux/event_loop.rs`

@ahqsoftwares So your gnome is working?

ahqsoftwares:
> 

I spotted a typo in the code actually

pewsheen:
> > @pewsheen how did `WindowId::dummy` break it?
> 
> Nah it's not related to WindowId::dummy, the terminal output has changed. On ubuntu it returns ubuntu:GNOME i.e. a string.includes is enough

Mine is returned `unity` so I didn't aware of that

pewsheen:
Oh, I see the commit, should we remove `WindowRequest::ProgressBarState(state)` from https://github.com/tauri-apps/tao/pull/742/files#diff-a55deb213c2ceea1a601a8db19e26e0b3214a80210093214b8406d407f68537fR902 and just keep another one in `WindowId::Dummy()` block?

ahqsoftwares:
I am using github actions to build the tao-test binary, that takes a very long time

ahqsoftwares:
> Oh, I see the commit, should we remove `WindowRequest::ProgressBarState(state)` from https://github.com/tauri-apps/tao/pull/742/files#diff-a55deb213c2ceea1a601a8db19e26e0b3214a80210093214b8406d407f68537fR902 and just keep another one in `WindowId::Dummy()` block?

Can we reference it somehow like one forwards to another?

ahqsoftwares:
image
the gnome impl doesn't seem to work

ahqsoftwares:
strange, the util functions works as expected

ahqsoftwares:
image

I shall check it tomorrow.. It's 11:12PM here

pewsheen:
> > Oh, I see the commit, should we remove `WindowRequest::ProgressBarState(state)` from https://github.com/tauri-apps/tao/pull/742/files#diff-a55deb213c2ceea1a601a8db19e26e0b3214a80210093214b8406d407f68537fR902 and just keep another one in `WindowId::Dummy()` block?
> 
> Can we reference it somehow like one forwards to another?

Maybe just leave a noop branch like how `GlobalHotKey` does. I don't think it would match that branch (`window.id`!=`WindowId::dummy()`) since we only use `WindowId::dummy()` for setting the progress bar
```
WindowRequest::ProgressBarState(_state) => {}
```

amrbashir:
yeah @pewsheen is right, the logic shouldn't be duplicated

ahqsoftwares:
Sorry for the trouble, it seems like I typed the unity_uri wrong 💀

ahqsoftwares:
Gonna only impl for the dummy one

amrbashir (approved):
Thank you for your contributions

MegaSa1nt:
Finally!!!

ahqsoftwares:
> Finally!!!

Yup

PR #751, opened 2023-06-24
enhance(windows): apply dark mode to menus and other controls

### LINKED ISSUE #3240: [feat] Dark Right Click menus
### Describe the problem

Make Right Click Menus Blend In With Dark Mode App

Make A Setting For Light or Dark Right Click Menus

### Alternatives considered

IDK

### Additional context

_No response_

### LINKED ISSUE #6645: [feat] theme customization for system tray
### Describe the problem

When you have a Tauri app that uses the System Tray, when you right click your icon it opens a context menu, if configured to.
This menu is **always light theme, regardless of OS theme** (on Windows 11, at least).
If you look at Electron apps like Discord or Figma, they have a dark theme context menu.

As far as I have seen (I've spent an hour looking), there is no current way to change the colors or theme of the system tray menu for Tauri.

I am not yet familiar with Rust myself, so I cannot propose a code solution.

If it is possible, I would like the option to change background and foreground colors of the system tray, on Windows at least - I imagine other OSs use their system's theme for that.

If that is not possible, I would like the option to switch between dark and light system tray theme, on Windows at least - see above.

If neither is possible, I wonder what workarounds would be possible, like spawning a frameless window where the user clicked with your styled options.

### Alternatives considered

Spawning a frameless window where the user clicked with your styled options - a horrible way to do it as it would be unreliable and very OS-specific.

### Additional context

_No response_

ahodesuka:

- [x] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve tauri-apps/tauri#3240 (I think so at least)

### Other information
The current dark mode only applies to the window titlebar.  An additional undocumented function 
from uxtheme.dll must be called for the menus to appear with the dark theme.

The function that must be called has changed once since being introduced, in Windows 10 version 1903 (build 18362), 
this latest version allows you to force the theme color you want instead of just allowing dark to 
be set if it is the system default.

From what I have tested this works for system tray menus, menubar menus, and the system menu when right clicking the titlebar as seen below.
image

A somewhat related issue is tauri-apps/muda#97 although this issue seems to be more about the menubar control itself and not the menus.

Also closes https://github.com/tauri-apps/tauri/issues/6645


src/platform_impl/windows/dark_mode.rs
```diff
@@ -181,6 +191,71 @@ fn should_apps_use_dark_mode() -> bool {
...
 }
 
+// This enables dark mode menus
+fn set_preferred_app_mode(is_dark_mode: bool) -> bool {
+  let build_version = (*WIN10_BUILD_VERSION).unwrap();
+
+  // SetPreferredAppMode is for Windows 1903+ (18362)
+  if build_version >= 18362 {
+    type SetPreferredAppMode = unsafe extern "system" fn(PreferredAppMode) -> bool;
+    lazy_static! {
```
wusyong: This function doesn't feel like really need to be a static one. Could it just be code block?

wusyong:
Not sure how will this work with external crate like `muda` since we plan to move menu feature into this crate.
@amrbashir could you help review this?

src/platform_impl/windows/dark_mode.rs
ahodesuka: I just copied the pattern used in `should_apps_use_dark_mode`, making these static makes more sense since the value returned by `GetProcAddress` won't change.

ahodesuka:
This should work with `muda` and probably belongs in `tao` since it is in charge of managing themes.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -141,6 +150,7 @@ fn set_dark_mode_for_window(hwnd: HWND, is_dark_mode: bool) -> bool {
       };
 
       let status = set_window_composition_attribute(hwnd, &mut data as *mut _);
+      set_preferred_app_mode(is_dark_mode);
```
wusyong: It should handle the return value as well, right?

ahodesuka: I found that the `SetPreferredAppMode` function returns the previous value that was set which will always start at 0 (`PreferredAppMode::Default`) when the application first runs, I don't really know if checking this value is relevant.

I do not have a version of Windows 10 to test the older `AllowDarkModeForApp`.

Removing the return value from `set_preferred_app_mode` might make more sense.

amrbashir:
> Not sure how will this work with external crate like `muda` since we plan to move menu feature into this crate. @amrbashir could you help review this?

I will try to review ASAP (currently on a sick leave and some national holidays coming) but I agree with @ahodesuka, I think it is better to keep this in tao since it has info about the theme management and would be redundant to implement both in `tao` and `muda`, However, I think `muda` should add an easy API to call this undocumented method, so crates other than `tao` could call it whenever they want.

amrbashir:
Actually since we don't plan on releasing this for 1.x, lets just implement this in `muda` and let tauri decide if it wants to use this API, does that sound alright?

ahodesuka:
Looking at it now I think you are right.  It should be implemented in `muda` and should perhaps expand to theme the menubar itself as well.
I found a quick example [here in ppsspp](https://github.com/hrydgard/ppsspp/blob/master/Windows/W32Util/UAHMenuBar.cpp#L13) that uses the method you mentioned in tauri-apps/muda#97, but uses the theme menu color avoiding the use of any hard coded values.

amrbashir:
Yeah sounds good, feel free to open a PR there.

amrbashir:
After taking a second look, I was wrong, this API that switches the context menu should be called in `tao` and the custom drawing of the menubar should be in `muda`.

amrbashir (approved):
@ahodesuka thanks for your contributions and I apologize for dismissing the PR earlier without a deeper look.

PR #752, opened 2023-06-28
feat(ios): add support to universal links

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

### Checklist

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -164,6 +164,16 @@ pub enum Event<'a, T: 'static> {
   /// This is irreversable - if this event is emitted, it is guaranteed to be the last event that
   /// gets emitted. You generally want to treat this as an "do on quit" event.
   LoopDestroyed,
+
+  /// Emitted when the app is opening an URL.
+  Opened { event: OpenEvent },
+}
+
+/// What the app is opening.
+#[derive(Debug, PartialEq, Clone)]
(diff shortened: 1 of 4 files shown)


lucasfernog:
I tried to make the event generic so it can be reused on #422. We should finally agree on an event name thinking about Tauri v2.

So with this change my idea is that #422 is changed to use the `OpenEvent` and add the `File` variant (and maybe having both `OpenEvent::Url` and `OpenEvent::File` taking a Vec instead of a single value).

src/event.rs
```diff
@@ -164,6 +164,16 @@ pub enum Event<'a, T: 'static> {
   /// This is irreversable - if this event is emitted, it is guaranteed to be the last event that
   /// gets emitted. You generally want to treat this as an "do on quit" event.
   LoopDestroyed,
+
+  /// Emitted when the app is opening an URL.
```
amrbashir:
```suggestion
  /// Emitted when the app is open by external resources, like opening a Url.
```

amrbashir (approved):

amrbashir (approved):

PR #761, opened 2023-07-06
fix(android): use a crossbeam channel instead of a Mutex<VecDeque>

### LINKED ISSUE #760: SendEvent blocks the thread on Android due to double locking of user_queue mutex
**Describe the bug**
In Dioxus, we progress async tasks by attaching the waker to the event loop via `SendEvent`.

The code is as follows:
```rust
    impl ArcWake for DomHandle {
        fn wake_by_ref(arc_self: &Arc<Self>) {
            println!("waking by ref");
            _ = arc_self
                .proxy
                .send_event(UserWindowEvent(EventData::Poll, arc_self.id));
            println!("Send event succeeded!");
        }
    }
```

When I run our hello world examples, I see the `waking by ref` logged in the console but never `Send event succeeded`. The app did not panic.

The app hangs during the `send_event` method. I've tried tracking it down by adding logging in send_event.

```rust
  pub fn send_event(&self, event: T) -> Result<(), event_loop::EventLoopClosed<T>> {
    println!("Internal quuee locking...");
    let queue = self.queue.lock();
    println!("Internal queue locked");
    queue.unwrap().push_back(event);
    println!("Waking looper....");
    self.looper.wake();
    println!("Looper waked");
    Ok(())
  }
```

It looks like the hang occurs due to the lock on the mutex. By swapping the vecdeque out, I can get the bug to go away. A channel might be a better abstraction.

src/platform_impl/android/mod.rs:347

```rust
          let mut user_queue = {
            let mut user_queue = self.user_queue.lock().unwrap();
            let mut queue = VecDeque::new();
            std::mem::swap(&mut *user_queue, &mut queue);
            queue
          };
```

I think, instead of an `Arc<Mutex<VecDeque>>`, tao should be using a lock-free queue which won't run into this issue.

OS: macOS
Rustc: 1.70

- [ ] Yes
- [x] No

jkelleyrtp:
This fixes #760 by using a crossbeam channel, like its sibling iOS implementation.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #760

### DIFF
--- a/.changes/android-deadlock-fix.md
+++ b/.changes/android-deadlock-fix.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+On Android, use a lockfree queue (crossbeam channel) to prevent deadlocks inside send_event.
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -14,6 +14,7 @@ use crate::{
   monitor,
   window::{self, Theme},
 };
+use crossbeam_channel::{Receiver, Sender};
 use ndk::{
   configuration::Configuration,
   event::{InputEvent, KeyAction, MotionAction},
@@ -26,7 +27,7 @@ use raw_window_handle::{
 use std::{
   collections::VecDeque,
   convert::TryInto,
-  sync::{Arc, Mutex, RwLock},
+  sync::RwLock,
   time::{Duration, Instant},
 };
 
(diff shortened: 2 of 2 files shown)


wusyong (approved):

wusyong:
Thanks for the PR! Could you also sign the commits?

jkelleyrtp:
Done!

amrbashir (approved):

PR #762, opened 2023-07-07
Support right-to-left layout on Windows

abdnh:

This PR adds support for setting right-to-left layout on Windows using the [WS_EX_LAYOUTRTL](https://learn.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles) extended style and others. My main use for this is to make menus appear to the right in an Arabic app. I plan to add support for this to Tauri.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -121,6 +121,9 @@ pub trait WindowExtWindows {
   ///
   /// Enabling the shadow causes a thin 1px line to appear on the top of the window.
   fn set_undecorated_shadow(&self, shadow: bool);
+
+  /// Sets right-to-left layout.
+  fn set_rtl(&self, rtl: bool);
 }
 
 impl WindowExtWindows for Window {
@@ -169,6 +172,11 @@ impl WindowExtWindows for Window {
   #[inline]
   fn set_undecorated_shadow(&self, shadow: bool) {
     self.window.set_undecorated_shadow(shadow)
+  }
+
+  #[inline]
+  fn set_rtl(&self, rtl: bool) {
+    self.window.set_rtl(rtl)
   }
 }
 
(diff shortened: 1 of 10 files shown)


wusyong:
Let's make this platform-specific method if this is only for Windows.

abdnh:
It should be possible to implement on Linux and macOS at least. I'll need macOS support in my app in the future so I might add that too then (currently don't have the time to investigate it and want to get Windows support merged as soon as possible). 

Will make it a platform-specific method and add a .changes file.

src/platform_impl/windows/window_state.rs
```diff
@@ -108,6 +108,8 @@ bitflags! {
         /// Drop shadow for undecorated windows.
         const MARKER_UNDECORATED_SHADOW = 1 << 21;
 
+        const RTL = 1 << 22;
```
wusyong:
```suggestion
        const RIGHT_TO_LEFT_LAYOUT = 1 << 22;
```

src/platform/windows.rs
```diff
@@ -121,6 +121,9 @@ pub trait WindowExtWindows {
   ///
   /// Enabling the shadow causes a thin 1px line to appear on the top of the window.
   fn set_undecorated_shadow(&self, shadow: bool);
+
+  /// Sets right-to-left layout.
```
wusyong, line 125: Could you explain more about this layout in this documentation?

wusyong (changes requested):

amrbashir (approved):

amrbashir:
Oh looks like you didn't sign your commits and I can't merge the PR unless they are signed. You need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928) for example.

Seems like you also need to run `cargo fmt`

abdnh:
Actually signed off the first commit and then forgot to sign the rest. Should be fine now I guess.

amrbashir (approved):

wusyong (approved):

PR #766, opened 2023-07-12
feat(macOS): support progress bar on dock

pewsheen:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

#### Implement `macOS` functionality for #418 

On macOS, it didn't provide a simple API to set the progress indicator. Instead, we had to create an `NSProgressIndicator` and attach it to the dock tile.

Also, I found the original `ProgressState::Intermediate` might be a typo. It should be `Indeterminate`.
https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-itaskbarlist3-setprogressstate

### DIFF
--- a/.changes/macos-taskbar-progress.md
+++ b/.changes/macos-taskbar-progress.md
@@ -1,5 +1,0 @@
----
-"tao": minor
----
-
-On macOS, support setting progress indicator in the Dock.
--- a/.changes/macos-taskbar-progress.md
+++ b/.changes/macos-taskbar-progress.md
@@ -1,0 +1,5 @@
+---
+"tao": minor
+---
+
+On macOS, support setting progress indicator in the Dock.
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -265,7 +265,7 @@ impl<T> EventLoopWindowTarget<T> {
   /// - **iOS / Android:** Unsupported.
   #[inline]
   pub fn set_progress_bar(&self, _progress: ProgressBarState) {
-    #[cfg(target_os = "linux")]
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
     self.p.set_progress_bar(_progress)
   }
 }
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -261,7 +261,6 @@ impl<T> EventLoopWindowTarget<T> {
   ///
   /// - **Windows:** Unsupported. Use the Progress Bar Function Available in Window (Windows can have different progress bars for different window)
   /// - **Linux:** Only supported desktop environments with `libunity` (e.g. GNOME).
-  /// - **macOS:** Unimplemented.
   /// - **iOS / Android:** Unsupported.
   #[inline]
   pub fn set_progress_bar(&self, _progress: ProgressBarState) {
--- a/src/platform_impl/macos/progress_bar.rs
+++ b/src/platform_impl/macos/progress_bar.rs
@@ -146,8 +146,8 @@ extern "C" fn draw_progress_bar(this: &Object, _: Sel, rect: NSRect) {
     // draw progress
     let state: u8 = *(this.get_ivar("state"));
     let progress_color: id = match state {
-      3 => msg_send![class!(NSColor), grayColor], // paused
-      4 => msg_send![class!(NSColor), redColor],  // error
+      x if x == ProgressState::Paused as u8 => msg_send![class!(NSColor), grayColor],
+      x if x == ProgressState::Error as u8 => msg_send![class!(NSColor), redColor],
       _ => msg_send![class!(NSColor), whiteColor],
     };
     let _: () = msg_send![progress_color, set];
--- a/src/platform_impl/macos/progress_bar.rs
+++ b/src/platform_impl/macos/progress_bar.rs
@@ -31,6 +31,7 @@ pub fn set_progress_indicator(progress_state: ProgressBarState) {
       let _: () = msg_send![progress_indicator, setHidden: NO];
     }
     if let Some(state) = progress_state.state {
+      (*progress_indicator).set_ivar("state", state as u8);
       let _: () = msg_send![
         progress_indicator,
         setHidden: matches!(state, ProgressState::None)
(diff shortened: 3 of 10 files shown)


.changes/macos-taskbar-progress.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": minor
+---
+
+On macOS, support setting progress indicator in the Dock.
```
amrbashir: we haven't released a new version yet, so instead modify the change file from the other PR instead of creating a new one

src/window.rs
```diff
@@ -1095,7 +1096,8 @@ impl Window {
       target_os = "dragonfly",
       target_os = "freebsd",
       target_os = "netbsd",
-      target_os = "openbsd"
+      target_os = "openbsd",
+      target_os = "macos",
```
amrbashir, line 1099: you need to also update the documentation

src/event_loop.rs
```diff
@@ -265,7 +265,7 @@ impl<T> EventLoopWindowTarget<T> {
   /// - **iOS / Android:** Unsupported.
   #[inline]
   pub fn set_progress_bar(&self, _progress: ProgressBarState) {
-    #[cfg(target_os = "linux")]
```
amrbashir, line 268: you need to also update the documentation

src/platform_impl/macos/progress_bar.rs
```diff
@@ -0,0 +1,156 @@
...
+};
+use objc::{
+  declare::ClassDecl,
+  runtime::{Class, Object, Sel, NO},
+};
+
+use crate::window::{ProgressBarState, ProgressState};
+
+/// Set progress indicator in the Dock.
+pub fn set_progress_indicator(progress_state: ProgressBarState) {
```
amrbashir, line 15: could we instead port this logic from electron https://github.com/electron/electron/blob/b142fce229c5821083ad6b7c9aa6aa5790032127/shell/browser/native_window_mac.mm#L1270 ? it is simpler and doesn't require manual drawing

amrbashir (changes requested):

src/platform_impl/macos/progress_bar.rs
lucasfernog: the implementation actually looks similar. Electron is also using a custom redraw function: https://github.com/electron/electron/blob/b142fce229c5821083ad6b7c9aa6aa5790032127/shell/browser/native_window_mac.mm#L60

I've tried without the custom draw function and it looks a little weird, it only refreshes on the second update.

lucasfernog: Though I can see the native one looks better.

lucasfernog: custom:
Screenshot 2023-07-12 at 13 30 59
native:
Screenshot 2023-07-12 at 13 31 34

pewsheen: Yeah, we can use the native one. It would be much simpler.

lucasfernog: @pewsheen the only issue is I can't seem to synchronize it. To use you just need to swap your custom class with NSProgressIndicator and call `setIndeterminate: NO`, but increasing the progress is never done immediately, I need to update to a higher value again to trigger it. Decreasing the progress works fine though.

lucasfernog: Also indeterminate mode doesn't seem to make a change (with both classes).

pewsheen: > @pewsheen the only issue is I can't seem to synchronize it. To use you just need to swap your custom class with NSProgressIndicator and call `setIndeterminate: NO`, but increasing the progress is never done immediately, I need to update to a higher value again to trigger it. Decreasing the progress works fine though.

Hmm... this is really weird. I randomly set `usesThreadedAnimation` to `NO` and it draws immediately. I'll need some time to figure it out.

> Also indeterminate mode doesn't seem to make a change (with both classes).

 It looks like it should call `startAnimation` to start the animation, but it doesn't seem to work here 😢

lucasfernog: Yeah I tried the animation thing too. I didn't check the threaded animation thing though, weird :D

lucasfernog: Indeed it works, nice :) maybe the issue is we're updating the value on the main thread?

amrbashir: > the implementation actually looks similar. Electron is also using a custom redraw function: [electron/electron@`b142fce`/shell/browser/native_window_mac.mm#L60](https://github.com/electron/electron/blob/b142fce229c5821083ad6b7c9aa6aa5790032127/shell/browser/native_window_mac.mm#L60)
> 
> I've tried without the custom draw function and it looks a little weird, it only refreshes on the second update.

my bad, didn't notice but if we are drawing it manually, I believe we can support the different states like windows instead of treating all variants as Normal?

pewsheen: > Indeed it works, nice :) maybe the issue is we're updating the value on the main thread?

I tried using `DispatchQueue` to send it to other threads and it still not drawing until the next update. 

> my bad, didn't notice but if we are drawing it manually, I believe we can support the different states like windows instead of treating all variants as Normal?

Yes, we can easily change styles like colors for different states. But the animation would be another story 👀

pewsheen: I created a SwiftUI project from Xcode and tested the native `NSProgressIndicator` on the dock tile, the behavior is the same.

It feels like they don't encourage us to apply animation on the dock tile.  I saw some implementations that animated the dock icon by calling `display()` in each animation frame.

amrbashir: > Yes, we can easily change styles like colors for different states. But the animation would be another story 👀

Yeah I think modifying color is enough

lucasfernog: > > Yes, we can easily change styles like colors for different states. But the animation would be another story eyes
> 
> Yeah I think modifying color is enough

I agree.

pewsheen: Ok! I pushed a commit that can change colors for `paused` and `error` states.

I'm really bad at styling, so feel free to edit it :joy:

lucasfernog (approved):

amrbashir (approved):

PR #710, opened 2023-03-06
fix: not get actual ns_view when it's replace by setContentView

BillGoldenWater:
### What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve tauri-apps/tauri#5229
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tao/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary
- [ ] It can be built on all targets and pass CI/CD.

PR #772, opened 2023-07-26
fix(macos): fix `WindowExtMacOS::ns_view` invalid pointer

amrbashir:
continuation of https://github.com/tauri-apps/tao/pull/710

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

### Checklist

### DIFF
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -446,6 +446,7 @@ impl From<WindowAttributes> for SharedState {
 
 pub struct UnownedWindow {
   pub ns_window: IdRef, // never changes
+  pub ns_view: IdRef,   // never changes
   input_context: IdRef, // never changes
   pub shared_state: Arc<Mutex<SharedState>>,
   decorations: AtomicBool,
@@ -535,6 +536,7 @@ impl UnownedWindow {
     let cloned_preferred_theme = win_attribs.preferred_theme.clone();
 
     let window = Arc::new(UnownedWindow {
+      ns_view,
       ns_window,
       input_context,
       shared_state: Arc::new(Mutex::new(win_attribs.into())),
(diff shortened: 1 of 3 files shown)


src/platform_impl/macos/window.rs
```diff
@@ -1441,7 +1441,7 @@ impl WindowExtMacOS for UnownedWindow {
 
   #[inline]
   fn ns_view(&self) -> *mut c_void {
-    *self.ns_view as *mut _
+    unsafe { (*self.ns_window).contentView() }
```
lucasfernog:
```suggestion
    unsafe { (*self.ns_window).contentView() as *mut _ }
```

lucasfernog (changes requested):

lucasfernog (approved):

src/platform_impl/macos/window.rs
```diff
@@ -1441,7 +1441,7 @@ impl WindowExtMacOS for UnownedWindow {
 
   #[inline]
   fn ns_view(&self) -> *mut c_void {
-    *self.ns_view as *mut _
```
lucasfernog, line 1444: I think we should remove the .ns_view field, it can lead to errors and it's referenced in other places too

lucasfernog (changes requested):

src/platform_impl/macos/window.rs
lucasfernog: We can't really do this because the NSView subclass from wry doesn't have the taoState ivar. We get back to that issue about some missing tao events when using wry.

wusyong (approved):

lucasfernog (approved):

PR #800, opened 2023-10-13
fix(linux): Fix `Window::theme` may return incorrect theme

### LINKED ISSUE #799: `Window::theme` may return incorrect theme on Linux
**Describe the bug**

When setting a theme different from system settings at `WindowBuilder::with_theme`, a window created with the builder returns incorrect theme value by `Window::theme`. For example, when 'Light' is set to the system settings and 'Dark' is set to `WindowBuilder::with_theme`, the window's `Window::theme` returns `Theme::Light`.

**Steps To Reproduce**

1. Go to `Settings -> Appearance -> Style` and set 'Light' style
2. Run the following Rust code
3. Confirm the window is rendered with the dark theme
4. Check the `dbg!` output in the terminal window indicates that `Window::theme` returned `Theme::Light`

```rust

use tao::{
  event::{Event, WindowEvent},
  window::WindowBuilder,
};

fn main() {

  let mut window = Some(
    WindowBuilder::new()
      .with_inner_size(tao::dpi::LogicalSize::new(300.0, 300.0))
      .with_min_inner_size(tao::dpi::LogicalSize::new(200.0, 200.0))
      .with_theme(Some(tao::window::Theme::Dark)) // Set Dark theme
      .build(&event_loop)
      .unwrap(),
  );

  dbg!(window.as_ref().unwrap().theme()); // Show the current theme

    println!("{:?}", event);

    match event {
      Event::WindowEvent {
        window_id: _,
        ..
      } => {
        // drop the window to fire the `Destroyed` event
        window = None;
      }
      Event::WindowEvent {
        event: WindowEvent::Destroyed,
        window_id: _,
        ..
      } => {
      }
      Event::MainEventsCleared => {
        if let Some(w) = &window {
          w.request_redraw();
        }
      }
      _ => (),
    }
  });
}
```

Terminal output:

```
[examples/window.rs:25] window.as_ref().unwrap().theme() = Light
```

**Expected behavior**

Expected `dbg!()` output in terminal:

```
[examples/window.rs:25] window.as_ref().unwrap().theme() = Dark
```

**Screenshots**

screenshot

OS: Ubuntu 22.04
Rustc: 1.72
tao: 71d744cf90a6bcc5248561ec0f6f3a53a40e1fd9

- [x] Yes
- [ ] No

**Additional context**

Nothing

rhysd:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #799

### Other information

On Ubuntu 22.04, the default theme name is ["Yaru"](https://github.com/ubuntu/yaru) in both dark mode and light mode. So it is not possible to check the theme mode by suffix like "-dark". The only way I could come up with was remembering what theme was set on building the window and returning the value at `Window::theme`. This PR adopts the idea.

### DIFF
--- a/src/platform_impl/linux/window.rs
+++ b/src/platform_impl/linux/window.rs
@@ -67,7 +67,7 @@ pub struct Window {
   inner_size_constraints: RefCell<WindowSizeConstraints>,
   /// Draw event Sender
   draw_tx: crossbeam_channel::Sender<WindowId>,
-  theme: Theme,
+  preferred_theme: Option<Theme>,
 }
 
 impl Window {
@@ -184,16 +184,10 @@ impl Window {
       window.set_icon(Some(&icon.inner.into()));
     }
 
-    let settings = Settings::default();
-
-    let mut theme = Theme::Light;
-    if let Some(settings) = settings {
+    let preferred_theme = if let Some(settings) = Settings::default() {
       if let Some(preferred_theme) = attributes.preferred_theme {
         match preferred_theme {
-          Theme::Dark => {
-              settings.set_gtk_application_prefer_dark_theme(true);
-              theme = Theme::Dark;
-          }
+          Theme::Dark => settings.set_gtk_application_prefer_dark_theme(true),
           Theme::Light => {
             let theme_name = settings.gtk_theme_name().map(|t| t.as_str().to_owned());
             if let Some(theme) = theme_name {
@@ -209,7 +203,10 @@ impl Window {
           }
         }
       }
-    }
+      attributes.preferred_theme
+    } else {
+      None
+    };
 
     if attributes.visible {
       window.show_all();
@@ -308,7 +305,7 @@ impl Window {
       minimized,
       fullscreen: RefCell::new(attributes.fullscreen),
       inner_size_constraints: RefCell::new(attributes.inner_size_constraints),
-      theme,
+      preferred_theme,
     };
 
     win.set_skip_taskbar(pl_attribs.skip_taskbar);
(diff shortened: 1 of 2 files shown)


src/platform_impl/linux/window.rs
```diff
@@ -773,15 +779,7 @@ impl Window {
   }
 
   pub fn theme(&self) -> Theme {
-    if let Some(settings) = Settings::default() {
-      let theme_name = settings.gtk_theme_name().map(|s| s.as_str().to_owned());
-      if let Some(theme) = theme_name {
-        if GTK_THEME_SUFFIX_LIST.iter().any(|t| theme.ends_with(t)) {
```
wusyong, line 779: If we choose to keep a theme state, we probably don't need this constant.
Can you remove `GTK_THEME_SUFFIX_LIST`'s definition as well?

amrbashir: We still need to check using `GTK_THEME_SUFFIX_LIST` because a developer may not specify an explicit theme to be set so it could be either light or dark but this PR implementation assumes that it will always be light theme.

src/platform_impl/linux/window.rs
```diff
@@ -185,10 +186,14 @@ impl Window {
 
     let settings = Settings::default();
 
+    let mut theme = Theme::Light;
```
amrbashir: Instead of assuming that it is always light we should:
1. change the newly added `window.theme` to be `explicit_theme: Option<Theme>`
2. in the `Window.theme()` getter we check if `self.explicit_theme.unwrap_or_else(|| /* old logic here */)`

amrbashir (changes requested):

src/platform_impl/linux/window.rs
rhysd: OK, that also works in my use case. Since the configuration name is `preferred_theme`, I think it represents the field well. I'll update this branch.

rhysd: I applied the change at 1236933.

rhysd:
I removed two redundant clones (`to_owned`) at 45a4a29ea39128b770433f04fc11421fd2b2423d.

rhysd:
I removed two redundant clones (`to_owned`) at 94bf32546ab00b6b95d3926071dda88dffa53528.

amrbashir (approved):

PR #831, opened 2023-10-27
Fix `WindowEvent::Destroyed` might fire twice on macOS

wusyong:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/src/platform_impl/macos/util/async.rs
+++ b/src/platform_impl/macos/util/async.rs
@@ -20,15 +20,7 @@ use objc::{
 
 use crate::{
   dpi::LogicalSize,
-  event::{Event, WindowEvent},
-  platform_impl::platform::{
-    app_state::AppState,
-    event::EventWrapper,
-    ffi,
-    util::IdRef,
-    window::{get_window_id, SharedState},
-  },
-  window::WindowId,
+  platform_impl::platform::{ffi, util::IdRef, window::SharedState},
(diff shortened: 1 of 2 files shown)


lucasfernog (approved):

src/platform_impl/macos/util/async.rs
```diff
@@ -251,14 +243,9 @@ pub unsafe fn set_focus(ns_window: id) {
 // through the `IdRef` because otherwise it would dereference free'd memory
 pub unsafe fn close_async(ns_window: IdRef) {
   let ns_window = MainThreadSafe(ns_window);
-  Queue::main().exec_async(move || {
+  run_on_main(move || {
     autoreleasepool(move || {
       ns_window.close();
-      let event = Event::WindowEvent {
-        window_id: WindowId(get_window_id(*ns_window.0)),
-        event: WindowEvent::Destroyed,
```
amrbashir, line 259: why was this event removed? is the destroyed event sent somewhere else?

wusyong: This is a workaround we did before.
It should be sent by `windowWillClose` in window's delegate.

PR #833, opened 2023-11-01
fixed transparent windows

Icekey:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### Other information
This fix was found in a [old unmerged winit pullrequest](https://github.com/rust-windowing/winit/issues/2502). This fix removes the buggy inital white background from transparent windows and fixes transparent fullscreen windows too. I tested it on Win10.

### DIFF
--- a/src/platform_impl/windows/window_state.rs
+++ b/src/platform_impl/windows/window_state.rs
@@ -255,6 +255,9 @@ impl WindowFlags {
     if self.contains(WindowFlags::NO_BACK_BUFFER) {
       style_ex |= WS_EX_NOREDIRECTIONBITMAP;
     }
+    if self.contains(WindowFlags::TRANSPARENT) {
+      style_ex |= WS_EX_LAYERED;
+    }
     if self.contains(WindowFlags::CHILD) {
       style |= WS_CHILD; // This is incompatible with WS_POPUP if that gets added eventually.
     }


dklassic:
Does seem to work, thanks for your contribution!

dklassic (approved):

amrbashir:
This PR shouldn't have been merged and should be reverted asap, Layered Windows is complex and has special attributes, in cases, it might be even invisible and cannot be embedded as a child window, so setting this style is not the fix for transparency in this crate. Also Layered Windows delegates the alpha blending to the compositor which is more performant but might conflict with crates like egui, wgpu and so on.

Note that the window is actually transparent, but the surface has not been cleared so you see a white artifact, and to fix this, if you're using a drawing crate like wgpu or egui, you can just clear the surface.

The main problem is that when adding `wry` to a `tao` window, you will still see a white background and I think it is the responsibility of `wry` to clear the surface and I am working on that so the next release of wry should have this fixed and transitively tauri.

amrbashir:
Also one of the reasons that `transparent` example might have artifacts until you resize, is because we enable `shadows` by default for undecorated window, you can disable that when creating the window and then enabling it back up later and the surface should be clear.

dklassic:
@amrbashir sorry about that I'll submit a revert immediately.

Icekey:
@amrbashir good to know that you working on a proper fix. I hope that it will also solve the issue with the [buggy tao title](https://github.com/tauri-apps/tauri/issues/7369), as that is currently main problem when creating an overlay with tauri.

PR #871, opened 2024-01-29
fix(windows): fix mouseup events do not take effect when the data-tauri-d…

linguofeng:
See https://github.com/tauri-apps/tauri/issues/3811

Ref: https://github.com/tauri-apps/tao/commit/f497b5dc828c202376f37ed835d5fd48b1a93530#diff-bb05c5e1236bdaa7f268bcbbbb8cc9f378a48eb3ed480884f0c6471a9cc933edL1029

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1019,10 +1019,12 @@ unsafe fn public_window_callback_inner<T: 'static>(
     }
 
     win32wm::WM_EXITSIZEMOVE => {
-      subclass_input
-        .window_state
-        .lock()
-        .set_window_flags_in_place(|f| f.remove(WindowFlags::MARKER_IN_SIZE_MOVE));
+      let mut state = subclass_input.window_state.lock();
+      if state.dragging {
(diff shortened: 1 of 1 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -1019,10 +1019,12 @@ unsafe fn public_window_callback_inner<T: 'static>(
     }
 
     win32wm::WM_EXITSIZEMOVE => {
-      subclass_input
-        .window_state
-        .lock()
-        .set_window_flags_in_place(|f| f.remove(WindowFlags::MARKER_IN_SIZE_MOVE));
+      let mut state = subclass_input.window_state.lock();
+      if state.dragging {
```
pewsheen, line 1023: `dragging` does not exist in struct `WindowState`

pewsheen (commented):
Will also need to add back `handle_os_dragging` and related functions from https://github.com/tauri-apps/tao/commit/f497b5dc828c202376f37ed835d5fd48b1a93530 if we want to back port the commit.

PR #880, opened 2024-02-21
refactor(linux): remove zbus and use unity launcher APIs directly

amrbashir:

### DIFF
--- a/.changes/bump-msrv.md
+++ b/.changes/bump-msrv.md
@@ -1,0 +1,5 @@
+---
+"tao": minor
+---
+
+Updated the minimum supported Rust version to 1.70.
--- a/.changes/progress-bar-refactor.md
+++ b/.changes/progress-bar-refactor.md
@@ -1,0 +1,5 @@
+---
+"tao": minor
+---
+
+Progress bar on Linux no longer relies on zbus. Changed  `ProgressBarState`'s field `unity_uri` to `desktop_filename`.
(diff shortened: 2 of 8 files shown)


amrbashir:
I have wasted so much time today trying to see if this implementation works but I can't confirm that it works at all, I couldn't also get the zbus implementation to work as well.

lucasfernog (approved):

zeenix:
> I have wasted so much time today trying to see if this implementation works but I can't confirm that it works at all,

This also adds more code and hence complexity but if maintainers are happy, why not? :)

> I couldn't also get the zbus implementation to work as well.

In the future, if you have problems using zbus, do consider asking for help, either in our [issues](https://github.com/dbus2/zbus/issues), [discussions](https://github.com/dbus2/zbus/discussions) or [our Matrix channel](https://matrix.to/#/#zbus:matrix.org).

amrbashir:
@zeenix thank you, it wasn't an issue with zbus itself at all. I simply wasn't able to observe the effects of the implementation on the window manager.

As for the decision to remove zbus, was because it added a bit of size (around 500kb) to the minimalist tauri app. We still use zbus in other places like single-instance and notifications (through notify-rust) plugins but that's optional additions.

zeenix:
> @zeenix thank you, it wasn't an issue with zbus itself at all. I simply wasn't able to observe the effects of the implementation on the window manager.

Ah ok.

> As for the decision to remove zbus, was because it added a bit of size (around 500kb) to the minimalist tauri app.

Fair enough. I'm working on some changes slowly that should help with the size issue.

> We still use zbus in other places like single-instance and notifications (through notify-rust) plugins but that's optional additions.

:+1:

PR #911, opened 2024-04-28
feat: add app activated event for macos

thewh1teagle:
This PR adds Activated event for `macOS`, resolving [#218](https://github.com/tauri-apps/tao/issues/218).

Useful in two cases:

1. When the app is open but no windows are visible, clicking its icon in the dock will emit the `Activated` event.
2. When the app is running in the background with no windows, and you try to open it again, `macOS` usually brings the already open instance to the front (without activating windows, so nothing happens). With this change, the `Activated` event will be emitted in such cases.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -84,6 +84,12 @@ pub enum Event<'a, T: 'static> {
   /// Emitted when the application has been resumed.
   Resumed,
 
+  /// Emitted when the application has been activated.
+  /// ## Platform-specific
+  ///
+  /// - Only available on **macOS**.
+  Activated,
+
   /// Emitted when all of the event loop's input events have been processed and redraw processing
   /// is about to begin.
   ///
@@ -158,6 +164,7 @@ impl<T: Clone> Clone for Event<'static, T> {
       LoopDestroyed => LoopDestroyed,
       Suspended => Suspended,
       Resumed => Resumed,
+      Activated => Activated,
       Opened { urls } => Opened { urls: urls.clone() },
     }
   }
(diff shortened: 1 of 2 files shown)


FabianLars:
Thanks for contributing! I am not one of the main maintainers so take this with a grain of salt, but i've worked with events a while back and i _think_ that this feature should use the `NSApplicationDelegate`'s [shouldHandleReopen](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428638-applicationshouldhandlereopen) method instead. The problem with that is that we'd need to figure out how to return a bool dynamically (or always `false` since the default behavior doesn't seem to apply to tauri apps anyway).

FabianLars:
Ahh, i forgot. We already have a PR for what i just said: https://github.com/tauri-apps/tao/pull/517

thewh1teagle:
Oh I didn't knew about this event, that looks better because we can control the default behavior too.
Does tao / tauri already use winit? (Like mentioned there) So we can merge it?

FabianLars:
No, and winit still doesn't support this. But we've changed the approach since then and started merging PRs that use platform specific apis even if winit doesn't support them yet.

pronebird:
[applicationShouldHandleReopen](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428638-applicationshouldhandlereopen) sounds like a thing used for document based apps.

I found it simpler to rely on [applicationDidBecomeActive](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428577-applicationdidbecomeactive) which fires every time the app activates.

BillGoldenWater:
I think this event isn't a solution for #218, it's not emitted for the click on dock icon, and it's emitted every time the app gains focus

I think it's usage would be pair with [applicationDidResignActive](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428636-applicationdidresignactive?language=objc) and their "will" variant for handle lose/gain focus of entire app

pronebird:
@BillGoldenWater I think you're right, it doesn't trigger if the app is in focus but doesn't have any windows open.

PR #937, opened 2024-06-09
feat: add a function to set theme dynamically

### LINKED ISSUE #582: [feat] Change Window theme dynamically at runtime
> This issue has been upstreamed from https://github.com/tauri-apps/tauri/issues/5279 

### Describe the problem

As far as I can tell from the docs, concerning the theme of the window, there are JS APIs for only (1) [getting the window's current theme](https://tauri.app/v1/api/js/window/#theme), and (2) [listening to the system theme change](https://tauri.app/v1/api/js/window#onthemechanged).

I'm trying to set the window's theme to "dark"/"light" as well when the user chooses the "dark"/"light" theme in my app.

**Is there a way to dynamically change the theme of the window through JavaScript?**

A JS API to change the theme of the window.

### Alternatives considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #891: Migrate from `SetWindowCompositionAttribute` to `DwmSetWindowAttribute`
According to https://learn.microsoft.com/en-us/windows/win32/dwm/setwindowcompositionattribute

> The use of this API is not recommended. Use [DwmSetWindowAttribute](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmsetwindowattribute) instead.

https://github.com/tauri-apps/tao/blob/88e1e32901fff8de329ec948b5b858ed05dcde78/src/platform_impl/windows/dark_mode.rs#L175-L211

(should be opening a blank issue, not from the bug template, sorry about that)

Legend-Master:
Closes #582
Closes #891


Legend-Master:
~~I don't quite understand how `ThemeChanged` works in `tao` to be honest, when should it be fired? We currently only fires it when the window's theme is set to auto/no preference and the theme is actually changed
I think it should be fired whenever the theme changes no matter if the theme is set to `None` or `Some(Theme)`~~ Never mind, only auto will can have theme changes

I'm not sure how to send events from `set_theme`, do we do a `RegisterWindowMessage` and use `PostMessage` here?

src/platform_impl/windows/mod.rs
```diff
@@ -65,10 +65,10 @@ unsafe impl Send for PlatformSpecificWindowBuilderAttributes {}
...
 
 // Cursor name in UTF-16. Used to set cursor in `WM_SETCURSOR`.
-#[derive(Debug, Clone, Copy)]
-pub struct Cursor(pub *const u16);
-unsafe impl Send for Cursor {}
-unsafe impl Sync for Cursor {}
+// #[derive(Debug, Clone, Copy)]
+// pub struct Cursor(pub *const u16);
+// unsafe impl Send for Cursor {}
+// unsafe impl Sync for Cursor {}
```
Legend-Master: Do we remove those?

Legend-Master:
We probably can just mirror the mac and linux implementations from winit (they should be much more straight forward compared to the Windows one 😂), but I can't test them, so I'll just do the Windows one for now

FabianLars:
Moving my discord message here: The macos implementation worked on my machine. 👍

FabianLars:
Linux impl works for me on WSL

tweidinger (approved):
Worked for me on Linux (Arm64/Ubuntu) from light to dark.

pewsheen:
Looks good on macOS.

Just a note that the `ThemeChagned` event does not emit, but I think it has broken for a while.

src/platform_impl/windows/mod.rs
pewsheen: It seems these were removed at the very beginning of this project 👀

pewsheen (approved):
It looks good and works on my mac and Windows. But I want to let @amrbashir double-check the Windows implementation.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -160,40 +165,31 @@ pub fn allow_dark_mode_for_window(hwnd: HWND, is_dark_mode: bool) {
...
-        pvData: &mut is_dark_mode_bigbool as *mut _ as _,
-        cbData: std::mem::size_of_val(&is_dark_mode_bigbool) as _,
-      };
-      let _ = unsafe { set_window_composition_attribute(hwnd, &mut data as *mut _) };
+    } else {
+      let dark_mode = BOOL::from(is_dark_mode);
+      unsafe {
+        let _ = DwmSetWindowAttribute(
+          hwnd,
+          DWMWA_USE_IMMERSIVE_DARK_MODE,
```
amrbashir: `DWMWA_USE_IMMERSIVE_DARK_MODE` is only available on Windows 11 Build 22000 and later, we should keep the `SetWindowCompositionAttribute` API for better compatibility with older windows 10 versions.

Alternatively we could use `DWMWA_USE_IMMERSIVE_DARK_MODE` on Windows 11 and higher.

src/platform_impl/windows/event_loop.rs
```diff
@@ -616,6 +616,12 @@ lazy_static! {
     pub static ref SET_RETAIN_STATE_ON_SIZE_MSG_ID: u32 = unsafe {
         RegisterWindowMessageA(s!("Tao::SetRetainMaximized"))
     };
+    /// Message sent by a `Window` when a new theme is set.
+    /// WPARAM is 1 for dark mode and 0 for light mode.
+    /// and LPARAM is unused.
+    pub static ref SET_THEME_MSG_ID: u32 = unsafe {
```
amrbashir:
```suggestion
    pub static ref EMIT_THEME_MSG_ID: u32 = unsafe {
```

src/platform_impl/windows/event_loop.rs
```diff
@@ -616,6 +616,12 @@ lazy_static! {
...
+        RegisterWindowMessageA(s!("Tao::SetTheme"))
```
amrbashir:
```suggestion
        RegisterWindowMessageA(s!("Tao::EmitTheme"))
```

amrbashir (changes requested):

src/platform_impl/windows/dark_mode.rs
Legend-Master: The official docs on this is not correct, this is available since way before that, I don't have the device to test if it works on all versions after windows 10 build 18362+ but I think it does (works on my machine running windows 10 19045 at least)

amrbashir: I'd rather be safe than sorry I guess. Unlick `SetWindowCompositionAttribue` function, there is no way to reliably check if `DWMWA_USE_IMMERSIVE_DARK_MODE` is supported by the OS and it could lead to undefined behavior or not working at all.

.changes/set-theme.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": "patch"
+---
+
+Add a function `set_theme` to set theme dynamically after the window is created
```
amrbashir:
```suggestion
Add a function `Window::set_theme` to set theme dynamically after the window is created
```

src/window.rs
```diff
@@ -1106,6 +1106,25 @@ impl Window {
     self.window.theme()
   }
 
+  /// Sets the theme for this window.
+  ///
+  /// ## Platform-specific
+  ///
```
amrbashir, line 1112: We should mention that on Linux and macOS, this changes the whole application and not just this window

src/window.rs
```diff
@@ -1106,6 +1106,25 @@ impl Window {
...
+  /// - **iOS / Android:** Unsupported.
+  #[inline]
+  pub fn set_theme(&self, #[allow(unused)] theme: Option<Theme>) {
```
amrbashir, line 1116: We should add an equivalent method on `EventLoop` and `EventLoopWindowTarget`

amrbashir (changes requested):

src/platform_impl/windows/dark_mode.rs
Legend-Master: Are there any docs about `WCA_USEDARKMODECOLORS`? I can't really find any, also `SetWindowCompositionAttribute` should always be there according to https://learn.microsoft.com/en-us/windows/win32/dwm/setwindowcompositionattribute this is there from windows 7, checking if this thing exists doesn't do anything for us I don't think

amrbashir: > checking if this thing exists doesn't do anything for us I don't think

hmm, I guess the check here then is whether we failed to load it from `user32.dll` or not. Just a misunderstanding on my part then.

amrbashir: `WCA_USEDARKMODECOLORS` is not documented, but it exists in older Windows 10 versions, and can be used as a fallback for older versions. 

Since the documentation, says `DWMWA_USE_IMMERSIVE_DARK_MODE` is supported in `Windows 11 Build 22000` and later, I think we should use it as documented. 

We can use `DWMWA_USE_IMMERSIVE_DARK_MODE` for versions older than that but we should test it and confirm first and keep as default fallback `WCA_USEDARKMODECOLORS`.

Legend-Master: I believe `DWMWA_USE_IMMERSIVE_DARK_MODE` works on older versions, I can find a lot of references in a google search about that, but I can't find much about `WCA_USEDARKMODECOLORS`

The docs was pointing to it once but later removed saying this shouldn't be documented that way https://github.com/MicrosoftDocs/sdk-api/pull/966/files

amrbashir: I guess that's not to cause confusion because the value of `DWMWA_USE_IMMERSIVE_DARK_MODE` is different in different operating systems, however `windows-rs` and `windows-sys` are generated on newer versions of Windows and it will always be `20` https://docs.rs/windows-sys/latest/windows_sys/Win32/Graphics/Dwm/constant.DWMWA_USE_IMMERSIVE_DARK_MODE.html.

So based on the documentation you provided we should define our own constant `DWMWA_USE_IMMERSIVE_DARK_MODE_BEFORE_20H1 = 19` on lower versions and use that instead.

Legend-Master: Will do

Legend-Master: Do you know if `UseImmersiveDarkModeColors` even works on older versions? we currently use it as a fallback on version < 18362

amrbashir: I don't remember ever testing it tbh, I think it was something that winit team has implemented, However versions less than 18362 is super old, let's just keep it as is using `UseImmersiveDarkModeColors`.

src/event_loop.rs
```diff
@@ -296,6 +300,19 @@ impl<T> EventLoopWindowTarget<T> {
     #[cfg(any(target_os = "linux", target_os = "macos"))]
     self.p.set_progress_bar(_progress)
   }
+
+  /// Sets the theme for this window.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / macOS**: Theme is app-wide and not specific to this window.
```
amrbashir:
```suggestion
```

src/event_loop.rs
```diff
@@ -296,6 +300,19 @@ impl<T> EventLoopWindowTarget<T> {
     #[cfg(any(target_os = "linux", target_os = "macos"))]
     self.p.set_progress_bar(_progress)
   }
+
+  /// Sets the theme for this window.
```
amrbashir:
```suggestion
  /// Sets the theme for the application.
```

src/platform_impl/windows/dark_mode.rs
Legend-Master: I changed it to < 17763 since `DWMWINDOWATTRIBUTE(19)` should work between 17763 and 18985, but sure if it'll work but keeping it on < 17763 for now then

src/event_loop.rs
```diff
@@ -296,6 +300,19 @@ impl<T> EventLoopWindowTarget<T> {
...
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / macOS**: Theme is app-wide and not specific to this window.
+  /// - **Windows:** Unsupported. Use the set_theme function available in Window (Windows can have different themes for different windows)
```
amrbashir: This is not entirely true, we have a state for perferred theme here https://github.com/Legend-Master/tao/blob/c626141feeda2a0f7e703b6501a02a46c99d619f/src/platform_impl/windows/event_loop.rs#L109 and this function could change this state.

src/platform_impl/windows/dark_mode.rs
```diff
@@ -160,40 +165,37 @@ pub fn allow_dark_mode_for_window(hwnd: HWND, is_dark_mode: bool) {
...
+      let dark_mode = BOOL::from(is_dark_mode);
+      unsafe {
+        let _ = DwmSetWindowAttribute(
+          hwnd,
+          dwmwa_use_immersive_dark_mode,
+          &dark_mode as *const BOOL as *const c_void,
+          std::mem::size_of::<BOOL>() as u32,
+        );
+      }
+      if redraw_title_bar {
```
amrbashir: why not always re-render titelbar?

Legend-Master: It's not needed on initial load

src/event_loop.rs
Legend-Master: Is there anyway we can get the windows from here? Just changing the state feels incomplete to be honest

amrbashir: Same as you did with emitting event, you will need to use `PostMessageW`

src/platform_impl/windows/dark_mode.rs
amrbashir: does it have any side effect if always redraw then? it would just make the code simpler

src/event_loop.rs
Legend-Master: I mean we need the `hwnd` or else how can we make the existing windows change their theme?

src/platform_impl/windows/dark_mode.rs
Legend-Master: Slightly worth performance, I would like to keep it this way, but if you think cleaner code is more important I wouldn't against it

amrbashir: nah it is fine. code is not that much worse.

src/event_loop.rs
amrbashir: Posting `WM_SETTINGCHANGE` message to the `HWND` in `EventLoopWindowTarget` should trigger this line https://github.com/tauri-apps/tao/blob/c626141feeda2a0f7e703b6501a02a46c99d619f/src/platform_impl/windows/event_loop.rs#L2072

Legend-Master: I think the `preferred_theme` here is a clone of `EventLoopWindowTarget::preferred_theme`?

Legend-Master: Maybe just document this will only affect new windows?

amrbashir: but why? we have a way to update existing windows as well.

amrbashir: > I think the `preferred_theme` here is a clone of `EventLoopWindowTarget::preferred_theme`?

we can change it to use `Mutex` and clone that mutex instead, so we can change the state later

amrbashir: that ofc depends on if `PostMessageW(null, WM_SETTINGCHANGE)` actually triggers our `WM_SETTINGCHANGE` handler for each window

Legend-Master: Got it, made the change

src/platform_impl/windows/event_loop.rs
```diff
@@ -330,6 +330,12 @@ impl<T> EventLoopWindowTarget<T> {
   pub fn cursor_position(&self) -> Result<PhysicalPosition<f64>, ExternalError> {
     util::cursor_position().map_err(Into::into)
   }
+
+  #[inline]
+  pub fn set_theme(&self, theme: Option<Theme>) {
+    *self.preferred_theme.lock() = theme;
+    let _ = unsafe { PostMessageW(HWND_BROADCAST, *CHANGE_THEME_MSG_ID, WPARAM(0), LPARAM(0)) };
```
amrbashir: Does using `HWND::default()` here not work? because `HWND_BROADCAST` will post to all windows and we want to limit it only to our windows.

Legend-Master: No, I tried it first and it didn't work

Legend-Master: Sorry, didn't see we have `owned_windows` here, oops

Legend-Master:
I don't know if sending `WM_SETTINGCHANGE` ourselves here is fine or not since the wparam and lparam are actually not 0 from the system one, and by the way, if you do this, you need to set `redraw_title_bar` to `true`

amrbashir (approved):

amrbashir:
> I don't know if sending `WM_SETTINGCHANGE` ourselves here is fine or not since the wparam and lparam are actually not 0 from the system one

Doesn't matter since we don't read either of them.

> and by the way, if you do this, you need to set `redraw_title_bar` to `true`

From my testing, this doesn't seem to be needed.

Legend-Master:
@amrbashir the windows one is not working right now, you need to set `redraw_title_bar` to `true`, see my comment above

amrbashir:
I guess Windows 10, really needs to redraw, just changed it to always redraw then

amrbashir (approved):

amrbashir:
BTW, sorry for hijacking the PR, I was just testing a few things out and thought might as well push it instead of requesting another review.

Legend-Master:
All good, feel free to make changes 😃

PR #938, opened 2024-06-12
feat: Add Linux method to use existing GTK window

### LINKED ISSUE #925: Access GTK window before it's mapped
I'm trying to do wl-layer-shell stuff on the window the webview is in.  Per https://github.com/tauri-apps/wry/issues/1253 I am able to access the gtk window, but it's already mapped at that point it seems (is this due to the redraw call?):
```
custom_shell_surface_init: assertion '!gtk_widget_get_mapped (GTK_WIDGET (gtk_window))' failed
GtkWindow is not a layer surface. Make sure you called gtk_layer_init_for_window ()
GtkWindow is not a layer surface. Make sure you called gtk_layer_init_for_window ()
GtkWindow is not a layer surface. Make sure you called gtk_layer_init_for_window ()
GtkWindow is not a layer surface. Make sure you called gtk_layer_init_for_window ()
GtkWindow is not a layer surface. Make sure you called gtk_layer_init_for_window ()
GtkWindow is not a layer surface. Make sure you called gtk_layer_init_for_window ()
GtkWindow is not a layer surface. Make sure you called gtk_layer_init_for_window ()
GtkWindow is not a layer surface. Make sure you called gtk_layer_init_for_window ()
```

I'm not sure the best way to do this... maybe in the long run a `wl_layer_shell` abstraction that could handle other platforms as well would be good.

Either being able to prevent the initial draw/put it off until later (maybe until the event loop starts?) or provide a callback to do stuff with the window earlier on would probably work.

A way to "bring your own window" might be a more flexible solution, for when cross-platform support isn't needed.

**Additional context**

andrewbaxter:
Fix #925

The existing `Window::new()` code can be roughly split into two parts: applying generic attributes to the window and connecting the window to the event loop. These can be fairly cleanly separated, so I did that - the new `new_from_gtk_window` method contains the latter, where `new` now just contains the former and calls `new_from_gtk_window` for the latter.

There are a few attributes that affect both the former and latter parts, and I presume in such cases the former is required for the latter to function properly. It's up to the caller to make sure that those prerequisites are met, but I'd like to add specific guidance to the new attribute struct documentation if this approach looks reasonable. (regarding `transparent`, `cursor_moved`, `fullscreen`, etc)

I also made `set_skip_taskbar` public since it seemed simpler than adding another attribute, but happy to go another way.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -227,6 +227,12 @@ impl<T> Deref for EventLoop<T> {
 }
 
 impl<T> EventLoopWindowTarget<T> {
+  /// Returns the platform-specific window target struct
+  #[inline]
+  pub fn platform_window_target(&self) -> &platform_impl::EventLoopWindowTarget<T> {
+      return &self.p;
+  }
+
   /// Returns the list of all the monitors available on the system.
   #[inline]
   pub fn available_monitors(&self) -> impl Iterator<Item = MonitorHandle> {
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -230,7 +230,7 @@ impl<T> EventLoopWindowTarget<T> {
   /// Returns the platform-specific window target struct
   #[inline]
   pub fn platform_window_target(&self) -> &platform_impl::EventLoopWindowTarget<T> {
-      return &self.p;
+    return &self.p;
   }
 
   /// Returns the list of all the monitors available on the system.
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -227,12 +227,6 @@ impl<T> Deref for EventLoop<T> {
 }
 
 impl<T> EventLoopWindowTarget<T> {
-  /// Returns the platform-specific window target struct
-  #[inline]
-  pub fn platform_window_target(&self) -> &platform_impl::EventLoopWindowTarget<T> {
-    return &self.p;
-  }
-
   /// Returns the list of all the monitors available on the system.
   #[inline]
   pub fn available_monitors(&self) -> impl Iterator<Item = MonitorHandle> {
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -206,7 +206,7 @@ pub trait EventLoopWindowTargetExtUnix {
   // /// The pointer will become invalid when the winit `EventLoop` is destroyed.
   // fn wayland_display(&self) -> Option<*mut raw::c_void>;
 
-  /// Returns the gtk application for this event loop
+  /// Returns the gtk application for this event loop.
   fn gtk_app(&self) -> &gtk::Application;
 }
 
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -18,11 +18,9 @@ pub use crate::platform_impl::x11;
 
 pub use crate::platform_impl::EventLoop as UnixEventLoop;
 use crate::{
+  error::OsError,
   event_loop::{EventLoopBuilder, EventLoopWindowTarget},
-  platform_impl::{
-    x11::xdisplay::XError, EventLoopWindowTarget as UnixEventLoopWindowTarget, Parent,
-    Window as UnixWindow,
-  },
+  platform_impl::{x11::xdisplay::XError, Parent, Window as UnixWindow},
   window::{Window, WindowBuilder},
 };
 
@@ -62,14 +60,14 @@ impl<T> EventLoopBuilderExtUnix for EventLoopBuilder<T> {
 }
 
 /// Additional methods on `Window` that are specific to Unix.
-pub trait WindowExtUnix<T> {
+pub trait WindowExtUnix {
   /// Create a new Tao window from an existing GTK window. Generally you should use
   /// the non-Linux `WindowBuilder`, this is for those who need lower level window access
   /// and know what they're doing.
-  fn new_from_gtk_window(
-    event_loop_window_target: &UnixEventLoopWindowTarget<T>,
+  fn new_from_gtk_window<T: 'static>(
+    event_loop_window_target: &EventLoopWindowTarget<T>,
     window: gtk::ApplicationWindow,
-  ) -> UnixWindow;
+  ) -> Result<Window, OsError>;
 
   /// Returns the `gtk::ApplicatonWindow` from gtk crate that is used by this window.
   fn gtk_window(&self) -> &gtk::ApplicationWindow;
@@ -82,7 +80,7 @@ pub trait WindowExtUnix<T> {
   fn set_skip_taskbar(&self, skip: bool);
 }
 
-impl<T> WindowExtUnix<T> for Window {
+impl WindowExtUnix for Window {
   fn gtk_window(&self) -> &gtk::ApplicationWindow {
     &self.window.window
   }
@@ -95,11 +93,17 @@ impl<T> WindowExtUnix<T> for Window {
     self.window.set_skip_taskbar(skip);
   }
 
-  fn new_from_gtk_window(
-    event_loop_window_target: &UnixEventLoopWindowTarget<T>,
+  fn new_from_gtk_window<T: 'static>(
+    event_loop_window_target: &EventLoopWindowTarget<T>,
     window: gtk::ApplicationWindow,
-  ) -> UnixWindow {
-    UnixWindow::new_from_gtk_window(event_loop_window_target, window)
+  ) -> Result<Window, OsError> {
+    let window = match UnixWindow::new_from_gtk_window(&event_loop_window_target.p, window) {
+      Ok(w) => w,
+      Err(e) => {
+        return Err(os_error!(e));
+      }
+    };
+    Ok(Window { window: window })
   }
 }
 
@@ -190,7 +194,7 @@ impl WindowBuilderExtUnix for WindowBuilder {
 }
 
 /// Additional methods on `EventLoopWindowTarget` that are specific to Unix.
-pub trait EventLoopWindowTargetExtUnix<T> {
+pub trait EventLoopWindowTargetExtUnix {
   /// True if the `EventLoopWindowTarget` uses Wayland.
   fn is_wayland(&self) -> bool;
 
@@ -207,14 +211,11 @@ pub trait EventLoopWindowTargetExtUnix<T> {
   // /// The pointer will become invalid when the winit `EventLoop` is destroyed.
   // fn wayland_display(&self) -> Option<*mut raw::c_void>;
 
-  /// Returns the linux-specific window target struct
-  fn window_target(&self) -> &UnixEventLoopWindowTarget<T>;
-
   /// Returns the gtk application for this event loop
   fn gtk_app(&self) -> &gtk::Application;
 }
 
-impl<T> EventLoopWindowTargetExtUnix<T> for EventLoopWindowTarget<T> {
+impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
   #[inline]
   fn is_wayland(&self) -> bool {
     self.p.is_wayland()
@@ -248,10 +249,6 @@ impl<T> EventLoopWindowTargetExtUnix<T> for EventLoopWindowTarget<T> {
   //         _ => None,
   //     }
   // }
-  #[inline]
-  fn window_target(&self) -> &UnixEventLoopWindowTarget<T> {
-    return &self.p;
-  }
 
   #[inline]
   fn gtk_app(&self) -> &gtk::Application {
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -97,12 +97,7 @@ impl WindowExtUnix for Window {
     event_loop_window_target: &EventLoopWindowTarget<T>,
     window: gtk::ApplicationWindow,
   ) -> Result<Window, OsError> {
-    let window = match UnixWindow::new_from_gtk_window(&event_loop_window_target.p, window) {
-      Ok(w) => w,
-      Err(e) => {
-        return Err(os_error!(e));
-      }
-    };
+    let window = UnixWindow::new_from_gtk_window(&event_loop_window_target.p, window)?;
     Ok(Window { window: window })
   }
 }
@@ -252,7 +247,7 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
 
   #[inline]
   fn gtk_app(&self) -> &gtk::Application {
-    return &self.p.app;
+    &self.p.app
   }
 }
 
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -19,7 +19,10 @@ pub use crate::platform_impl::x11;
 pub use crate::platform_impl::EventLoop as UnixEventLoop;
 use crate::{
   event_loop::{EventLoopBuilder, EventLoopWindowTarget},
-  platform_impl::{x11::xdisplay::XError, Parent},
+  platform_impl::{
+    x11::xdisplay::XError, EventLoopWindowTarget as UnixEventLoopWindowTarget, Parent,
+    Window as UnixWindow,
+  },
   window::{Window, WindowBuilder},
 };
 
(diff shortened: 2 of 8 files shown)


src/platform_impl/linux/window.rs
```diff
@@ -68,15 +68,24 @@ pub struct Window {
...
+pub struct FromGtkWindowAttributes {
+  pub transparent: bool,
+  /// Whether the window should be set as fullscreen upon creation.
+  ///
+  /// The default is `None`.
+  pub fullscreen: Option<Fullscreen>,
+  pub cursor_moved: bool,
+  /// The window size constraints
+  pub inner_size_constraints: WindowSizeConstraints,
+}
```
amrbashir: No need for a new struct, we can just use the normal `WindowAttributes` and `PlatformSpecificWindowBuilderAttributes`

src/platform_impl/linux/window.rs
```diff
@@ -233,6 +242,43 @@ impl Window {
       signal_id.borrow_mut().replace(id);
     }
 
+    let mut win = Self::new_from_gtk_window(
```
amrbashir: I am not sure what is the main use case for this new function, is it expected to just take the gtk window and wire it up to the eventloop or should it apply other attributes like setting preferred theme and window icon?

I prefer the latter, and so `new_from_gtk_window` should include all the logic from line 99 to line 245

amrbashir (changes requested):
Thanks for contribution, I marked the PR as draft for you, you can mark it as ready for review later when you're done.

I left a few comments but the direction you're going with LGTM

src/platform_impl/linux/window.rs
andrewbaxter: So for my immediate use case that would work, but I was thinking in the more general sense this would be a way to have interoperability with other ecosystems.

Like: maybe there's a library there that builds a window with various settings and returns it, and then the user wants to attach a webview to it.

In that case, if this method took the attributes and applied them, it could potentially undo or apply conflicting settings to the window.  The user could work around it but they'd have to be very careful to make all the attributes match what the library set where there's overlap, and it would make it more fragile if the upstream library changes or if Tao changes (to re-sync their attributes to the behaviors of both).

Why do you prefer the latter here? Are some of those attributes critical for correct webview operation? FWIW I'm happy to change this, but just want to make sure the discussion has covered all the issues first.

andrewbaxter:
Thanks for the review! And whoops, I thought `Draft: ` would make it a draft like in gitlab... getting rusty here.

src/platform_impl/linux/window.rs
amrbashir: I still think I'd prefer if we took the gtk window and applied the different attributes so it feels closer to what we would create, however I see you have a valid use-case about interoperability with other libraries, and so we could add both APIs.

Or actually we could go with only one API but apply no attributes at all, and just hook it to the event loop and so it will be the responsibility of the gtk window provider to apply the different attributes they need. This makes it a lot more predictable IMO.

andrewbaxter: I like the one API idea, how were you thinking that would work with `WindowBuilder` in that case? Is one api = `new_from_gtk_window` only (named `new`) and the existing `new` would be replaced?

My understanding is that right now `WindowBuilder` calls `new` with attributes and results in similar behavior on all platforms for a given set of attributes. `Window::new_from_gtk_window` would only be called by linux developers who specifically need advanced gtk customization.  So if the attribute application doesn't go in `new` it'd need to be in some other platform-specific function that gets called by `WindowBuilder`, I presume.

amrbashir: yeah I would only add `Window::new_from_gtk_window` which will only hook to the event loop, and return a tao `Window`, we won't provide a `WindowBuilder` variant nor we will change the behavior of `Window::new`

andrewbaxter: Ah okay, got it.  So we'll have `new_from_gtk_window` which is like what was added here, and I'll revert the changes to `new` so it's exactly the same as before. Thanks for the clarification!

src/platform_impl/linux/window.rs
andrewbaxter: So since the other discussion's been resolved, I just wanted to confirm which of these I should do here:
1. Keep `FromGtkWindowAttributes` as it is now
2. Replace `FromGtkWindowAttributes` with `Attributes` and `PlatformAttributes` and ignore most of the fields
3. Remove `FromGtkWindowAttributes` and the code that does stuff with it: `transparent` `fullscreen` etc. When creating the Tao `Window` I'll use default values for `inner_size_constraints` and `fullscreen`

Thanks again for taking such a detailed look at this, I really appreciate it.

andrewbaxter: I guess ignoring `inner_size_constraints` seems fine, I can't find it being used anywhere.

AFAICT `fullscreen` is only used in `Drop`, and that could reasonably be the caller's responsibility to do if necessary.

amrbashir: I would remove this altogether since we won't modify the window at all, we will just hook our internal events tracking and that's it

andrewbaxter: Okay great, will do! Thanks again.

src/platform_impl/linux/window.rs
andrewbaxter: Done.

src/platform_impl/linux/window.rs
andrewbaxter: Done.

andrewbaxter:
I made the changes above, and made a few other small changes while testing:
- Make `platform_impl` pub to access linux `Window` - the name feels kind of internal-ish, but the `pub` contents seems to be limited to types that are public and exposed elsewhere, so I don't think this is exposing sensitive internals
- Methods to get `platform_impl::EventLoopWindowTarget` from `event_loop::EventLoopWindowTarget` and to get the gtk Application from that (required to create the gtk window)

~FWIW I think `event_loop.rs` and a couple other files I touched here aren't rustfmt'd (2 space indent, missing newlines) so I had to be careful when making changes - I think I've undone all the formatting but apologies in advance if I missed something.~ I'm not sure what happened, it looks like rust-analyzer ignored rustfmt.toml and reformatted everything, but now it's working again (and I fixed the formatting issues).

I think this is ready to go now.

src/event_loop.rs
```diff
@@ -227,6 +227,12 @@ impl<T> Deref for EventLoop<T> {
 }
 
 impl<T> EventLoopWindowTarget<T> {
+  /// Returns the platform-specific window target struct
+  #[inline]
+  pub fn platform_window_target(&self) -> &platform_impl::EventLoopWindowTarget<T> {
+    return &self.p;
+  }
```
amrbashir: If I understand correctly, this is exposed so you can get the gtk Application used by the event loop, in which case this is not the ideal way to expose this. Generally, when exposing a platform specific method, we expose them through a platform specific traits, which can be found in https://github.com/tauri-apps/tao/blob/dev/src/platform/unix.rs

src/platform_impl/linux/event_loop.rs
```diff
@@ -61,6 +61,12 @@ pub struct EventLoopWindowTarget<T> {
 }
 
 impl<T> EventLoopWindowTarget<T> {
+  /// Returns the gtk application for this event loop
+  #[inline]
+  pub fn app(&self) -> &gtk::Application {
+    return &self.app;
+  }
```
amrbashir: same as a above

src/platform_impl/linux/window.rs
```diff
@@ -314,6 +314,86 @@ impl Window {
     Ok(win)
   }
 
+  /// Create a new Tao window from an existing GTK window. Generally you should use
+  /// the non-Linux `WindowBuilder`, this is for those who need lower level window access
+  /// and know what they're doing.
+  pub fn new_from_gtk_window<T>(
```
amrbashir: marking this function public here doesn't mean it is actually public, you will need to expose it on `WindowExtUnix` trait in https://github.com/tauri-apps/tao/blob/dev/src/platform/unix.rs

amrbashir (changes requested):

src/event_loop.rs
andrewbaxter: This was needed for `new_from_gtk_window` for things like `window_requests_tx` and `draw_tx`, but I was also using it for getting the application for `gtk::Window::new()`.

Okay thanks for the pointer! I think I've made the relevant changes.

src/platform_impl/linux/event_loop.rs
andrewbaxter: Done.

src/platform_impl/linux/window.rs
andrewbaxter: Done.

andrewbaxter:
With the changes to move methods into the traits, using an external gtk window now needs an invocation like:
```rust
<tao::window::Window as tao::platform::unix::WindowExtUnix<UserEvent>>::new_from_gtk_window(...)
```
which is pretty unintuitive.  I'm not sure there's an easy way around that, but is there somewhere I could document that or provide an example?

amrbashir:
you can juse use import the trait and use directly like this:
```rs
tao::platform::unix::WindowExtUnix;

let window = Window::from_gtk_window();
```

src/platform/unix.rs
```diff
@@ -59,7 +62,15 @@ impl<T> EventLoopBuilderExtUnix for EventLoopBuilder<T> {
 }
 
 /// Additional methods on `Window` that are specific to Unix.
-pub trait WindowExtUnix {
+pub trait WindowExtUnix<T> {
```
amrbashir: No need for a generic here, see bloew

src/platform/unix.rs
```diff
@@ -59,7 +62,15 @@ impl<T> EventLoopBuilderExtUnix for EventLoopBuilder<T> {
...
+  /// Create a new Tao window from an existing GTK window. Generally you should use
+  /// the non-Linux `WindowBuilder`, this is for those who need lower level window access
+  /// and know what they're doing.
+  fn new_from_gtk_window(
+    event_loop_window_target: &UnixEventLoopWindowTarget<T>,
+    window: gtk::ApplicationWindow,
+  ) -> UnixWindow;
```
```suggestion
  fn new_from_gtk_window<T: 'static>(
    event_loop: &EventLoopWindowTarget<T>,
    window: gtk::ApplicationWindow,
  ) -> Self;
```

src/platform/unix.rs
```diff
@@ -172,7 +190,7 @@ impl WindowBuilderExtUnix for WindowBuilder {
 }
 
 /// Additional methods on `EventLoopWindowTarget` that are specific to Unix.
-pub trait EventLoopWindowTargetExtUnix {
+pub trait EventLoopWindowTargetExtUnix<T> {
```
amrbashir: no need to add a generic here as well

src/platform/unix.rs
```diff
@@ -188,9 +206,15 @@ pub trait EventLoopWindowTargetExtUnix {
   // ///
   // /// The pointer will become invalid when the winit `EventLoop` is destroyed.
   // fn wayland_display(&self) -> Option<*mut raw::c_void>;
+
+  /// Returns the linux-specific window target struct
+  fn window_target(&self) -> &UnixEventLoopWindowTarget<T>;
```
amrbashir: this shouldn't be exposed, and where ever this is needed, should be replaced with the user facing type `EventLoopWindowTarget` not the platform specific one

src/platform/unix.rs
```diff
@@ -83,6 +94,13 @@ impl WindowExtUnix for Window {
   fn set_skip_taskbar(&self, skip: bool) {
     self.window.set_skip_taskbar(skip);
   }
+
+  fn new_from_gtk_window(
+    event_loop_window_target: &UnixEventLoopWindowTarget<T>,
+    window: gtk::ApplicationWindow,
+  ) -> UnixWindow {
+    UnixWindow::new_from_gtk_window(event_loop_window_target, window)
+  }
```
amrbashir: this shouldn't return a UnixWindow, instead it should be wrapped in the user facing type `Window` and not the platform specific one

amrbashir: also this method should return a result

andrewbaxter:
I think in this case the generic parameters for `Window::from_gtk_window()` are ambiguous so you need to use the explicit syntax, but good point... I guess rust-analyzer will quick-fix the import and suggest `<Window as ...>` so maybe it's more discoverable than I was thinking at first.

src/platform_impl/linux/window.rs
```diff
@@ -314,6 +314,83 @@ impl Window {
     Ok(win)
   }
 
+  pub(crate) fn new_from_gtk_window<T>(
+    event_loop_window_target: &EventLoopWindowTarget<T>,
+    window: gtk::ApplicationWindow,
+  ) -> Self {
```
amrbashir: this should return a result, even though the logic deoesn;t have anything that propagates errors, we still should return result to make it consistent with `Window::new` and avoids breaking change in the futuer.

amrbashir (changes requested):

src/platform/unix.rs
andrewbaxter: Ah okay, I think I figured out what you're getting at.  Removed this method and I think everything works now.

src/platform/unix.rs
andrewbaxter: Sorry, forgot to apply the change but I think I changed this as suggested. Right, since the type generics were eliminated the generic can be on the method itself.

Done.

src/platform/unix.rs
andrewbaxter: Done.

src/platform/unix.rs
andrewbaxter: Done.

src/platform/unix.rs
andrewbaxter: Done.

src/platform_impl/linux/window.rs
andrewbaxter: Done.

andrewbaxter:
I made all the changes! I feel like this is pretty close now but let me know if there's anything else I missed. Thanks again for the reviews!

src/platform/unix.rs
```diff
@@ -224,6 +249,11 @@ impl<T> EventLoopWindowTargetExtUnix for EventLoopWindowTarget<T> {
   //         _ => None,
   //     }
   // }
+
+  #[inline]
+  fn gtk_app(&self) -> &gtk::Application {
+    return &self.p.app;
```
```suggestion
    &self.p.app
```

src/platform_impl/linux/window.rs
```diff
@@ -314,6 +314,83 @@ impl Window {
     Ok(win)
   }
 
+  pub(crate) fn new_from_gtk_window<T>(
+    event_loop_window_target: &EventLoopWindowTarget<T>,
+    window: gtk::ApplicationWindow,
+  ) -> Result<Self, OsError> {
```
```suggestion
  ) -> Result<Self, RootOsError> {
```

src/platform_impl/linux/window.rs
```diff
@@ -32,7 +32,7 @@ use crate::{
 use super::{
   event_loop::EventLoopWindowTarget,
   monitor::{self, MonitorHandle},
-  util, Parent, PlatformSpecificWindowBuilderAttributes,
+  util, OsError, Parent, PlatformSpecificWindowBuilderAttributes,
```
```suggestion
  util, Parent, PlatformSpecificWindowBuilderAttributes,
```

src/platform/unix.rs
```diff
@@ -83,6 +92,19 @@ impl WindowExtUnix for Window {
...
+  fn new_from_gtk_window<T: 'static>(
+    event_loop_window_target: &EventLoopWindowTarget<T>,
+    window: gtk::ApplicationWindow,
+  ) -> Result<Window, OsError> {
+    let window = match UnixWindow::new_from_gtk_window(&event_loop_window_target.p, window) {
+      Ok(w) => w,
+      Err(e) => {
+        return Err(os_error!(e));
+      }
+    };
```
```suggestion
    let window = UnixWindow::new_from_gtk_window(&event_loop_window_target.p, window)?;
```

amrbashir (approved):
Thanks for contribution

amrbashir (approved):

amrbashir:
I had to force push so I can sign your commits, for future contributions, you need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928) for example.

andrewbaxter:
Thanks! Yeah fwiw I sign my commits, but I don't include personal information in my key so github refuses to verify them.

PR #941, opened 2024-06-19
fix(linux): prevent firing duplicate mouse events (fixes #939)

### LINKED ISSUE #939: Pressing a mouse button has duplicit events on Linux
When a mouse button is pressed, the event loop sends "Pressed, Pressed, Released, Released" events on the Linux platform. Sometimes it sends even more than just one duplicit event. This may cause unwanted multi-click in some applications.

To reproduce:
1. Create a Rust project using Tao 
2. Copy the example reproduction code from bellow
3. Click
4. Read the terminal output

```rust
use tao::{
    event::{Event, WindowEvent},
    window::WindowBuilder,
};

fn main() {
    let window = WindowBuilder::new().build(&event_loop).unwrap();

        match event {
            Event::WindowEvent {
                ..
            } => {
                println!("The close button was pressed; stopping");
                *control_flow = ControlFlow::Exit
            }
            Event::MainEventsCleared => {
                window.request_redraw();
            }
            Event::WindowEvent {
                window_id, event, ..
            } => {
                let WindowEvent::MouseInput { .. } = event else {
                    return;
                };
                dbg!(&window_id);
                dbg!(&event);
            }
            _ => (),
        }
    });
}
```

Expected behavior is receiving only one "Pressed" and one "Released" event to match the other platforms. 

**Platform and Versions:**
OS: Linux
Rustc: cargo 1.81.0-nightly (4dcbca118 2024-06-11)
Tao version: 0.28.1

Sanae6:
Prevents the press, release, and motion events from firing twice on Linux, fixing #939.

### DIFF
--- a/.changes/bump.md
+++ b/.changes/bump.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Prevent duplicate mouse press, release, and motion events from firing on Linux (fixes #939)
--- a/.changes/bump.md
+++ b/.changes/bump.md
@@ -1,5 +1,5 @@
 ---
-"tao": patch
+"tao": "patch"
 ---
 
 Prevent duplicate mouse press, release, and motion events from firing on Linux (fixes #939)
--- a/src/platform_impl/linux/event_loop.rs
+++ b/src/platform_impl/linux/event_loop.rs
@@ -640,7 +640,7 @@ impl<T: 'static> EventLoop<T> {
                   }
                 }
               }
-              glib::Propagation::Proceed
+              glib::Propagation::Stop
             });
 
             let tx_clone = event_tx.clone();
@@ -675,11 +675,11 @@ impl<T: 'static> EventLoop<T> {
                 },
               }) {
                 log::warn!(
-                  "Failed to send mouse input preseed event to event channel: {}",
+                  "Failed to send mouse input pressed event to event channel: {}",
                   e
                 );
               }
-              glib::Propagation::Proceed
+              glib::Propagation::Stop
             });
 
             let tx_clone = event_tx.clone();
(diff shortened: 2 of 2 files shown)


wusyong:
Thanks for help fix this issue! Could you add a change file before it can be merged?

Sanae6:
Should be good now :+1:

wusyong (approved):

.changes/bump.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
```
wusyong:
```suggestion
"tao": "patch"
```

Try this instead.

wusyong:
@jbolda Do you know why covector returned error here? It seems fine for me.

amrbashir:
This is because, PR is from a fork, it is a known issue, wet can ignore it for now

src/platform_impl/linux/event_loop.rs
```diff
@@ -640,7 +640,7 @@ impl<T: 'static> EventLoop<T> {
                   }
                 }
               }
-              glib::Propagation::Proceed
+              glib::Propagation::Stop
```
amrbashir, line 643: How does this change affects Widgets added inside the window, do they still receive events? or will be swallowed by these handlers here?

Sanae6: Is there a way I can easily test that?

amrbashir: you can get the gtk window created by tao using https://docs.rs/tao/latest/tao/platform/unix/trait.WindowExtUnix.html#tymethod.gtk_window and then you can create a gtk Button https://docs.rs/gtk/latest/gtk/struct.Button.html and add it to the window using `gtk_window.add()`, then you will need to add these same handlers on the button you created, and ensure:
1. the button can be clicked normally
2. the handlers on the button are triggered

Sanae6: The button worked fine, I checked all three event handlers and they fire correctly as well. Should be good to go!

amrbashir (approved):

Issue #942, opened 2024-06-21
[windows] After minimizing the window by clicking the taskbar icon, the `requestUserAttention` function does not take effect


muwoo:
**Describe the bug**
When I minimize a window by clicking on the taskbar icon on Windows, calling the `requestUserAttention` function does not trigger the taskbar icon to flash.
**Steps To Reproduce**
1. Click the `icon` icon in the taskbar to minimize the window
2. Trigger the execution of the `appWindow.requestUserAttention(UserAttentionType.Informational)` function
3. The taskbar icon will not flash

**Expected behavior**
The taskbar icon will flash

**Screenshots**
![img_v3_02c2_6a2f267c-26a4-4e6b-a697-eb87767706bg](https://github.com/tauri-apps/tao/assets/21073039/0fbdf52a-2574-4509-82c5-28ce86f242d8)

muwoo:
This is a rather strange phenomenon. When I minimize the window through the minimize button on the window, `requestUserAttention` works normally. This phenomenon only occurs when the window is minimized by clicking the icon in the taskbar.

I guess it may be that when the window is minimized through the taskbar, the obtained [active_window_handle](https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/windows/window.rs#L878) does not change.


But I think that no matter which method is used, as long as the window is minimized, when the `requestUserAttention` behavior is triggered, the icon in the taskbar should flash normally. 

```rust
  pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
    let window = self.window.clone();
    let window_flags = self.window_state.lock().window_flags();
    let is_minimized = window_flags.contains(WindowFlags::MINIMIZED);
    let active_window_handle = unsafe { GetActiveWindow() };
    // If the window is already active and not minimized, we don't need to do anything.
    // but if the window is minimized, we need to attention the user.
    if window.0 == active_window_handle && !is_minimized {
      return;
    }
    // ...
}

PR #943, opened 2024-06-21
fix(windows): `Window::request_user_attention` not taking effect after minimizing the window by clicking the taskbar icon

muwoo:
fix(windows): When I minimize a window by clicking on the taskbar icon on Windows, calling the `requestUserAttention` function does not trigger the taskbar icon to flash. #942

### DIFF
--- a/.changes/fix-user-attention-minimized.md
+++ b/.changes/fix-user-attention-minimized.md
@@ -2,4 +2,4 @@
 "tao": patch
 ---
 
-Fix: after minimizing the window by clicking the taskbar icon, the `requestUserAttention` function does not take effect.
+Fix `Window::request_user_attention` not taking effect after minimizing the window by clicking the taskbar icon
--- a/.changes/fix-user-attention-minimized.md
+++ b/.changes/fix-user-attention-minimized.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fix: after minimizing the window by clicking the taskbar icon, the `requestUserAttention` function does not take effect.
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -874,13 +874,15 @@ impl Window {
   #[inline]
   pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
     let window = self.window.clone();
-    let window_flags = self.window_state.lock().window_flags();
-    let is_minimized = window_flags.contains(WindowFlags::MINIMIZED);
     let active_window_handle = unsafe { GetActiveWindow() };
     // If the window is already active and not minimized, we don't need to do anything.
     // but if the window is minimized, we need to attention the user.
-    if window.0 == active_window_handle && !is_minimized {
-      return;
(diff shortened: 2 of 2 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -874,8 +874,12 @@ impl Window {
   #[inline]
   pub fn request_user_attention(&self, request_type: Option<UserAttentionType>) {
     let window = self.window.clone();
+    let window_flags = self.window_state.lock().window_flags();
+    let is_minimized = window_flags.contains(WindowFlags::MINIMIZED);
```
amrbashir: let's move this lock and its check inside the if statement, so locking the mutex can be avoided if the active window is not this window, something like this.

```rs
if window.0 == active_window_handle {
    let window_flags = self.window_state.lock().window_flags();
    let is_minimized = window_flags.contains(WindowFlags::MINIMIZED);
    if !is_minimized {
      return
    }
}
```

.changes/fix-user-attention-minimized.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fix: after minimizing the window by clicking the taskbar icon, the `requestUserAttention` function does not take effect.
```
amrbashir:
```suggestion
Fix `Window::request_user_attention` not taking effect after minimizing the window by clicking the taskbar icon

**Review by amrbashir — CHANGES_REQUESTED** [review]

File: src/platform_impl/windows/window.rs (hunk 1 above)
Comment (muwoo) [reply src/platform_impl/windows/window.rs]: Thanks for your suggestion, it has been modified

File: .changes/fix-user-attention-minimized.md (hunk 2 above)
Comment (muwoo) [reply .changes/fix-user-attention-minimized.md]: Thanks for your suggestion, it has been modified

**Review by amrbashir — APPROVED** [review]

File: src/platform_impl/windows/window.rs [hunk 3]
```diff
@@ -875,8 +875,14 @@ impl Window {
...
     let window = self.window.clone();
     let active_window_handle = unsafe { GetActiveWindow() };
+    // If the window is already active and not minimized, we don't need to do anything.
+    // but if the window is minimized, we need to attention the user.
     if window.0 == active_window_handle {
-      return;
+      let window_flags = self.window_state.lock().window_flags();
+      let is_minimized = window_flags.contains(WindowFlags::MINIMIZED);
+      if !is_minimized {
+        return
```
```suggestion
        return;
```

amrbashir (approved):

amrbashir:
Thank you

amrbashir:
Oh looks like you didn't sign your commits and I can't merge the PR unless they are signed. You need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928) for example.

muwoo:
@amrbashir Okay, I'll try to submit a PR again

PR #945, opened 2024-06-26
chore: Optimize Covector Config

jbolda:
## Motivation

We have made some quality-of-life updates to covector. I am optimizing the config to take advantage of these changes.

## Changes

- use `fetch:check` to query the crate API instead of command which should be less brittle
- remove extra code block backticks on publish
- remove `cargo package` as we no longer want to publish `.crate` to the GitHub Release

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -12,9 +12,10 @@
         }
       },
       "prepublish": [
-        "sudo apt-get update",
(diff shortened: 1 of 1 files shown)


jbolda:
Note that we might be able to remove the `cargo package` based on our current practices, but living this in draft until we decide.

amrbashir (approved):

PR #953, opened 2024-07-16
ci: update unmaintained actions

amrbashir:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
 
     runs-on: ${{ matrix.platform.os }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       - name: Cache cargo folder
         uses: actions/cache@v1
@@ -50,9 +50,9 @@ jobs:
           path: ~/.cargo
           key: ${{ matrix.platform.target }}-cargo-${{ matrix.rust_version }}
 
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          target: ${{ matrix.platform.target }}
+          targets: ${{ matrix.platform.target }}
           toolchain: ${{ matrix.rust_version }}
 
       - name: Install Gtk (ubuntu only)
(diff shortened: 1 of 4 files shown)


wusyong (approved):

.github/workflows/ci.yml
```diff
@@ -42,17 +42,17 @@ jobs:
...
+      - uses: actions/checkout@v4
 
       - name: Cache cargo folder
         uses: actions/cache@v1
         with:
           path: ~/.cargo
           key: ${{ matrix.platform.target }}-cargo-${{ matrix.rust_version }}
 
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
```
tweidinger, line 53: The recommended way is it to use `@stable`.

```suggestion
      - uses: dtolnay/rust-toolchain@stable
```

.github/workflows/ci.yml
```diff
@@ -108,13 +108,13 @@ jobs:
           !contains(matrix.platform.target, 'ios'))
         run: cargo $CMD test --verbose --target ${{ matrix.platform.target }} $OPTIONS --features serde,$FEATURES
 
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
```
tweidinger, line 111: The recommended way is it to use `@nightly `.

```suggestion
      - uses: dtolnay/rust-toolchain@nightly
```

.github/workflows/ci.yml
```diff
@@ -147,12 +145,8 @@ jobs:
     name: fmt check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
```
tweidinger, line 149: The recommended way is it to use `@stable`.

```suggestion
      - uses: dtolnay/rust-toolchain@stable
```

.github/workflows/ci.yml
amrbashir: we are using an explicit toolchain here not `stable`, here is what dtolnay recommends in the action documentation:
> When passing an explicit toolchain as an input instead of @rev, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action.

.github/workflows/ci.yml
amrbashir: will open a PR for this one

.github/workflows/ci.yml
amrbashir: will open a PR for this one

.github/workflows/ci.yml
amrbashir: actually looking further up, we are testing using stable only, so I will change the matrix and update this usage as well

amrbashir: https://github.com/tauri-apps/tao/pull/957

.github/workflows/ci.yml
amrbashir: https://github.com/tauri-apps/tao/pull/957

.github/workflows/ci.yml
amrbashir: https://github.com/tauri-apps/tao/pull/957

PR #955, opened 2024-07-17
chore(deps): Upgrade to `ndk 0.9`

amrbashir:

PR #956, opened 2024-07-17
chore(deps): Upgrade to `ndk 0.9` and fix thread-safety interactions

MarijnS95:
Closes #807
Closes #808
Closes #955

Most changes were copied over from my `wry` PR at https://github.com/tauri-apps/wry/pull/1296.  Most notably a `&ThreadLooper` is now passed to `wry::android_setup()` because we have one, and it gives important safety guarantees when it comes to registering callbacks on this looper, without (unnecessarily!) requiring `Send`.  A thread-local requirement already exists for the `JNIEnv` that is passed around anyway.

Also note that certain workarounds and illogical inverted passes around `key_code()` handling are no longer needed, as the `ndk` crate now passes an `enum` with the raw `i32` around so that the `.into()` conversion for `i32` (the correct type) now becomes lossless.

---

In the end I'm also quite surprised how many linter warnings trigger in this crate, which don't seem to be down to just linting on an older Rust toolchain?

### DIFF
--- a/.changes/ndk-0.9.md
+++ b/.changes/ndk-0.9.md
@@ -1,0 +1,7 @@
+---
+"tao": minor
+---
+
+**Breaking change**: Upgrade `ndk` crate to `0.9` and `ndk-sys` crate to `0.6`.  Types from the `ndk` crate are used in public API surface.
+**Breaking change**: Change `NativeKeyCode::Android(u32)` type to use `i32`, which is the native type used by all Android API.
+**Breaking change**: The `setup` function passed to `android_binding!()` must now take a `&ThreadLooper` instead of `&ForeignLooper`, matching the `wry` change in https://github.com/tauri-apps/wry/pull/1296.
--- a/examples/transparent.rs
+++ b/examples/transparent.rs
@@ -2,6 +2,7 @@
 // Copyright 2021-2023 Tauri Programme within The Commons Conservancy
 // SPDX-License-Identifier: Apache-2.0
 
+#[cfg(windows)]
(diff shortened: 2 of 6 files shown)


amrbashir (approved):

amrbashir (changes requested):
Could you also add a change file in `.changes` directory?

examples/transparent.rs
```diff
@@ -2,6 +2,7 @@
 // Copyright 2021-2023 Tauri Programme within The Commons Conservancy
 // SPDX-License-Identifier: Apache-2.0
 
+#[cfg(windows)]
 use std::{num::NonZeroU32, rc::Rc};
```
MarijnS95, line 6: Note that there are many more "unused import" warnings like this: why the CI not catching those?

.changes/ndk-0.9.md
```diff
@@ -0,0 +1,7 @@
+---
+"tao": major
```
amrbashir:
```suggestion
"tao": minor
```

MarijnS95: Your documentation says to use `major`:

https://github.com/tauri-apps/tao/tree/HEAD/.changes

I cannot help the fact that the documentation doesn't understand what the most significant version is in semver, when major (or even minor) is still at `0`, you should fix that at the documentation level.

amrbashir: yeah that file need to be updated, it was generated from the changes tool (`covector`) and never really had much refactors.

amrbashir (approved):

amrbashir:
Thank you

PR #970, opened 2024-08-28
fix: return error instead of panic in `set_skip_taskbar` method

amrbashir:
ref: https://github.com/tauri-apps/tauri/issues/10422#issuecomment-2315014576

### DIFF
--- a/src/platform/unix.rs
+++ b/src/platform/unix.rs
@@ -18,7 +18,7 @@ pub use crate::platform_impl::x11;
 
 pub use crate::platform_impl::EventLoop as UnixEventLoop;
 use crate::{
-  error::{OsError, ExternalError},
+  error::{ExternalError, OsError},
   event_loop::{EventLoopBuilder, EventLoopWindowTarget},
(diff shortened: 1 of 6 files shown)


src/platform/unix.rs
```diff
@@ -18,7 +18,7 @@ pub use crate::platform_impl::x11;
 
 pub use crate::platform_impl::EventLoop as UnixEventLoop;
 use crate::{
-  error::OsError,
+  error::{OsError, ExternalError},
```
```suggestion
  error::{ExternalError, OsError},
```

src/platform_impl/linux/window.rs
```diff
@@ -920,13 +920,15 @@ impl Window {
...
-  pub fn set_skip_taskbar(&self, skip: bool) {
+  pub fn set_skip_taskbar(&self, skip: bool) -> Result<(), ExternalError> {
     if let Err(e) = self
       .window_requests_tx
       .send((self.window_id, WindowRequest::SetSkipTaskbar(skip)))
     {
       log::warn!("Fail to send skip taskbar request: {}", e);
     }
+
+    ok(())
```
```suggestion
    Ok(())
```

pewsheen (approved):

PR #979, opened 2024-09-23
fix(wayland): add client side decorations & fix error protocol 71

### LINKED ISSUE #977: [bug][linux][WSL] Wayland protocol error 71 when restoring maximized window state
**Describe the bug**

```
Gdk-Message: Error 71 (Protocol error) dispatching to Wayland display.

`wl_display@1.error(xdg_wm_base@20, 4, "xdg_surface buffer (1160 x 757) does not match the configured maximized state (0 x 0)"

[3521097-278] -> xdg_toplevel@22.set._maximized()
[3521097.300] -> xdg_toplevel@22.unset_fullscreen()
[3521098.399] xdg_toplevel@22.configure(0, 0, array[4])
[3521098.410] xdg_surface@21.configure (900)
[3521101.5681] → xdg_toplevel@22.set_min_size(900, 587)
[3521101.586] -> xdg_toplevel@22.set_max_size (3840, 2160)
[3521101.617] -> xdg_surface@21.set_window_geometry(0, 0, 1160, 757)
```

I'm encountering a Wayland protocol error when restoring a maximized window state using tauri-plugin-window-state plugin on WSL. This issue is based on https://github.com/gitbutlerapp/gitbutler/pull/4864 and the [xdg-shell.xml](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml) specification.

According to the xdg-shell.xml specification:

```xml
<request name="set_window_geometry">
  <description summary="set the new window geometry">
    The window geometry of a window is its "visible bounds" from the user's perspective. It includes title bars and borders if they are drawn by the client, but excludes drop shadows. The geometry is double-buffered, and will be applied at the time wl_surface.commit of the corresponding wl_surface is called.

    **A client cannot set the window geometry of a maximized or fullscreen window, as the window dimensions are determined by the compositor.**

    When applied, the effective window geometry will be the rectangular region of the visible area of the window.
  </description>
  <!-- ... -->
</request>
```

In the Wayland protocol, the compositor determines the size of a maximized window, and the client must adhere to this size. If a client attempts to change the size of a maximized window, it is considered a protocol violation, resulting in an error.

**Steps To Reproduce**

https://github.com/gitbutlerapp/gitbutler/issues/4039
https://github.com/tauri-apps/tauri/issues/10702

1. The maximized window state in Wayland has 0 x 0.
2. The saved window size is 1167 x 575.
3. When trying to restore the window to the saved size, a resize attempt is made.
4. This leads to a surface mismatch, causing the display server to terminate.

**Expected behavior**

The window restoration logic should match the inner size and the window size when in the maximized state on Wayland.

OS: WSL, ubuntu 22.04
Rustc: 1.71.0

**Additional context**

In my opinion, it seems necessary to modify the resize-related code in `window.rs` or `event_loop.rs` within the tao source code for Linux. I'd be happy to work on a fix for this issue if you think it's appropriate.

Zamoca42:
Fixes #977

## Reproduce Errror Protocol 71

In the resizable example, when `resizable` is set to `false` and `with_maximized(true)` is used, it works fine on macOS, but an “Error Protocol 71” occurs on Linux using Wayland. 
This behavior seems a bit unusual.

### MacOS

reproduce-in-mac

### Wayland on Linux

reproduce-in-wayland

## Changes

1. Changed the event handling for maximizing to process events sequentially to avoid conflicts.
~~1. Confirmed that the resize state is true when maximizing.~~
3. Added buttons for maximize and minimize in the Wayland title bar.
4. Modified the event loop for resizing in Wayland:
    - Fixed an issue where the window was not resizing when dragging the window borders.
    - Noted that, except for window.window(), Wayland may not have the concept of GdkWindow like X11.
    - Managed the resizable state internally and removed the conditions.

### ~~1. Added conditions for maximizing in the Linux environment.~~

~~To prevent “Error Protocol 71,” I added a condition to check the resizable state when maximizing in the Linux environment.~~

### 2. Wayland Header

While X11 has built-in headers for maximize and minimize buttons, Wayland does not have these buttons by default. Therefore, I added headers specifically for Wayland.

#### before
before

#### after
after

### 3. resizing in Wayland

Unlike X11, in Wayland, even when the `resizable` condition is set to true, resizing was not possible, and moving the window was also not allowed. However, by modifying the conditions in the event loop, window movement is now enabled.

#### before
before-border
before-move

#### after

after-move
after-border

### ~~Issues still not resolved~~

~~In Linux, when creating a window, it should maximize on the screen with `with_maximized(true)`, just like on macOS, even if `resizable` is set to `false`.~~


src/platform_impl/linux/event_loop.rs
```diff
@@ -421,6 +422,33 @@ impl<T: 'static> EventLoop<T> {
                 | EventMask::SCROLL_MASK,
             );
 
+            const TITLEBAR_HEIGHT: f64 = 30.0;
```
amrbashir: this feels like a magic number, shouldn't we query and check the actual height of the header bar?

src/platform_impl/linux/event_loop.rs
```diff
@@ -462,35 +490,29 @@ impl<T: 'static> EventLoop<T> {
               glib::Propagation::Proceed
             });
             window.connect_button_press_event(move |window, event| {
-              if !window.is_decorated()
-                && window.is_resizable()
-                && !window.is_maximized()
```
amrbashir: why was these checks removed?

src/platform_impl/linux/window.rs
```diff
@@ -87,6 +89,10 @@ impl Window {
 
     let window = window_builder.build();
 
+    if is_wayland {
+      WlHeader::setup(&window);
```
amrbashir: how does this look like on Desktop environments that has server side decorations. I think KDE has server side decorations.

src/platform_impl/linux/window.rs
```diff
@@ -569,9 +575,12 @@ impl Window {
   }
 
   pub fn set_maximized(&self, maximized: bool) {
+    let resizable = self.is_resizable();
```
amrbashir, line 592: instead of just ignoring the maximized request, should we instead make the window resizable, maximize it, then make it non-resizable again?

src/platform_impl/linux/window.rs
```diff
@@ -594,7 +603,7 @@ impl Window {
   }
 
   pub fn is_maximizable(&self) -> bool {
-    true
+    self.window.is_resizable() && self.maximized.load(Ordering::Acquire)
```
amrbashir: `is_maximizable` is not the same as `is_maximized`

amrbashir (changes requested):
@Zamoca42 thank you, this looks awesome, left a few comments

src/platform_impl/linux/event_loop.rs
Zamoca42: Interestingly, the behavior was the same in both Wayland and X11, regardless of whether the condition was present or not. Since the functionality was identical, I removed the condition.

src/platform_impl/linux/window.rs
Zamoca42: In the KDE environment, it looks like this. The button icons are slightly different.
스크린샷 2024-09-24 오전 3 54 42

src/platform_impl/linux/window.rs
Zamoca42: In the event loop, using window.set_resizable(true); followed immediately by window.maximize(); did not work. However, adding a delay of 0.1 seconds between the two calls made it work.

```rust
      WindowRequest::Maximized(maximized, resizable) => {
            if maximized {
              window.set_resizable(true);

              let window_clone = window.clone();
              glib::timeout_add_local(std::time::Duration::from_millis(100), move || {
                window_clone.maximize();
                glib::ControlFlow::Break
              });

              let window_clone = window.clone();
              glib::timeout_add_local(std::time::Duration::from_millis(200), move || {
                window_clone.set_resizable(resizable);
                glib::ControlFlow::Break
              });
            } else {
              window.unmaximize();
            }
          }
```

While it’s not precise, in Wayland, a gap between events is necessary when committing a buffer.

src/platform_impl/linux/event_loop.rs
Zamoca42: I’ve tried the following:
```rust
let full_height = window.allocated_height();
let inner_height = window.size().1;
let header_height = full_height - inner_height;
```
this results in significant gaps in the length of the existing header.

src/platform_impl/linux/window.rs
Zamoca42: Oh, I thought that would affect the `maximized` state. I will revert it back.

Zamoca42: changed it back to `true`.

src/platform_impl/linux/window.rs
Zamoca42: I noticed that processing multiple events simultaneously causes conflicts. 
I modified the code to use `WindowMaximizeProcess` to handle `resizable` and `maximize` sequentially when creating a new window and during the `WindowRequest::Maximized` event.

src/platform_impl/linux/event_loop.rs
Zamoca42: Using `window.titlebar()`, I was able to retrieve the exact size of the header. 
I modified the code to get the title bar height without using magic numbers.

Zamoca42:
## Additional Changes

Now, the maximize event is not ignored even when `resizable` is set to `false` and `maximized` is set to `true`.

maximize-after

src/platform_impl/linux/event_loop.rs
amrbashir: I would still keep them, I remember some window managers didn't enforce the same restrictions.

src/platform_impl/linux/window.rs
amrbashir: apologies, could you show how they compare to another Gtk app? maybe gnome-disks

examples/resizable.rs
amrbashir, line 1: changes to this example should be reverted as this example is solely about resizable. For  testing combination of flags, see the `window_debug` example

src/platform_impl/linux/event_loop.rs
```diff
@@ -421,6 +427,33 @@ impl<T: 'static> EventLoop<T> {
                 | EventMask::SCROLL_MASK,
             );
 
+            const LMB: u32 = 1;
+            // Wayland add Events
+            if is_wayland {
+              window.connect_button_press_event(move |window, event| {
```
amrbashir: instead of detecting click in the whole window, can't we just attach this event on the `HeaderBar` itself when creating it, in `wayland/display.rs`

src/platform_impl/linux/wayland/display.rs
amrbashir, line 1: I think this file should be renamed to `titlebar.rs` or `header.rs`

src/platform_impl/linux/window.rs
Zamoca42: No problem! I ran firefox and gnome-disks. The button styles do look a bit different.
스크린샷 2024-09-25 오전 1 12 30
스크린샷 2024-09-25 오전 1 13 44

src/platform_impl/linux/window.rs
```diff
@@ -102,11 +108,13 @@ impl Window {
     window.set_default_size(1, 1);
     window.resize(width, height);
 
-    if attributes.maximized {
-      window.maximize();
-    }
+    let maximize_process =
+      util::WindowMaximizeProcess::new(window.clone(), attributes.maximized, attributes.resizable);
+    glib::timeout_add_local(std::time::Duration::from_millis(25), move || {
```
amrbashir: instead of using a `timeout_add_local`, maybe it is better to use `idle_add_full` with a priority of `PRIORITY_HIGH_IDLE + 10` so it runs on the next iteration instead of a timeout

see https://docs.gtk.org/glib/func.idle_add_full.html and https://docs.gtk.org/glib/const.PRIORITY_HIGH_IDLE.html

amrbashir (changes requested):

src/platform_impl/linux/window.rs
amrbashir: that's unfortunate, I will look around for a bit and see if we can replicate their behavior.

src/platform_impl/linux/event_loop.rs
Zamoca42: Apologies, I made a mistake. After adding those conditions back and running it again, the conditions worked correctly. However, I’m confused about `!window.is_decorated()`. On macOS, it’s possible to resize from the window borders even without window decorations. 

Shouldn’t it be `window.is_decorated()` instead of `!window.is_decorated()` if we want to allow resizing when the window is decorated?

amrbashir: macOS is the only OS that allows resizing even thought the window is without decorations but on Linux and Windows, we have to manually hit test and start resizing on our own.

`!window.is_decorated()` is the correct one, since we only want to do manual hit testing when the window is NOT decorated.

src/platform_impl/linux/wayland/display.rs
Zamoca42: it seems more aligned with the GUI, I will change it to `header.rs`.

examples/resizable.rs
Zamoca42: Apologies, I will revert the resizable example. Is it okay to add resizable flags to window_debug?

amrbashir: yup, that's fine

src/platform_impl/linux/event_loop.rs
Zamoca42: I still don’t fully understand, but if that condition exists, resizing does not work on Linux (Wayland) when the window is decorated. Do we need a separate resize event for Wayland?

resize-condition

src/platform_impl/linux/window.rs
Zamoca42: I found a way to align the button styles between non-KDE Wayland and KDE.  
This works well 😄

```rust
let header = HeaderBar::builder()
    .show_close_button(true)
    .decoration_layout("menu:minimize,maximize,close")
    .build();
```

ref: https://forum.manjaro.org/t/no-maximize-minimize-buttons-on-flatpak-wayland-gtk-applications/131742

and now, here’s how it looks:

### non-kde wayland
스크린샷 2024-09-25 오후 7 51 39

#### kde
스크린샷 2024-09-25 오후 7 48 04

examples/resizable.rs
Zamoca42: Revert example `resizable` and add the resizable flag to `window_debug`.

src/platform_impl/linux/wayland/display.rs
Zamoca42: rename `display.rs` to `header.rs`

src/platform_impl/linux/event_loop.rs
Zamoca42: In `header.rs`, I moved the header drag event to `Self::connect_move_window(&header, &window);`.

src/platform_impl/linux/window.rs
Zamoca42: You’re right! It seems that the priority assigned when registering to the main loop is more important than using a timeout. 
When creating a new window, `Priority::HIGH_IDLE` works, but it causes a "Protocol Error 71" in the event loop. 
So, for the maximize event in the event loop, I applied `Priority::DEFAULT_IDLE`.

src/platform_impl/linux/event_loop.rs
amrbashir: we can just tweak the condition to always enable on wayland, `(is_wayland || window.is_decorated()) && window.is_resizable() && !window.is_maximized()`

src/platform_impl/linux/window.rs
amrbashir: Wow!! good job.

src/platform_impl/linux/wayland/header.rs
```diff
@@ -0,0 +1,44 @@
...
+      .decoration_layout("menu:minimize,maximize,close")
+      .build();
+
+    window.set_titlebar(Some(&header));
+    Self::connect_move_window(&header, &window);
+  }
+
+  fn connect_move_window(header: &HeaderBar, window: &ApplicationWindow) {
+    let header_weak = header.downgrade();
+    window.connect_button_press_event(move |window, event| {
```
amrbashir: why not just `header.connect_button_press_event`?

amrbashir: and then you probably don't need to check 
```
 if x >= window_x as f64
            && x <= (window_x + window_width) as f64
            && y >= window_y as f64
            && y <= (window_y + header_height) as f64
```

amrbashir (changes requested):
This is looking very good and just a few things and should be good to merge, could you add a change file (or multiple change files for each bug your are fixing in this PR) in `.changes` directory?

src/platform_impl/linux/wayland/header.rs
Zamoca42: I tried using `header.connect_button_press_event`. Unfortunately, it didn’t seem to recognize the full width of the header.
I will try to find a way to get the width of the header.

header-drag

amrbashir: maybe try checking the width of the header bar using GTK inspector https://developer.gnome.org/documentation/tools/inspector.html

Zamoca42: Wow, that’s really useful. I wrapped the header in an event box. ~~and the drag event works perfectly.~~
The drag event worked, but the buttons were not updated according to the resizable event.

event-box-work

src/platform_impl/linux/event_loop.rs
Zamoca42: That condition seems to work well.
drag

src/platform_impl/linux/wayland/header.rs
Zamoca42: It seems that using `set_titlebar()` passes events through to the `gtkWindow`.
See: https://stackoverflow.com/questions/44670188/gtk3-widget-mouse-events-pass-through

This is likely the reason why `header.connect_button_press_event` is not working. 
Ultimately, we need to decide whether to use `window.connect_button_press_event` or stick with `event_box`.

Both options work the same.

.changes/improve-wayland-display.md
```diff
@@ -0,0 +1,11 @@
+---
+"tao": patch
+---
+
+Changed the event handling for maximizing to process events sequentially to avoid "Error 71(Protocol error): dispatching to Wayland display".
```
amrbashir:
```suggestion
On Linux, changed the event handling for maximizing to process events sequentially to avoid "Error 71(Protocol error): dispatching to Wayland display".
```

.changes/improve-wayland-display.md
```diff
@@ -0,0 +1,11 @@
...
+
+Added buttons for maximize and minimize in the Wayland title bar.
+
+Fixed an issue where the window was not resizing when dragging the window borders.
```
amrbashir:
```suggestion
On Linux Wayland, fixed an issue where the window was not resizing when dragging the window borders.
```
and let's move it to another change file

.changes/improve-wayland-display.md
```diff
@@ -0,0 +1,11 @@
...
+
+Fixed an issue where the window was not moving when dragging the header bar area.
```
amrbashir:
```suggestion
On Linux wayland, fixed an issue where the window was not moving when dragging the header bar area.
```
and let's move it to another change file

.changes/improve-wayland-display.md
```diff
@@ -0,0 +1,11 @@
+---
+"tao": patch
+---
+
+Changed the event handling for maximizing to process events sequentially to avoid "Error 71(Protocol error): dispatching to Wayland display".
+
+Added buttons for maximize and minimize in the Wayland title bar.
```
amrbashir:
```suggestion
On Linux Wayland, added buttons for maximize and minimize in the title bar.
```
and let's move this to another change file

src/platform_impl/linux/event_loop.rs
```diff
@@ -462,35 +468,34 @@ impl<T: 'static> EventLoop<T> {
               glib::Propagation::Proceed
             });
             window.connect_button_press_event(move |window, event| {
-              if !window.is_decorated()
+              const LMB: u32 = 1;
+              if (is_wayland && window.is_decorated())
```
amrbashir: This should be `||` not `&&` and should be `!window.is_decorated()` because we want to always keep this handling on wayland (because it doesn't have server side decorations) but if not wayland we only handle the non decorated windows

```suggestion
              if (is_wayland || !window.is_decorated())
```

src/platform_impl/linux/util.rs
```diff
@@ -70,3 +74,46 @@ pub fn set_size_constraints<W: GtkWindowExt + WidgetExt>(
...
+  pub fn next_step(&mut self) -> glib::ControlFlow {
+    match self.step {
+      0 => {
+        if self.maximized {
+          self.window.set_resizable(true);
+          self.step += 1;
+          glib::ControlFlow::Continue
+        } else {
+          self.step = 2;
+          glib::ControlFlow::Continue
```
amrbashir: should we just break here? why we go to step 2 when we didn't change anything?

amrbashir (changes requested):

src/platform_impl/linux/wayland/header.rs
amrbashir: EventBox seems okay

src/platform_impl/linux/wayland/header.rs
```diff
@@ -0,0 +1,36 @@
...
+
+impl WlHeader {
+  pub fn setup(window: &ApplicationWindow, title: &str) {
+    let header = HeaderBar::builder()
+      .show_close_button(true)
+      .decoration_layout("menu:minimize,maximize,close")
+      .title(title)
+      .build();
+
+    let event_box = EventBox::new();
```
amrbashir, line 13: how about instead of using an `EventBox` we use the `HeaderBar` but call `headerbar.add_events(EventMask::BUTTON_PRESS_MASK)`

- https://docs.rs/gtk/latest/gtk/prelude/trait.WidgetExtManual.html#method.add_events
- https://docs.rs/gdk/0.18.0/gdk/struct.EventMask.html

amrbashir:
Btw I haven noticed you didn't sign your commits and I can't merge the PR unless they are signed. You need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928). `git rebase --exec 'git commit --amend --no-edit -n -S' -i dev`

src/platform_impl/linux/wayland/header.rs
Zamoca42: I’ve tried this before, but to show you directly, it doesn’t work.
event-mask

amrbashir: ah, nvm then

src/platform_impl/linux/util.rs
Zamoca42: That was necessary because I removed `window.set_resizable(attributes.resizable)` when creating the new window.

```rust
    let maximize_process =
      util::WindowMaximizeProcess::new(window.clone(), attributes.maximized, attributes.resizable);
    glib::idle_add_local_full(glib::Priority::HIGH_IDLE, move || {
      let mut maximize_process = maximize_process.borrow_mut();
      maximize_process.next_step()
    });

    //window.set_resizable(attributes.resizable); -> remove this
    window.set_deletable(attributes.closable);
```

If we go with this approach:

```suggestion
      glib::ControlFlow::Break
```

It seems like we’ll need to add `window.set_resizable(attributes.resizable);` back again.

amrbashir: yeah let's add that back and tweak the logic a bit to avoid unnecessary work, like if `attributes.maximized` is `false`, just skip the logic

amrbashir: actually let's do it like this
```rs
if attributes.maximized {
  // glib::idle_add_local ...etc
} else {
  window.set_resizable(attributes.resizable);
}
```
and keep `glib::ControlFlow::Continue` and just add note why we are doing that instead of breaking

src/platform_impl/linux/event_loop.rs
Zamoca42: OMG, I’ll change it right now.

src/platform_impl/linux/util.rs
```diff
@@ -70,3 +74,47 @@ pub fn set_size_constraints<W: GtkWindowExt + WidgetExt>(
...
+      0 => {
+        if self.maximized {
+          self.window.set_resizable(true);
+          self.step += 1;
+          glib::ControlFlow::Continue
+        } else {
+          //It should still revert to the previous resizable state.
+          self.step = 2;
+          glib::ControlFlow::Continue
+        }
```
amrbashir: taking another look at the usage of `WindowMaximizeProcess` we seem to be only using it when `maximized` is `true` so we can simplify this to 
```suggestion
       self.window.set_resizable(true);
       self.step += 1;
       glib::ControlFlow::Continue
```

and we can remove `maximized` field from `WindowMaximizeProcess` since it will always be true

amrbashir (changes requested):
Just one more thing to take care of and this should be good to go. Thanks for your patience.

src/platform_impl/linux/util.rs
Zamoca42: Indeed, since all the code using `WindowMaximizeProcess` is wrapped in a `maximized` condition, the suggestion seems better.

Zamoca42: It works well. I will go ahead and revise the code with the suggested changes.

maximize-update

amrbashir (approved):
LGTM, Thank you for this awesome PR

Zamoca42:
@amrbashir Thank you for your review. It was the kindest and most thoughtful review I’ve ever received. 
This was definitely a great experience and an enjoyable task to work on. 👍

amrbashir (approved):

PR #981, opened 2024-09-25
feat: add `Window::is_always_on_top` method 

thep0y:
I couldn't find a way to get the "keep above" property through GTK, so the `is_always_on_top` method is not implemented for Linux.

Relate: https://github.com/tauri-apps/tauri/issues/11078

### DIFF
--- a/.changes/is-always-on-top.md
+++ b/.changes/is-always-on-top.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Add is_always_on_top method to check if a window is always on top on macOS, Linux and Windows.
--- a/.changes/is-always-on-top.md
+++ b/.changes/is-always-on-top.md
@@ -2,4 +2,4 @@
 "tao": patch
 ---
 
-Add is_always_on_top method to check if a window is always on top on macOS, Linux and Windows.
+Add `Window::is_always_on_top` method to check if a window is always on top on macOS, Linux and Windows.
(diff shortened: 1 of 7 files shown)


amrbashir:
for linux, you can check https://docs.rs/gdk/0.18.0/gdk/struct.Window.html#method.state and check if it has thishttps://docs.rs/gdk/0.18.0/gdk/struct.WindowState.html#associatedconstant.ABOVE flag

amrbashir (changes requested):
Other than failing CI because of missing methods for Android and iOS backends, this LGTM.

Could you also add a change file in `.changes` directory?

thep0y:
Why use placeholder functions instead of using compile-time conditions to control platform-specific interfaces?

amrbashir:
> Why use placeholder functions instead of using compile-time conditions to control platform-specific interfaces?

This was something inherited from `winit` crate when we forked it.

.changes/is-always-on-top.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Add is_always_on_top method to check if a window is always on top on macOS, Linux and Windows.
```
amrbashir:
```suggestion
Add `Window::is_always_on_top` method to check if a window is always on top on macOS, Linux and Windows.
```

amrbashir (approved):

amrbashir (approved):

amrbashir:
Thank you

PR #983, opened 2024-09-28
fix(windows): blinking title bar on changing system settings

Legend-Master:
Only redraw title bar on manual set theme

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -2077,7 +2077,7 @@ unsafe fn public_window_callback_inner<T: 'static>(
     }
 
     win32wm::WM_SETTINGCHANGE => {
-      set_theme(subclass_input, window, false);
+      update_theme(subclass_input, window, true);
     }
 
     win32wm::WM_NCCALCSIZE => {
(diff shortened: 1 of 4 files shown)


src/platform_impl/windows/event_loop.rs
```diff
@@ -2072,23 +2077,7 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
-      let mut window_state = subclass_input.window_state.lock();
-
-      if window_state.current_theme != new_theme {
-        window_state.current_theme = new_theme;
-        mem::drop(window_state);
-        subclass_input.send_event(Event::WindowEvent {
-          window_id: RootWindowId(WindowId(window.0 as _)),
-          event: ThemeChanged(new_theme),
-        });
-      }
```
amrbashir, line 2091: I think we should keep this one separate and add back the check that discarded unnecessary logic which was

```rs
if preferred_theme.or(*subclass_input.event_loop_preferred_theme.lock()).is_none() {
  // try set new theme based on OS
  // ...etc
}
```

amrbashir: or keep the new function but pass it a flag to know that it is called from `WM_SETTINGCHANGE` and skip above check matches
```rs
if from_os && preferred_theme.or(*subclass_input.event_loop_preferred_theme.lock()).is_none() {
  return;
}
```

amrbashir (approved):

Issue #561, opened 2022-09-21
[bug] When resizable is false window size is forced to be at least 200px (Linux)


Megamannen:
Using this sample
[resizable](https://github.com/tauri-apps/tao/blob/dev/examples/resizable.rs)

When starting with
```rs
  let mut resizable = true;

  let window = WindowBuilder::new()
    .with_title("Hit space to toggle resizability.")
    .with_inner_size(LogicalSize::new(220.0, 90.0))
    .with_resizable(resizable)
    .build(&event_loop)
    .unwrap();
```
It starts out at the correct size, but when pressing space (triggering resizable change), window size is forced to 200px height

Megamannen:
[tao-resizeable.zip](https://github.com/tauri-apps/tao/files/9613582/tao-resizeable.zip)

Or sample project with the same code as the sample

amrbashir:
cc @wusyong

amrbashir:
Just tested this on my KDE and it works fine, what is your Linux distro and Window manager?

Megamannen:
> Just tested this on my KDE and it works fine, what is your Linux distro and Window manager?

Manjaro, running Gnome Desktop. Can't give you exact version number right now since I'm not at work, but I'm 99% certain it's Gnome 43

amrbashir:
@FabianLars ☝️

FabianLars:
yep, same issue using the zip from above on ubuntu2204 on wsl, and fedora+gnome (both wayland and x11)

JakubGluszek:
Same issue on Ubuntu 22.10 (5.19.0-35-generic) (X11 / i3)

JakubGluszek:
> Same issue on Ubuntu 22.10 (5.19.0-35-generic) (X11 / i3)

I also wanted to add that the issue is present only when the window is in "floating" mode (when using i3 WM).

yuezk:
Same issue here on Ubuntu 22.04, GNOME version: 42.5, Wayland

Daniel-Ros:
Same issue on opensuse, hyprland/wayland

clemux:
Same issue here, both with GNOME 44.2 on Wayland and with i3 on X11. 
(webkit2gtk-4.0: 2.40.4, tauri: 1.4.1)

FoliageOwO:
Same issue, running on `Manjaro Linux` with `Plasma 5.27.10` as DE and `KWin` as WM. When will it be fixed? :worried:

[...]

PR #984, opened 2024-09-29
fix(wayland): fix the inaccurately applied inner_size when creating a new window.

### LINKED ISSUE #929: Setting initial window size on GTK sets the wrong size
**Describe the bug**
I am using tao on weston with CentOS 9. Using the WindowBuilder `with_inner_size` doesn't seem to give the same inner_size.

**Steps To Reproduce**
I have attached an adjusted example from here to show this, just start it and press spacebar a few times.

Firstly it seems like inner_size and outer_size are always equal on Linux?

**Expected behavior**
`with_inner_size(PhysicalSize::new(240, 240))` should make a window with a content area of 240x240. At the moment I get 240x277

OS: CentOS Stream 9, weston 8
Rustc: rustc 1.77.2

**Additional context**
It seems like I can work around this by setting the inner_size again after window decorations have been disabled.

```

use tao::{
  dpi::LogicalSize,
  event::{ElementState, Event, KeyEvent, WindowEvent},
  keyboard::KeyCode,
  window::WindowBuilder,
};

fn main() {
  env_logger::init();

    .with_title("Test")
    .build(&event_loop)
    .unwrap();

  let mut time = 0;

    match event {
        WindowEvent::KeyboardInput {
          event:
            KeyEvent {
              physical_key: KeyCode::Space,
              ..
            },
          ..
        } => {
          println!("inner_size: {:?}, outer_size: {:?}, scale_factor: {}", window.inner_size(), window.outer_size(), window.scale_factor());

          if time == 0 {
            println!("window.set_inner_size(LogicalSize[width: 240, height: 240])");
            window.set_inner_size(LogicalSize{width: 240, height: 240});
          } else if time == 1 {
            println!("window.set_decorations(true)");
            window.set_decorations(true);
          } else if time == 2 {
            println!("window.set_inner_size(LogicalSize[width: 240, height: 240])");
            window.set_inner_size(LogicalSize{width: 240, height: 240});
          } else if time == 3 {
            println!("window.set_decorations(false)");
          } else if time == 4 {
            println!("window.set_inner_size(LogicalSize[width: 240, height: 240])");
            window.set_inner_size(LogicalSize{width: 240, height: 240});
          }

          time = time + 1
        }
        _ => ()
      },
      _ => (),
    };
  });
}

```

Zamoca42:
Fixes #929 

## Reproduce

1. When setting `with_inner_size` to `300 x 300` and running with `cargo run --example window`, a window size discrepancy occurs, resulting in `300 x 261`.

    window

    box

2. The size of the window border is 45px on KDE Wayland, and it’s the same for all apps.
3. This is because the window border is included in the inner_size value, causing a mismatch when `window.resize()`.

The 39px discrepancy in height is due to the inclusion of the titlebar’s height.

스크린샷 2024-09-28 오후 11 41 07

## Changes 

Added width and height settings to the window box properties corresponding to the inner size.
Added a height property to the header so that it is included in the window border, not the inner_size.

See: https://blog.gtk.org/2019/03/27/layout-managers-in-gtk-4/

## Additional Context

In Tao’s Linux window box properties, `width_request = -1`, `height_request = -1` is passed, and when resized to the min_size, the internal window size changes to include the window borders. 
- min_height = 200, 
- box allocation_height = 70 + 45(border) + 85(border + titlebar_height) = 200.

스크린샷 2024-09-29 오후 8 04 38
스크린샷 2024-09-29 오후 7 42 28

### Compared with the Firefox app on KDE Wayland.

In Firefox, the properties are set with values other than `-1` for `size_request`, so when reduced to size, the inner window size does not include the border values.

스크린샷 2024-09-28 오후 11 22 28
스크린샷 2024-09-29 오후 7 12 39
스크린샷 2024-09-29 오후 7 11 51

### Conclusion

To prevent inaccurate `inner_size`, we need to set `size_request` and configure the width and height values in the window properties.


amrbashir (approved):

amrbashir:
Thank you

amrbashir (changes requested):
Actually after a second test, there is a few things I noticed:
1. if you launch the app and cursor is out of the window (i.e the eventloop doesn't recieve any events) the whole window size will be 300x300, once you move the mouse inside the window, it will jump into a bigger size (352x399)

2. if I create a window with 300x300 inner size, then `window.inner_size` should report the same size later on, I think this is a problem in dev as well though

Zamoca42:
> * if you launch the app and cursor is out of the window (i.e the eventloop doesn't recieve any events) the whole window size will be 300x300, once you move the mouse inside the window, it will jump into a bigger size (352x399)
> * if I create a window with 300x300 inner size, then `window.inner_size` should report the same size later on, I think this is a problem in dev as well though

Thank you for your review. After testing the window example on GNOME + Wayland, I found that the event loop fetches the size of the top layout, GtkApplicationWindow (352 x 399), instead of the inner size. If the event loop is intended to receive the inner size rather than the window size including decorations, I believe we need to modify it to fetch the size of its child (GtkBox).

As a side note, I previously mentioned that the total size of the app includes a border width of 45px, but that applies to KDE. I have confirmed that on GNOME, the default border width is 26px.

src/platform_impl/linux/event_loop.rs
```diff
@@ -567,7 +567,13 @@ impl<T: 'static> EventLoop<T> {
...
               }
 
-              let (w, h) = event.size();
+              let (w, h) = if let Some(child) = window.child() {
+                let allocation = child.allocation();
+                (allocation.width() as u32, allocation.height() as u32)
+              } else {
+                event.size()
+              };
+
```
Zamoca42, line 612: In `window.connect_configure_event`, `event.size()` represented the full window size. I have modified it to fetch the child size and display the inner size instead.

## Before the fix
스크린샷 2024-10-01 오후 4 26 58

## After the fix
스크린샷 2024-10-01 오후 4 27 41

src/platform_impl/linux/window.rs
```diff
@@ -325,7 +335,7 @@ impl Window {
       preferred_theme: RefCell::new(preferred_theme),
     };
 
-    win.set_skip_taskbar(pl_attribs.skip_taskbar);
+    let _ = win.set_skip_taskbar(pl_attribs.skip_taskbar);
```
Zamoca42: I fixed it because the following warning message occurred.
스크린샷 2024-10-02 오후 6 40 26

src/platform_impl/linux/window.rs
```diff
@@ -101,17 +107,20 @@ impl Window {
...
+    let window_clone = window.clone();
+    glib::idle_add_local_once(move || {
+      window_clone.set_resizable(true);
+      window_clone.resize(width, height);
+    });
 
     if attributes.maximized {
       let maximize_process = util::WindowMaximizeProcess::new(window.clone(), attributes.resizable);
-      glib::idle_add_local_full(glib::Priority::HIGH_IDLE, move || {
+      glib::idle_add_local_full(glib::Priority::DEFAULT_IDLE, move || {
```
Zamoca42: A dispatching error occurred when maximize was set to HIGH_IDLE, so I changed it to DEFAULT_IDLE. The error occurred even when wrapping the header, resize and maximize in HIGH_IDLE using idle_add, so I changed only maximize to DEFAULT_IDLE.

스크린샷 2024-10-02 오후 6 17 35

src/platform_impl/linux/window.rs
```diff
@@ -101,17 +107,20 @@ impl Window {
       .insert(window_id);
 
     // Set Width/Height & Resizable
-    let win_scale_factor = window.scale_factor();
     let (width, height) = attributes
       .inner_size
       .map(|size| size.to_logical::<f64>(win_scale_factor as f64).into())
       .unwrap_or((800, 600));
-    window.set_default_size(1, 1);
```
Zamoca42, line 109: The reason this was removed is that it functions the same as `box_.set_size_request(min_width, min_height);`

src/platform_impl/linux/window.rs
```diff
@@ -101,17 +107,20 @@ impl Window {
...
-    window.resize(width, height);
+
+    let window_clone = window.clone();
+    glib::idle_add_local_once(move || {
+      window_clone.set_resizable(true);
+      window_clone.resize(width, height);
+    });
```
Zamoca42: About https://github.com/tauri-apps/tao/issues/561, when resizable is set to false in the dev branch, a window with the size set to 220 x 90 is created as 332 x 200 without generating the inner size. 
However, if set_resizable(true) is added before resizing in the current branch, the window is created with the specified size.

## In dev branch
스크린샷 2024-10-02 오후 5 43 23

## Before adding `set_resizable(true)`

스크린샷 2024-10-02 오후 5 48 47

## After adding `set_resizable(true)`

스크린샷 2024-10-02 오후 6 07 53

src/platform_impl/linux/window.rs
```diff
@@ -101,17 +107,21 @@ impl Window {
...
     let (width, height) = attributes
       .inner_size
       .map(|size| size.to_logical::<f64>(win_scale_factor as f64).into())
       .unwrap_or((800, 600));
-    window.set_default_size(1, 1);
-    window.resize(width, height);
+
+    let window_clone = window.clone();
+    glib::idle_add_local_once(move || {
+      window_clone.set_resizable(true);
```
amrbashir: why setting resizable to true `here`?  and we are not resetting it again at the end of the callback

src/platform_impl/linux/window.rs
```diff
@@ -101,17 +107,21 @@ impl Window {
...
+      window_clone.set_default_size(min_width, min_height);
+      window_clone.resize(width, height);
```
amrbashir, line 122: resizing here makes the window create with a smaller size than expected, and then after a brief moment, it will jump into the requested size

amrbashir (changes requested):

src/platform_impl/linux/window.rs
Zamoca42: the initially created size is the correct window size. Changing the resizable property only adjusts the width so that the title is not truncated.

If the title is shortened, the size does not change. For example, if you change “Hit space to toggle resizability.” to “Hit space,” the size does not change after creation.

src/platform_impl/linux/window.rs
Zamoca42: initially thought it would help create the accurate inner size. However, it was not useful, and I realized that the title actually affects the inner size. I’ll revert that change and then go to the header to adjust the title.

src/platform_impl/linux/wayland/header.rs
```diff
@@ -1,25 +1,61 @@
...
+  fn connect_configure_event(window: &ApplicationWindow, title_label: &Label, char_width: f64) {
+    let title_label_clone = title_label.clone();
+    window.connect_configure_event(move |_, event| {
+      let (width, _) = event.size();
+      let max_chars = (width as f64 / char_width).floor() as i32;
+      title_label_clone.set_max_width_chars(if width < 220 { 0 } else { max_chars });
+      false
+    });
+  }
+
```
Zamoca42, line 59: > resizing here makes the window create with a smaller size than expected, and then after a brief moment, it will jump into the requested size

Previously, the header’s title affected the inner size, which made the size appear to jump.

https://github.com/user-attachments/assets/b19a9c15-3cec-4f2c-9047-7650dc4878af

I added an event to prevent the title from affecting the inner size.

https://github.com/user-attachments/assets/74b49956-3b0b-482c-affb-2065efae57fd

Additionally, the number of title characters displayed is updated when the window is resized based on its width.

https://github.com/user-attachments/assets/e5c52787-33ab-44e9-a3d3-c4411dbfe5c0

amrbashir: I can still see the size changes, see this gif
a

Zamoca42: Sorry, I was unable to reproduce the issue. 
The only potentially suspicious part I can think of is the priority when creating a new window, so I’ve tried setting all related priorities to a higher level. 
Could you check if this makes a difference?

yuezk:
Hi guys,

Thanks for your great work on this project. I encountered this problem and want to have a discussion about it.

By referring to the comments on https://github.com/tauri-apps/tauri/issues/10686#issuecomment-2325240664, the Hello World application seems to work on Wayland OOTB.

Why must we use the hack solution (by introducing a custom header on Wayland) to fix the moving problem? Is it possible that we are using the wrong API?

I'm totally new to the Tao system, above is just my humble opinion.

PR #1001, opened 2024-10-30
fix(linux): include border and titlebars in `Window::{outer_position, outer_size}` and `WindowEvent::Moved` event

### LINKED ISSUE #11510: [bug] Getting outer position of window on Ubuntu ignores caption and borders
### Describe the bug

When I try to get outer position of window the `outerPosition` should take into account window system caption and border but it is not. Because of this I get window position with offset x=1 px and y=34 px instead of 0-0.

Event in full screen mode y have offset identical to window caption.

### Reproduction

1. Place this code in main.ts

```
setInterval(async () => {
	const currentWindow = getCurrentWindow();
	const outerPosition = await currentWindow.outerPosition();

	console.log('outerPosition:', outerPosition);
}, 1000);
```

2. Run app

3. Move app to top-left corner of screen

4. Check log in console

### Expected behavior

The `console.log('outerPosition:', outerPosition);` should present window position x=0 and y=0.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Ubuntu 24.4.0 x86_64 (X64)
    ✔ webkit2gtk-4.1: 2.46.1
    ✔ rsvg2: 2.58.0
    ✔ rustc: 1.81.0 (eeb90cda1 2024-09-04)
    ✔ cargo: 1.81.0 (2dbb1af80 2024-08-20)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (default)
    - node: 22.5.1
    - npm: 10.8.2

[-] Packages
    - tauri 🦀: 2.0.6
    - tauri-build 🦀: 2.0.2
    - wry 🦀: 0.46.3
    - tao 🦀: 0.30.3
    - @tauri-apps/api : 2.0.3
    - @tauri-apps/cli : 2.0.5

[-] Plugins
    - tauri-plugin-shell 🦀: 2.0.2
    - @tauri-apps/plugin-shell : 2.0.1
    - tauri-plugin-fs 🦀: 2.0.3
    - @tauri-apps/plugin-fs : 2.0.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:
closes tauri-apps/tauri#11510

### DIFF
--- a/.changes/outer-position-linux.md
+++ b/.changes/outer-position-linux.md
@@ -2,4 +2,4 @@
 "tao": "patch"
 ---
 
-On Linux, `Window::outer_position` and `Window::outer_size` to include border and titlebar.
+On Linux, `Window::outer_position`, `Window::outer_size` and `WindowEvent::Moved` to include/account for borders and titlebar.
--- a/.changes/outer-position-linux.md
+++ b/.changes/outer-position-linux.md
@@ -2,4 +2,4 @@
 "tao": "patch"
 ---
 
-On Linux, fix position returned in `Window::outer_position` to include border and titlebar.
+On Linux, `Window::outer_position` and `Window::outer_size` to include border and titlebar.
--- a/.changes/outer-position-linux.md
+++ b/.changes/outer-position-linux.md
@@ -1,0 +1,5 @@
+---
+"tao": "patch"
+---
+
+On Linux, fix position returned in `Window::outer_position` to include border and titlebar.
--- a/src/platform_impl/linux/event_loop.rs
+++ b/src/platform_impl/linux/event_loop.rs
@@ -579,7 +579,10 @@ impl<T: 'static> EventLoop<T> {
             window.connect_configure_event(move |window, event| {
               let scale_factor = window.scale_factor();
 
-              let (x, y) = event.position();
+              let (x, y) = window
+                .window()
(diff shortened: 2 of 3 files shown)


pewsheen (approved):
Tested on my Ubuntu 24.04 with Wayland

btw, I  found the doc of `WindowEvent::Moved` and `outer_position` says Wayland will always return (0,0), but it seems that it returns the actual position on my end.

amrbashir:
> Tested on my Ubuntu 24.04 with Wayland
> 
> btw, I  found the doc of `WindowEvent::Moved` and `outer_position` says Wayland will always return (0,0), but it seems that it returns the actual position on my end.

Make sure you're running under Wayland, launching the example from within vscode usually runs the window under x11 even if the session is Wayland.

pewsheen:
> > Tested on my Ubuntu 24.04 with Wayland
> > btw, I  found the doc of `WindowEvent::Moved` and `outer_position` says Wayland will always return (0,0), but it seems that it returns the actual position on my end.
> 
> Make sure you're running under Wayland, launching the example from within vscode usually runs the window under x11 even if the session is Wayland.

Oh, you're right. I was running from the vscode and it runs on X11 actually

PR #1002, opened 2024-10-31
feat: add `set_badge_count` for Linux, iOS; `set_badge_label` for Macos; `set_overlay_icon` for Windows

ahqsoftwares:
This is a step further to implement https://github.com/tauri-apps/tauri/issues/4489 into the main tauri crate


ahqsoftwares:
Fixing the macos and linux code issues...

src/platform_impl/android/ndk_glue.rs
```diff
@@ -53,7 +53,7 @@ macro_rules! android_binding {
   };
   ($domain:ident, $package:ident, $activity:ident, $setup:path, $main:ident, $tao:path) => {{
     // NOTE: be careful when changing how this use statement is written
-    use $tao::{platform::android::prelude::android_fn, platform::android::prelude::*};
```
amrbashir: this should be reverted

src/platform_impl/macos/event_loop.rs
```diff
@@ -125,6 +126,16 @@ impl<T: 'static> EventLoopWindowTarget<T> {
     set_progress_indicator(progress);
   }
 
+  #[inline]
+  pub fn set_badge_count(&self, count: Option<i64>, _desktop_filename: Option<String>) {
+    set_badge_label(count.and_then(|count| Some(format!("{}", count))));
```
amrbashir:
```suggestion
    set_badge_label(count.map(|c| c.to_string()));
```

src/platform_impl/macos/window.rs
```diff
@@ -1498,6 +1499,14 @@ impl UnownedWindow {
   pub fn set_progress_bar(&self, progress: ProgressBarState) {
     set_progress_indicator(progress);
   }
+
+  pub fn set_badge_count(&self, count: Option<i64>, _desktop_filename: Option<String>) {
+    set_badge_label(count.and_then(|count| Some(format!("{}", count))));
```
amrbashir:
```suggestion
    set_badge_label(count.map(|c| c.to_string()));
```

src/platform_impl/windows/window.rs
```diff
@@ -1023,6 +1024,43 @@ impl Window {
     }
   }
 
+  #[inline]
+  pub fn set_overlay_icon(&self, icon: Option<String>) {
+    let Some(icon) = icon else {
+      unsafe {
+        let taskbar_list: ITaskbarList =
+          CoCreateInstance(&TaskbarList, None, CLSCTX_SERVER).unwrap();
```
amrbashir: seems like we call create `ITaskbarList` in all code paths, let's refactor this code to create it once and reuse it

src/window.rs
```diff
@@ -1115,6 +1115,43 @@ impl Window {
...
+  }
+
+  /// Sets the Overlay Icon **(Windows only)**
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / macOS**: See set_taskbar_badge
+  /// - **iOS / Android:** Unsupported.
+  #[inline]
+  pub fn set_overlay_icon(&self, _icon_path: Option<String>) {
```
amrbashir: I think this should use the `Icon` type instead of it being a path.
```suggestion
  pub fn set_overlay_icon(&self, icon: Option<Icon>) {
```

amrbashir: also since this Windows only API, it should be moved to `WindowExtWindows` trait

src/platform_impl/macos/event_loop.rs
```diff
@@ -125,6 +126,16 @@ impl<T: 'static> EventLoopWindowTarget<T> {
...
+  }
+
+  #[inline]
+  pub fn set_badge_label(&self, label: Option<String>) {
+    set_badge_label(label);
+  }
```
amrbashir, line 136: looks like this is not used

src/platform_impl/macos/window.rs
```diff
@@ -1498,6 +1499,14 @@ impl UnownedWindow {
...
+  }
+
+  pub fn set_badge_label(&self, label: Option<String>) {
+    set_badge_label(label);
+  }
```
amrbashir, line 1504: and this one is not used as well

src/window.rs
```diff
@@ -1115,6 +1115,43 @@ impl Window {
...
+  /// ## Implementation Details
+  /// - _desktop_filename is for linux ONLY!
+  /// 
+  /// ## Platform-specific
+  ///
+  /// - **Windows**: Windows allows to set an arbitrary .ico file. See set_overlay_icon
+  /// - **Linux / macOS**: Badge count is app-wide and not specific to this window. Only supported desktop environments with `libunity` (e.g. GNOME).
+  /// - **iOS / Android:** Unsupported.
+  #[inline]
+  pub fn set_badge_count(&self, _count: Option<i64>, _desktop_filename: Option<String>) {
```
amrbashir: I wonder if we can support this API on Windows by creating a simple icon programmatically that is red circle, with the number inside it.

amrbashir (changes requested):

src/window.rs
ahqsoftwares: maybe, but will that not increase the complexity of the crate, but I guess we should at the same time

src/platform_impl/windows/window.rs
ahqsoftwares: Yeah, we can use LazyLock but that'll bump to rust 1.80, or we can use lazy_static

amrbashir: no that's not what I meant, I meant something like this:
```rs
let taskbar_list =  CoCreateInstance();
let icon = if let Some(icon) = icon {
	// load icon
} else {
  HICON::default()
};
taskbar_list.SetOverlayIcon()
```

src/platform_impl/macos/event_loop.rs
ahqsoftwares: Yep, this has been reserved, I shall add it as a macos specific api

src/window.rs
amrbashir: if we don't want to support it here, then I suggest moving `set_badge_count` to `WindowExtUnix` and `WindowExtMacos` since it is only supported on these two platforms and require different parameters anyways.

src/platform_impl/windows/window.rs
ahqsoftwares: got it, so we can add it to the Window struct

src/window.rs
ahqsoftwares: ✅

src/platform_impl/windows/window.rs
amrbashir: not quite, this is what I mean
```rs
  #[inline]
  pub fn set_overlay_icon(&self, icon: Option<String>) {
    let taskbar_list: ITaskbarList =
      CoCreateInstance(&TaskbarList, None, CLSCTX_SERVER).unwrap();
    let icon = icon.map(|icon| {
      let image = unsafe {
        LoadImageW(
          None,
          PCWSTR(HSTRING::from(icon).as_ptr()),
          IMAGE_ICON,
          0,
          0,
          IMAGE_FLAGS(0x00000040 | 0x00000010),
        )
      }
      .unwrap();

      HICON(image.0)
    }.unwrap_or(HICON::default());

    unsafe {
      taskbar_list
        .SetOverlayIcon(self.window.0, icon, None)
        .unwrap();
    }
  }
```

ahqsoftwares: Ohk got it now

src/event_loop.rs
```diff
@@ -301,6 +301,19 @@ impl<T> EventLoopWindowTarget<T> {
     self.p.set_progress_bar(_progress)
   }
 
+  /// Sets the badge count
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows:** Unsupported. Use the Progress Bar Function Available in Window (Windows can have different taskbar badges for different Window)
+  /// - **Linux:** Only supported desktop environments with `libunity` (e.g. GNOME).
```
amrbashir: I think we can support it on iOS, see https://github.com/tauri-apps/tauri/issues/4489#issuecomment-2276804628

ahqsoftwares: That'd be a great addition

src/platform/macos.rs
```diff
@@ -84,6 +84,9 @@ pub trait WindowExtMacOS {
   ///
   /// <https://developer.apple.com/documentation/appkit/nswindow/1419167-titlebarappearstransparent>
   fn set_titlebar_transparent(&self, transparent: bool);
+
+  /// Sets the badge label on the taskbar
+  fn set_badge_label(&self, label: Option<String>);
```
amrbashir, line 89: This should be on `EventLoopWindowTargetExtMacOS`as well

src/platform/unix.rs
```diff
@@ -78,6 +78,8 @@ pub trait WindowExtUnix {
 
   /// Whether to show the window icon in the taskbar or not.
   fn set_skip_taskbar(&self, skip: bool) -> Result<(), ExternalError>;
+
+  fn set_badge_count(&self, count: Option<i64>, desktop_filename: Option<String>);
```
amrbashir, line 82: This should be on `EventLoopWindowTargetExtUnix`as well

src/platform_impl/windows/window.rs
```diff
@@ -1023,6 +1023,30 @@ impl Window {
...
+      return;
+    };
+
+    let icon = inner.as_raw_handle();
+
+    unsafe {
+      taskbar
+        .SetOverlayIcon(self.window.0, icon, None)
+        .unwrap();
+    }
```
```suggestion
    let taskbar: ITaskbarList = unsafe {
      CoCreateInstance(&TaskbarList, None, CLSCTX_SERVER).unwrap()
    };

    let icon = icon.map(|i| i.inner.as_raw_handle()).unwrap_or(HICON::default());

    unsafe {
      taskbar
        .SetOverlayIcon(self.window.0, icon, None)
        .unwrap_or(());
    }
```

amrbashir (changes requested):

src/platform/unix.rs
ahqsoftwares: Done

src/platform/macos.rs
ahqsoftwares: Done

ahqsoftwares:
What do we do about these warnings (linux)?

```
warning: use of deprecated associated function `gtk::glib::main_context_channel::<impl gtk::glib::MainContext>::channel`: Use an async channel, from async-channel for example, on the main context using spawn_future_local() instead
   --> src/platform_impl/linux/event_loop.rs:237:71
    |
237 |     let (window_requests_tx, window_requests_rx) = glib::MainContext::channel(Priority::default());     
    |                                                                       ^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated associated function `gtk::glib::main_context_channel::<impl gtk::glib::MainContext>::channel`: Use an async channel, from async-channel for example, on the main context using spawn_future_local() instead
   --> src/platform_impl/linux/event_loop.rs:251:55
    |
251 |       let (device_tx, device_rx) = glib::MainContext::channel(glib::Priority::default());
    |                                                       ^^^^^^^

warning: `tao` (lib) generated 2 warnings
```

src/event_loop.rs
ahqsoftwares: Implemented, please review

src/window.rs
```diff
@@ -1115,6 +1115,31 @@ impl Window {
...
+  /// ## Implementation Details
+  /// - _desktop_filename is for linux ONLY!
+  /// 
+  /// ## Platform-specific
+  ///
+  /// - **Windows**: Windows allows to set an arbitrary .ico file. See set_overlay_icon 
+  /// - **Linux / macOS**: Badge count is app-wide and not specific to this window. Only supported desktop environments with `libunity` (e.g. GNOME).
+  /// - **iOS / Android:** Unsupported.
+  #[inline]
+  pub fn set_badge_count(&self, _count: Option<i64>, _desktop_filename: Option<String>) {
```
amrbashir: this is no longer needed since we have platform-specific methods on WindowExtMacos, WindowExtWindows and WindowExtUnix

src/platform_impl/ios/event_loop.rs
```diff
@@ -94,6 +95,11 @@ impl<T: 'static> EventLoopWindowTarget<T> {
     debug!("`EventLoopWindowTarget::cursor_position` is ignored on iOS");
     Ok((0, 0).into())
   }
+
+  /// Sets badge count on iOS launcher. 0 hides the count
+  pub fn set_badge_count(&self, count: i32) {
```
amrbashir, line 99: this needs to be exposed on `EventLoopWindowTargetExtIOS` as well as `WindowExtIOS`

.changes/badge_count.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Add **set_badge_count** for Linux, iOS; **set_badge_label** for Macos; **set_overlay_icon** for Windows
```
amrbashir: I prefer to list each method like this
```suggestion
Add `WindowExtUnix::set_badge_count`, `WindowExtWindow::set_overlay_icon`
```
and so on for the rest of added methods

amrbashir (changes requested):

src/window.rs
ahqsoftwares: Yeah, 😅 I forgot to remove that

src/platform_impl/ios/event_loop.rs
ahqsoftwares: I shall do it

ahqsoftwares: There's no `EventLoopWindowTargetExtIOS`

.changes/badge_count.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Add **WindowExtUnix::set_badge_count** for Linux; **WindowExtIos::set_badge_count** for iOS; **WindowExtMacos::set_badge_label** for Macos; **MacdowExtWindows::set_overlay_icon** for Windows
```
amrbashir:
```suggestion
Add `WindowExtUnix::set_badge_count` for Linux, `WindowExtIos::set_badge_count` for iOS, `WindowExtMacos::set_badge_label` for Macos, `MacdowExtWindows::set_overlay_icon` for Windows
```

amrbashir (changes requested):
LGTM, just need to fix failing CI due to the example failing to compile

ahqsoftwares:
Yeah, fixing that example

ahqsoftwares:
The windows implementation with Icon doesn't work for some reason

ahqsoftwares:
Somehow it auto fixed idk how and why.

Found why: the window needs to have an icon set for some reason?

ahqsoftwares:
@amrbashir Why is the Icon struct needed to be cloned inside the set_overlay_icon function?

ahqsoftwares:
nvm fixed it, Basically the Icon has to be borrowed for some unknown reason

amrbashir:
> nvm fixed it, Basically the Icon has to be borrowed for some unknown reason

otherwise, it will be dropped and freeed

ahqsoftwares:
> > nvm fixed it, Basically the Icon has to be borrowed for some unknown reason
> 
> otherwise, it will be dropped and freeed

No issue, I made it to accept a borrowed Icon

ahqsoftwares:
image
that's strange

image

src/platform/ios.rs
```diff
@@ -98,6 +98,8 @@ pub trait WindowExtIOS {
   /// and then calls
   /// [`-[UIViewController setNeedsStatusBarAppearanceUpdate]`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621354-setneedsstatusbarappearanceupdat?language=objc).
   fn set_prefers_status_bar_hidden(&self, hidden: bool);
+
+  fn set_badge_count(&self, count: i32);
```
amrbashir, line 104: we need to addd `EvnetLoopWindowTargetExtIOS` and add a method similar to this on it

amrbashir:
there is also a missing import for macOS, not sure why iOS is failing though

src/platform/ios.rs
ahqsoftwares: We need to add a trait, right?

ahqsoftwares:
> there is also a missing import for macOS, not sure why iOS is failing though

I have no idea either

ahqsoftwares:
Both the examples under cfg of macos and ios are not applying. Do you have any idea why?

ahqsoftwares:
That commit should be it

amrbashir (approved):

amrbashir:
Thank you

ahqsoftwares:
Thanks a lot, I shall also PR to tauri crate to attach those apis to the tauri app

ahqsoftwares:
@amrbashir We'll implement the overlay label for Windows under a feature in the main tauri crate (it'll need the image crate, ab_glyph and imageproc crate as pointed out by @retrouser955 to me)

amrbashir:
we already depend on `image` crate but I don't know, seems like it would bloat the size of the app, let's just follow what electron did which is to provide two separate methods `setBadgeCount` for macOS and Linux and `setOverlayIcon` for Windows only.

ahqsoftwares:
Sure ! I shall work on a PR tomorrow

PR #1003, opened 2024-11-02
feat(macos): add ExitRequested event

luoffei:
Added `event::ExitRequested` handle for dock icon exit on macOS.

### DIFF
--- a/src/platform_impl/macos/app_delegate.rs
+++ b/src/platform_impl/macos/app_delegate.rs
@@ -19,8 +19,11 @@ use std::{
 
 use cocoa::foundation::{NSArray, NSURL};
 use std::ffi::CStr;
+use std::os::raw::c_int;
 
 static AUX_DELEGATE_STATE_NAME: &str = "auxState";
+
+const NS_TERMINATE_LATER: c_int = 2;
 
 pub struct AuxDelegateState {
   /// We store this value in order to be able to defer setting the activation policy until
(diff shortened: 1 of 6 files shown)


src/platform_impl/macos/app_delegate.rs
```diff
@@ -106,6 +113,13 @@ extern "C" fn did_finish_launching(this: &Object, _: Sel, _: id) {
   trace!("Completed `applicationDidFinishLaunching`");
 }
 
+extern "C" fn application_should_terminate(_: &Object, _: Sel, _: id) -> c_int {
+  trace!("Triggered `applicationShouldTerminate`");
+  AppState::should_exit();
+  trace!("Completed `applicationShouldTerminate`");
+  NS_TERMINATE_LATER
```
amrbashir, line 120: If I understand correctly, this will cause the application to never quit unless the event loop is changed programatically. 

I see that in the associated tauri PR, you call `replyToApplicationShouldTerminate` which is fine I guess but the default behavior here should return `NS_TERMINATE_NOW` and need a way for that to be overridden.

luoffei: I didn't think of a good way to handle this problem.

Is it possible to transplant the processing method in tauri to tao? By default, `NSTerminateLater` is returned. If the user does not handle the event, reply to exit. Otherwise, reply, terminate or cancel according to the user's processing.

amrbashir: there is a similar mechanism used when ScaleFactorChanged event is fired.

PR #1034, opened 2024-12-10
refactor: replace static with const for constants

Integral-Tech:
- According to [The Rust Reference](https://doc.rust-lang.org/reference/items/static-items.html), 
> Constants should, in general, be preferred over statics unless one of the following are true:
> - Large amounts of data are being stored
> - The single-address property of statics is required.
> - Interior mutability is required.


FabianLars:
I wonder if the second point is why those are static since they have to be valid across the FFI boundary. In winit at least the `CURSOR_BYTES` is still static, `CURSOR_ROOT` is a normal string var and the AUX one is tao specific.

Did you test this on a macOS device?

Integral-Tech:
> I wonder if the second point is why those are static since they have to be valid across the FFI boundary. In winit at least the `CURSOR_BYTES` is still static, `CURSOR_ROOT` is a normal string var and the AUX one is tao specific.
> 
> Did you test this on a macOS device?

I don't think it will be a problem since constants are essentially inlined wherever they are used.

FabianLars:
auxState was changed in another pr already. i reverted CURSOR_BYTES for now since i don't trust the const in the msg_send call.

FabianLars (approved):

PR #1039, opened 2025-01-04
fix(windows): fix set_size doing nothing for undecorated window with shadows

### LINKED ISSUE #12168: [bug] important | tauri v2.2.1 can't set window's size !!!
### Describe the bug

My tauri version is 2.2.1, which project from offical NPM template. When I set the window's property that **_decorations is false_**, the _**Window.setSize()**_ function can't work.

### Reproduction

1. Pasting the code to your src/main.ts:

``` typescript
import { PhysicalSize, Window } from "@tauri-apps/api/window";

window.addEventListener("DOMContentLoaded", () => {
  document.querySelector("#greet-form")?.addEventListener("submit", async (e) => {
    e.preventDefault();

    const win = Window.getCurrent() as Window;

    let { width, height } = await win.innerSize();

    win.setSize(new PhysicalSize(width - 5, height - 5));
  });
});
```

2. Testing
### decorations is true (correctly effect):
https://github.com/user-attachments/assets/154c20d4-f338-4c17-88d9-fab5e9438402

### decorations is false:
https://github.com/user-attachments/assets/98a633e2-39d9-46a0-a0a1-72a56e0d0b55

4. Conclusion
I can't modify window's size when window's property decorations is false.

### Expected behavior

Whatever decorations be modified any value, we should get the same effect.

### Full `tauri info` output

```text
Please see the Reproduction chapter's video.
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:
closes tauri-apps/tauri#12168

### DIFF
--- a/.changes/windows-set-size-undecorated-window.md
+++ b/.changes/windows-set-size-undecorated-window.md
@@ -1,0 +1,5 @@
+---
+"tao": "patch"
+---
+
+On Windows, fix regression that caused `Window::set_size` to have no effect at all for undecorated window with shadows.
--- a/src/platform_impl/windows/window.rs
+++ b/src/platform_impl/windows/window.rs
@@ -287,7 +287,7 @@ impl Window {
   #[inline]
   pub fn set_inner_size(&self, size: Size) {
     let scale_factor = self.scale_factor();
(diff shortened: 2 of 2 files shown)


FabianLars (approved):
tested decorated and undecorated both with and without shadows and all combinations worked.

PR #1048, opened 2025-01-21
fix(macos): remove internal `From` implementation

madsmtm:
Removes `impl From<ActivationPolicy> for NSApplicationActivationPolicy`.

This is needed to avoid publicly exposing a dependency on cocoa, which will allow [using `objc2`](https://github.com/tauri-apps/wry/issues/1239) in the future.

### DIFF
--- a/.changes/remove-from-activation-policy.md
+++ b/.changes/remove-from-activation-policy.md
@@ -1,0 +1,5 @@
+---
+tao: minor
+---
+
+macOS: Remove `From<ActivationPolicy>` implementation for `cocoa::appkit::NSApplicationActivationPolicy`.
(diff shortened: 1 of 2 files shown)


amrbashir (approved):
Thank you.

I should also note that we may never refactor this crate to use `objc2` as it is currently in maintenance mode only, and I still think we should go back to use winit.

madsmtm:
Whoops, didn't see your comment before I posted https://github.com/tauri-apps/tao/pull/1049, let's discuss it over there?

amrbashir:
I suspected you maybe working on it so I wanted to give you a heads up, thanks anyways, we really appreciate it

PR #1049, opened 2025-01-21
Use `objc2`

madsmtm:
I suspect it'll take a while before `winit` will be in a state where `tao` can use that directly, so in the meantime, let's slowly start transitioning `tao` to `objc2`. I tried to make this transition in a way that minimizes the amount of code changes, but it's really hard to separate things out, so it ended up being quite large anyways.

Full disclosure: The biggest reason why I'm doing this is to reduce the number of dependencies on `objc` across the ecosystem, I don't intend on fully finishing the transition in `tao` / making full use of the features offered by `objc2` (hopefully `winit` will be ready to be usable for you one day).
But I think even this initial transition can be valuable to you, as it fixes at least some of the memory leaks currently present, and makes it easier to develop new features (including [visionOS support](https://github.com/tauri-apps/tauri/issues/9969)).

I've tested every example on macOS 14.7.1, and on Mac Catalyst (works very poorly here, but at least we don't trigger any obvious encoding validation errors).

Builds upon https://github.com/tauri-apps/tao/pull/1048.
Fixes the last part of https://github.com/tauri-apps/wry/issues/1239.

CC @pewsheen since you did [the work in `wry`](https://github.com/tauri-apps/wry/pull/1316).

### DIFF
--- a/.changes/objc2.md
+++ b/.changes/objc2.md
@@ -1,0 +1,5 @@
+---
+tao: patch
+---
+
+Use `objc2`.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,10 +104,41 @@ ndk-context = "0.1"
 tao-macros = { version = "0.1.0", path = "./tao-macros" }
 
 [target."cfg(any(target_os = \"ios\", target_os = \"macos\"))".dependencies]
-objc = "0.2"
+objc2 = "0.5"
 
 [target."cfg(target_os = \"macos\")".dependencies]
-cocoa = "0.26"
+objc2-foundation = { version = "0.2", default-features = false, features = [
+  "std",
+  "NSArray",
+  "NSAttributedString",
+  "NSAutoreleasePool",
+  "NSDate",
+  "NSDictionary",
+  "NSEnumerator",
+  "NSGeometry",
+  "NSObjCRuntime",
+  "NSRange",
+  "NSString",
+  "NSThread",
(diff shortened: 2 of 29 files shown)


madsmtm:
Though I guess rejecting this PR would push me towards working more on getting Winit into a state where `tao` can use it as a dependency... ;)

amrbashir:
> Though I guess rejecting this PR would push me towards working more on getting Winit into a state where `tao` can use it as a dependency... ;)

No way, objc2 is awesome and I would accept this PR asap but I don't have macOS test, @pewsheen @FabianLars can any of you test and review this PR?

amrbashir:
ofc the ultimate goal is to go back to winit as a dependency. I was

amrbashir:
oops..

amrbashir:
ofc the ultimate goal is to go back to winit as a dependency. I was thinking about creating a crate that uses git submodule to generate a glue code that re-exports winit on macOS and Windows, and delegates onto gtk on Linux, although I think a `Platfrom` trait that could be implemented externally could be the better way to go.

pewsheen (commented):
I've tested the functionality on macOS and it looks good overall!

As for iOS, maybe @FabianLars can help test it?

FabianLars:
seems to work fine in my testing as well :)

thanks for the pr - will update to 0.6 in a follow up pr

FabianLars (approved):

PR #1050, opened 2025-01-27
chore: Update msrv to 1.74 and test it in CI

FabianLars:
We could push this lower if we wanted to but cargo 1.84 couldn't handle that automatically 🤷


FabianLars:
clippy is far from happy but i only did the save changes here

PR #1052, opened 2025-02-08
fix(windows): calculate accurate window insets for undecorated shadows

amrbashir:
ref: https://github.com/zed-industries/zed/blob/7bddb390cabefb177d9996dc580749d64e6ca3b6/crates/gpui/src/platform/windows/events.rs#L1418-L1454

closes #905

### DIFF
--- a/src/platform_impl/windows/util.rs
+++ b/src/platform_impl/windows/util.rs
@@ -17,6 +17,7 @@ use crate::{
   window::CursorIcon,
 };
 
+use once_cell::sync::Lazy;
 use windows::{
   core::{HRESULT, PCSTR, PCWSTR},
   Win32::{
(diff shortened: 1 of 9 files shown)


amrbashir:
@Legend-Master can you check if this fixes the Windows 10 borders? I mean this one https://github.com/tauri-apps/tauri/issues/11654

Legend-Master:
> @Legend-Master can you check if this fixes the Windows 10 borders?

Seems like the title bar is still there with the change? The resize handle is at the bottom of the title bar and you can't click that title bar (will click through)

image

Tested on Windows 10 using the decorations example with `decorations` set to `false` by default

```rust
  .with_title("Hit space to toggle decorations.")
  .with_inner_size(LogicalSize::new(400.0, 200.0))
  .with_decorations(false)
  .build(&event_loop)
  .unwrap();
```

src/platform_impl/windows/util.rs
```diff
@@ -425,3 +426,72 @@ pub fn get_instance_handle() -> windows::Win32::Foundation::HMODULE {
...
+  // case.)
+  //
+  // - On Windows 11
+  // The top inset is calculated using an empirical formula that I derived through various
+  // tests. Without this, the top 1-2 rows of pixels in our window would be obscured.
+
+  let frame_thickness = get_frame_thickness(dpi);
+
+  let top_inset = match WIN_VERSION.build {
+    v if v >= 2200 => (dpi as f32 / USER_DEFAULT_SCREEN_DPI as f32).round() as i32,
```
Legend-Master:
```suggestion
    v if v >= 22000 => (dpi as f32 / USER_DEFAULT_SCREEN_DPI as f32).round() as i32,
```

Legend-Master: With this change, the window looks like this, the top border is missing and also not resizable from that edge

image

I have read about an article https://kubyshkin.name/posts/win32-window-custom-title-bar-caption/ talking about this before, and the idea is that you need to draw your own top border to fake it...

amrbashir: > With this change, the window looks like this, the top border is missing and also not resizable from that edge
> 
> I have read about an article https://kubyshkin.name/posts/win32-window-custom-title-bar-caption/ talking about this before, and the idea is that you need to draw your own top border to fake it...

I did read that one before but I don't want to add any custom drawing inside taol.

amrbashir: I think this PR is good enough for aesthetics, don't you think?

As for, functionality maybe we bring back custom hittesting for only top border only on Windows 10?

Legend-Master: > I did read that one before but I don't want to add any custom drawing inside taol.

Yeah, same thoughts here

> I think this PR is good enough for aesthetics, don't you think?

Yeah, it's good enough to me, at least it's much better than what we had before ;)

Also I have tested WindowsAppSdk sometime ago, and their `ExtendsContentIntoTitleBar` also had this problem (and they draw control buttons themselves as you tell can from the icons)

image

> As for, functionality maybe we bring back custom hittesting for only top border only on Windows 10?

I think we'll need this

amrbashir: @Legend-Master does top-left and top-right resizing edge work and only top doesn't or all 3 don't work?

amrbashir: actually looks like the issues exists in Windows 11 too.

amrbashir: should be good to go now

Legend-Master (approved):
Works great, tested maximized and normal window on Windows 10, the position and size seems to be right as well, awesome work!

PR #1055, opened 2025-02-09
fix(wayland): decoration not clickable when initial visible is false

yuezk:
## Problem

The decoration is not clickable if the window initial visible is `false`.

## Reproduce

1. Run the following code
2. The close button, the maximum button, and the minimum button is not clickable. 

```rs
use tao::{
  event::{Event, WindowEvent},
  window::WindowBuilder,
};

fn main() {
  env_logger::init();

    .with_visible(false)
    .build(&event_loop)
    .unwrap();

  window.set_visible(true);

    match event {
      Event::WindowEvent {
        ..
      } => *control_flow = ControlFlow::Exit,
      _ => (),
    }
  });
}

```

## Solution

Avoid setting the above child property of the `EventBox`. The default value is `false`. Seems there is no need to set it to `true`.

https://docs.gtk.org/gtk3/method.EventBox.set_above_child.html

### DIFF
--- a/.changes/fix-wayland-decoration-click.md
+++ b/.changes/fix-wayland-decoration-click.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+On Linux Wayland, fixed a bug where the window decoration is not clickable.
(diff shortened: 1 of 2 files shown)


yuriyhanysh:
I can reproduce this bug on Fedora (KDE Plasma 6.5.5, Wayland) in `cjpais/Handy` app.
Can confirm the fix resolves the issue.

How can we merge it?

FabianLars:
fixed in https://github.com/tauri-apps/tao/pull/1218

PR #1056, opened 2025-02-10
fix(windows): use correct position/size on WM_DPICHANGED for Windows 11

### LINKED ISSUE #1053: Window Resizing Bug When Moving Between Monitors with Different Scaling
**Describe the bug**  
When moving the window between monitors with different scaling settings causes the window’s size to become erratic. The window appears to resize incorrectly, making it difficult to move or position properly. Notably, dragging the window by its right half seems to offer slightly more stability, but this is not a reliable workaround.

**Steps To Reproduce**  
1. Run Tao’s `window` example.  
2. Ensure you have at least two monitors with different scaling settings (for example, one monitor at 100% scaling and another at 150% scaling).  
3. Drag the window from one monitor to the other.  
4. Observe how the window’s size becomes inconsistent or “bugged” upon moving between the monitors.

**Expected behavior**  
The window should either maintain its size or adjust appropriately when transitioning between monitors with different scaling settings. There should be no erratic resizing behavior during the move.

**Screenshots**  
A video demonstrating the issue is attached:  

https://github.com/user-attachments/assets/f297789f-0791-4751-90b7-1a4bf2d95023

- **OS:** Windows  
- **Rustc:** 1.84.1

**Additional context**  
I have observed this bug across various Tauri-based apps and believe it is rooted in Tao’s window management implementation. The issue persists in both Tauri 1.0 and 2.0 environments. I attempted to investigate the underlying cause, but the mechanism seems more complex than initially expected. Any insights or potential workarounds would be greatly appreciated. Additionally, I found a couple of related issues, but neither appears to be actively tracked:  
- https://github.com/tauri-apps/tauri/issues/10263
- https://github.com/tauri-apps/tao/issues/816

I couldn't find an existing issue in Tao's repository that directly addresses this bug, so I created this one. However, if this is a duplicate, please feel free to close it.

dgerhardt:
Adjustments to the OS-suggested position are now only applied on Windows 10. They are not needed anymore on Windows 11.

Related issue for winit: https://github.com/rust-windowing/winit/issues/4041

Fixes #1053, tauri-apps/tauri#10263, tauri-apps/tauri#12626.

### DIFF
--- a/.changes/fix-win11-dpichanged-position.md
+++ b/.changes/fix-win11-dpichanged-position.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+On Windows 11, fix incorrect window positioning and sizing on `WM_DPICHANGED`.
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -1998,99 +1998,107 @@ unsafe fn public_window_callback_inner<T: 'static>(
       }
 
       let new_outer_rect: RECT;
-      {
-        let suggested_ul = (
-          suggested_rect.left + margin_left,
-          suggested_rect.top + margin_top,
-        );
+      if util::WIN_VERSION.build < 22000 {
+        // The window position needs adjustment on Windows 10.
+        {
+          let suggested_ul = (
+            suggested_rect.left + margin_left,
+            suggested_rect.top + margin_top,
+          );
 
-        let mut conservative_rect = RECT {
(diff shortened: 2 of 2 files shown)


dgerhardt:
The same issue exists in winit and a PR with a similar fix was opened there a few days ago: https://github.com/rust-windowing/winit/pull/4119

acarl005:
Thank you for this!! 🙏

FabianLars (approved):
merge first, ask questions later - thanks for your contribution

Issue #1023, opened 2024-12-08
WebviewWindow doesn't adhere to x,y on Mac


someone2080:
**Describe the bug**
Bug is very similar to [11430](https://github.com/tauri-apps/tauri/issues/11430)

[Reproduction](https://github.com/someone2080/broken_webview_window-tauri_2) 

> When I try to create a new window using WebviewWindow with x,y options, the newly created Window appears on the Display, based on the x,y coordinates, but then jumps to the primary Display. I have to use setPosition method to move it to the right Display.

This behavior can be fixed with `setPosition`, ex: `await webview.setPosition(position)`, but the workaround won't work if the screens have different scale/ratio, see image:

<img width="474" alt="Screenshot 2024-12-08 at 4 27 20 PM" src="https://github.com/user-attachments/assets/b4501998-a6cb-4c6b-8247-2e74d789f6c0">

<img width="474" alt="Screenshot 2024-12-08 at 4 28 16 PM" src="https://github.com/user-attachments/assets/935e61b5-7f05-4851-8465-9f6deb861b2d">

someone2080:
Hey guys, are there any updates on this one? This is a regression from Tauri 1. I wish I could fix it myself, but I'm still learning Rust...

someone2080:
For anyone else having the same issue: It looks like on macOS, `WebviewWindow.setPosition` accepts `LogicalPosition` instead of PhysicalPosition. If, for example, you get your coords from `Monitor`, which provides `PhysicalPosition` on both Mac and Windows, you will need to convert this  `PhysicalPosition`  to `LogicalPosition`.

Similar issue here #1057

PR #1057, opened 2025-02-10
fix(macos): treat Monitor::from_point args as physical

### LINKED ISSUE #12676: [bug] Inconsistent behavior of Monitor::from_point across platforms
### Describe the bug

`Monitor::from_point` treats `x` and `y` values as physical on Windows while on macOS the values are treated as logical. Currently, the method cannot be reliably be used across platforms without checking for the platform and converting the values conditionally. The documentation does not specify if the values passed should be physical or logical, so it is unclear which of both behaviors is the correct one.

### Reproduction

```typescript
const monitor = await primaryMonitor();
if (!monitor) {
  console.error('No monitor found.');
  return;
}
if (monitor.scaleFactor <= 1) {
  console.error('Scaling is not enabled for primary monitor.');
  return;
}
console.info('Monitor 1: ' + JSON.stringify(monitor));
const x = monitor.size.width - 1;
const y = monitor.size.height - 1;
console.info(`Monitor 2 <- monitorFromPoint(${x}, ${y})`);
const monitor2 = await monitorFromPoint(x, y);
console.info('Monitor 2: ' + JSON.stringify(monitor2));
if (monitor.name !== monitor2?.name) {
  console.error('Expected point to be on primary monitor.');
} else {
  console.info('The point is on the primary monitor as expected.');
}
```

### Expected behavior

`Monitor::from_point` should treat passed `x` and `y` values the same way across platforms.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.3.0 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.84.0 (9fc6b4312 2025-01-07)
    ✔ cargo: 1.84.0 (66221abde 2024-11-19)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 23.6.1
    - yarn: 4.6.0
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.2.5
    - tauri-build 🦀: 2.0.5
    - wry 🦀: 0.48.1
    - tao 🦀: 0.31.1
    - @tauri-apps/api : 2.2.0
    - @tauri-apps/cli : 2.2.7

[-] Plugins
    - tauri-plugin-cli 🦀: 2.2.0
    - @tauri-apps/plugin-cli : 2.2.0
    - tauri-plugin-single-instance 🦀: 2.2.1
    - tauri-plugin-store 🦀: 2.2.0
    - @tauri-apps/plugin-store : 2.2.0
    - tauri-plugin-process 🦀: 2.2.0
    - @tauri-apps/plugin-process : 2.2.0
    - tauri-plugin-log 🦀: 2.2.1
    - @tauri-apps/plugin-log : 2.2.1
    - tauri-plugin-shell 🦀: 2.2.0
    - @tauri-apps/plugin-shell : 2.2.0
    - tauri-plugin-updater 🦀: 2.5.0
    - @tauri-apps/plugin-updater : 2.5.0
    - tauri-plugin-deep-link 🦀: 2.2.0
    - @tauri-apps/plugin-deep-link : 2.2.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist/app/browser
    - framework: Angular
    - bundler: Webpack
```

### Stack trace

```text

```

### Additional context

_No response_

dgerhardt:
The `x` and `y` values passed to `from_point` are now treated as physical and therefore converted to logical values on macOS. This makes the behavior consistent with the Windows implementation.

Closes tauri-apps/tauri#12676.

### DIFF
--- a/.changes/macos-monitor-from-point-physical-coordinates.md
+++ b/.changes/macos-monitor-from-point-physical-coordinates.md
@@ -1,0 +1,6 @@
+---
+"tao": patch
+---
+
+The `x` and `y` values passed to `Monitor::from_point` are now treated as physical and therefore converted to logical values on macOS.
+This makes the behavior consistent with the implementation for Windows.
--- a/src/platform_impl/macos/monitor.rs
+++ b/src/platform_impl/macos/monitor.rs
@@ -18,6 +18,7 @@ use core_graphics::{
   display::{CGDirectDisplayID, CGDisplay, CGDisplayBounds},
   geometry::CGPoint,
 };
+use dpi::LogicalPosition;
 use objc2::{msg_send, rc::Retained};
 use objc2_app_kit::NSScreen;
 use objc2_foundation::{MainThreadMarker, NSString, NSUInteger};
@@ -164,7 +165,9 @@ pub fn from_point(x: f64, y: f64) -> Option<MonitorHandle> {
   unsafe {
     for monitor in available_monitors() {
       let bound = CGDisplayBounds(monitor.0);
-      if CGRectContainsPoint(bound, CGPoint::new(x, y)) > 0 {
+      let position =
+        LogicalPosition::from_physical::<_, f64>((x as f64, y as f64), monitor.scale_factor());
+      if CGRectContainsPoint(bound, CGPoint::new(position.x, position.y)) > 0 {
         return Some(monitor);
       }
     }


pewsheen (commented):
Since macOS uses logical coordinates, there might be a chance that developers can only pass logical position to `from_point` (like the coordinate was retrieved from an OS event). If we only accept physical position then the function may become useless in the case.

I'm thinking that if we change the interface from `x:f64, y:f64` to `Position`, then we can take care of both scenarios. But this will be a breaking change and Tauri may require another PR for adapting this.

cc @FabianLars

dgerhardt:
This seems to be a tricky situation. Adjusting the behavior might be a breaking change for some users but at the same time I think that `from_point` can't be used for cross-platform development at all in its current state. At least not without applying your own workarounds which check for the OS. Also, most of TAO's API uses `PhysicalPosition` as return type, so you still would need to explicitly convert the type in many cases on macOS.

Using Position for the parameter sounds like a good solution which would also be more consistent with the rest of the API. Maybe we could introduce a new `from_position` method and deprecate `from_point` for a smoother transition?

FabianLars:
> Maybe we could introduce a new from_position method and deprecate from_point for a smoother transition?

with tao being 0.x this imo doesn't really matter unless we simply prefer one name over the other.

lucasfernog:
let's just document that monitor_from_point takes logical values on macOS and physical on other platforms, and add monitor_from_position

amrbashir:
we can just remove it altogether and release a minor for `monitor_from_position`, only `tauri` will need deprecation.

someone2080:
It looks like this is not the only function that works with logical values instead of physical ones on Mac. Look at #1023. 

In this case, I think updating the docs and deprecating the existing methods that are inconsistent between platforms in favor of creating new ones in the next versions would be a more preferable approach for many users.

PR #1058, opened 2025-02-14
feat: add `set_dock_visibility` method

### LINKED ISSUE #105: macOS: Dock management
Ability to hide app from Dock

There's a window-level `.skip_taskbar()` method, but as far as I can tell nothing for hiding the app itself from the Dock

1111mp:
Add `set_dock_visibility` method to support setting the visibility of the application in the dock.
Close #105

As mentioned [here](https://github.com/tauri-apps/tao/issues/105#issuecomment-2412796114), refer to `Electron` to implement `set_dock_visibility` method.

This is my first PR, so there are some issues that need further confirmation: 
- Does the implementation of `set_dock_visibility` need to be completely consistent with `set_activation_policy`. For example, we don’t need to do anything extra here: https://github.com/1111mp/tao/blob/c4d0f0555fb0f65993fa4f7b944c45465ed14211/src/platform_impl/macos/app_delegate.rs#L32
  ```rust
  pub struct AuxDelegateState {
      /// We store this value in order to be able to defer setting the activation policy until
      /// after the app has finished launching. If the activation policy is set earlier, the
      /// menubar is initially unresponsive on macOS 10.15 for example.
      pub activation_policy: ActivationPolicy,

      /// Whether the application is visible in the dock.
      pub dock_visibility: bool,

      pub activate_ignoring_other_apps: bool,
  }
  ```
- Does the implementation of [set_dock_show](https://github.com/1111mp/tao/blob/c4d0f0555fb0f65993fa4f7b944c45465ed14211/src/platform_impl/macos/dock.rs#L40) need to be exactly the same as [Electron](https://github.com/electron/electron/blob/9ecb848c677fa1bdd4ad423a83e3827c5c03b716/shell/browser/browser_mac.mm#L518). At present, I am testing locally that the display and hiding of `docker` are consistent with the behavior of `Electron`. More extensive testing may be needed...

This is my local demo:
Screen Recording 2025-02-14 at 21 29 41

Should the icon change only occur in the local environment? I pointed tauri and tao directly to the local code by `patch.crates-io` and then started the demo project.
```yml
[patch.crates-io]
tauri = { path = "../../tauri/crates/tauri" }
tao = { path = "../../tao" }
```
Demo code:
```rust
// Learn more about Tauri commands at https://tauri.app/develop/calling-rust/
#[tauri::command]
fn greet(name: &str) -> String {
    format!("Hello, {}! You've been greeted from Rust!", name)
}

#[tauri::command]
fn set_dock_visibility(app_handle: tauri::AppHandle, visible: bool) {
    let _ = app_handle.set_dock_visibility(visible);
    // let _ = app_handle.set_dock_visibility(true);
    // let _ = app_handle.set_dock_visibility(false);
    // let _ = app_handle.set_dock_visibility(true);
    // let _ = app_handle.set_dock_visibility(false);
}

#[cfg_attr(mobile, tauri::mobile_entry_point)]
pub fn run() {
    let builder = tauri::Builder::default()
        .plugin(tauri_plugin_opener::init())
        .setup(|app| {
            // app.set_activation_policy(tauri::ActivationPolicy::Regular);
            // app.set_dock_visibility(false);
            Ok(())
        })
        .invoke_handler(tauri::generate_handler![greet, set_dock_visibility]);

    let app = builder.build(tauri::generate_context!()).unwrap();

    app.run(|app, err| match err {
        tauri::RunEvent::ExitRequested { api, .. } => {
            api.prevent_exit();
            let _ = app.set_dock_visibility(false);
        }
        _ => {}
    });
}

```

> This is my first time to submit a PR. If I have any mistakes, please feel free to correct me. Thank you.


.changes/feat-dock-visibility-on-macos.md
```diff
@@ -0,0 +1,5 @@
+---
+'tao': minor
```
lucasfernog:
```suggestion
'tao': patch
```

lucasfernog:
> Should the icon change only occur in the local environment? I pointed tauri and tao directly to the local code by patch.crates-io and then started the demo project.

yeah that's a dev-only thing we'll need to fix on the tauri side later

lucasfernog (approved):
pretty good - let's wait and see if we need to port any of the electron patches

PR #1083, opened 2025-03-01
fix(macos): handle None return from standardWindowButton (fix #1081)

### LINKED ISSUE #1081: Crash on macOS when creating window with disabled buttons due to unwrap on None value
**Describe the bug**
The application crashes during startup when using tao 0.32.5 on macOS. The crash occurs because the code in `window.rs` ([line 288](https://github.com/tauri-apps/tao/blob/7f16fc14cea2d5936cc3627404221775a61e8812/src/platform_impl/macos/window.rs#L288)) unconditionally calls `unwrap()` on an Option value returned by `standardWindowButton`, which becomes `None` when certain window buttons are disabled in the configuration.

**Steps To Reproduce**
1. Create a macOS application using tao 0.32.5
2. Configure a window with some standard buttons disabled, such as:
```json
{
  "fullscreen": false,
  "resizable": false,
  "minimizable": false,
  "maximizable": false,
  "closable": false
}
```
3. Attempt to run the application
4. The application crashes during initialization with an "unwrap on None" panic

**Expected behavior**
The application should properly handle cases where window buttons are disabled in the configuration, checking if the `standardWindowButton` returns a value before attempting to unwrap it.

OS: macOS 15.3.1 (24D70)
Rustc: rustc 1.85.0 (4d91de4e4 2025-02-17)

**Additional context**
The issue occurs specifically in the macOS platform implementation. When window buttons are disabled through configuration, the `standardWindowButton` method returns `None`, but the code doesn't check for this possibility before calling `unwrap()`.

According to Apple's documentation (https://developer.apple.com/documentation/appkit/nswindow/standardwindowbutton(_:)), the `standardWindowButton(_:)` method can return nil:

This is precisely what's happening when buttons are disabled in the window configuration - the method correctly returns nil (None in Rust terms), but the tao implementation tries to unwrap this value unconditionally.

A proper fix would involve using `if let Some(button) = window.standardWindowButton(...)` or similar pattern matching to safely handle cases where buttons are disabled or not available, rather than directly unwrapping the returned value.

gezihuzi:
## Description
This PR fixes an application crash on macOS when certain window buttons are disabled in the configuration. The crash occurs because the code in `window.rs` (line 288) unconditionally calls `unwrap()` on an Option value returned by `standardWindowButton`, which becomes `None` when window buttons are disabled.

## Related Issue
Closes #1081

## Changes Made
- Modified the code to safely handle cases where `standardWindowButton` returns `None`
- Added proper pattern matching instead of unconditionally unwrapping the returned value
- Ensured the application doesn't panic when window buttons are disabled

## Testing
- Created a macOS application with various window button configurations (disabled minimizable, maximizable, etc.)
- Verified the application now starts successfully without crashing

## Additional Notes
According to Apple's documentation, `standardWindowButton(_:)` can return nil when buttons are not available, which is what happens when buttons are disabled in the configuration. This PR updates the implementation to properly handle this case.

### DIFF
--- a/.changes/fix-macos-window-buttons.md
+++ b/.changes/fix-macos-window-buttons.md
@@ -1,5 +1,6 @@
 ---
 "tao": patch
+tag: "bug"
 ---
 
 Fixed application crash during startup when certain window buttons are disabled in the configuration on macOS. The crash occurred because the code unconditionally called `unwrap()` on an Option value returned by `standardWindowButton`, which becomes `None` when window buttons are disabled.
--- a/.changes/fix-macos-window-buttons.md
+++ b/.changes/fix-macos-window-buttons.md
@@ -1,6 +1,5 @@
 ---
 "tao": patch
-tag: "bug"
 ---
 
 Fixed application crash during startup when certain window buttons are disabled in the configuration on macOS. The crash occurred because the code unconditionally called `unwrap()` on an Option value returned by `standardWindowButton`, which becomes `None` when window buttons are disabled.
--- a/.changes/fix-macos-window-buttons.md
+++ b/.changes/fix-macos-window-buttons.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fixed application crash during startup when certain window buttons are disabled in the configuration on macOS. The crash occurred because the code unconditionally called `unwrap()` on an Option value returned by `standardWindowButton`, which becomes `None` when window buttons are disabled.
--- a/.changes/fix-macos-window-buttons.md
+++ b/.changes/fix-macos-window-buttons.md
@@ -2,4 +2,4 @@
 "tao": patch
 ---
 
-Fixed application crash during startup when certain window buttons are disabled in the configuration on macOS. The crash occurred because the code unconditionally called `unwrap()` on an Option value returned by `standardWindowButton`, which becomes `None` when window buttons are disabled.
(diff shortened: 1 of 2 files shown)


amrbashir (commented):
Thank you! could you also add a change file in `.changes` directory?

gezihuzi:
> Thank you! could you also add a change file in `.changes` directory?

Hi @amrbashir, I've just added the change file in the .changes directory. Please let me know if anything else needs adjustment. Thanks!

.changes/fix-macos-window-buttons.md
```diff
@@ -0,0 +1,6 @@
+---
+"tao": patch
+tag: "bug"
```
amrbashir:
```suggestion
"tao": patch
```

tags are really only used in the main tauri repo and the syntax is `patch:bug` for future reference

.changes/fix-macos-window-buttons.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fixed application crash during startup when certain window buttons are disabled in the configuration on macOS. The crash occurred because the code unconditionally called `unwrap()` on an Option value returned by `standardWindowButton`, which becomes `None` when window buttons are disabled.
```
amrbashir:
```suggestion
Fixed application crash during startup when certain window buttons are disabled on macOS.
```

amrbashir (approved):
Thank you

PR #1085, opened 2025-03-01
ci: Update lockfile on versioning PRs

FabianLars:


.github/workflows/covector-version-or-publish.yml
```diff
@@ -33,13 +33,17 @@ jobs:
...
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          command: 'version-or-publish'
+          command: "version-or-publish"
           createRelease: true
           recognizeContributors: true
 
+      - name: Sync Cargo.lock
+        if: steps.covector.outputs.commandRan == 'version'
+        run: cargo tree --depth 0
```
amrbashir, line 42: This probably needs to be in covector config instead

FabianLars: i don't touch covector stuff 😅 

copied that from the tauri workflows btw, no idea who put that there

Issue #1086, opened 2025-03-02
macos crash on command+h


starpit:
**Describe the bug**
When using cmd+h keyboard shortcut to hide my Tauri v2 app, i get a crash. The crash seems to point to Tao. See the stack trace below.

**Steps To Reproduce**
```shell
sh <(curl https://create.tauri.app/sh)
cd tauri-app/
npm install
npm run tauri dev
(hit command+h)
```

**Expected behavior**
No crash.

**Screenshots**
n/a

**Platform and Versions (please complete the following information):**
OS: MacOS 15.3.1
Rustc: 1.85.0

**Additional context**
```
thread 'main' panicked at /Users/nickm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tao-0.32.4/src/platform_impl/macos/app.rs:43:19:
messsaging sendEvent: to nil
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: objc2::runtime::message_receiver::panic_null
             at /Users/nickm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/objc2-0.6.0/src/runtime/message_receiver.rs:342:5
   3: objc2::runtime::message_receiver::msg_send_check
             at /Users/nickm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/objc2-0.6.0/src/runtime/message_receiver.rs:305:9
   4: objc2::runtime::message_receiver::MessageReceiver::send_message
             at /Users/nickm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/objc2-0.6.0/src/runtime/message_receiver.rs:428:13
   5: <MethodFamily as objc2::__macro_helpers::msg_send_retained::MsgSend<Receiver,Return>>::send_message
             at /Users/nickm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/objc2-0.6.0/src/__macro_helpers/msg_send_retained.rs:35:28
   6: tao::platform_impl::platform::app::send_event
             at /Users/nickm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tao-0.32.4/src/platform_impl/macos/app.rs:43:19
[...]

PR #1090, opened 2025-03-06
fix(macos): crash sending key event when app has no windows

lucasfernog:
I noticed this when clicking the `Close Window` menu item which sends a `Cmd+W` event to the application. It looks like the window was closed and tao tried to send the event to `keyWindow` which was null. This check also exists on winit:
https://github.com/rust-windowing/winit/blob/be1baf164cc80ebc8586f2c528fe854b1dafcb34/src/platform_impl/apple/appkit/app.rs#L36

### DIFF
--- a/.changes/fix-macos-send-event-crash.md
+++ b/.changes/fix-macos-send-event-crash.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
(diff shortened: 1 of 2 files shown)


amrbashir (approved):

dwosk:
https://github.com/tauri-apps/tao/issues/1086

PR #1107, opened 2025-04-15
fix(windows): initial position sets to 0

Legend-Master:
Reference: #1018

The initial position is set to account for the shadows in tauri, and we need to not reset the position to default when calculating its monitor

Maybe we should rethink about how the APIs work when dealing with this in the future, currently (0, 0) doesn't put the window to the exact top left corner

### DIFF
--- a/.changes/nagative-initial-position.md
+++ b/.changes/nagative-initial-position.md
@@ -1,0 +1,5 @@
+---
+tao: patch
+---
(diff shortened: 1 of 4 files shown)


amrbashir (approved):

amrbashir:
I think accounting for it here is fine atm

PR #1114, opened 2025-05-16
feat: add openharmony platform support

richerfu:
This is an experiential support for OpenHarmony. Lots of code base with `winit`(https://github.com/rust-windowing/winit/pull/4117) which already has some exmaples to test it.

Now i just make sure it can be compilerd but still maybe some crash so i need to use some examples to test it.


richerfu:
Now i have made some examples to test it. See details: https://github.com/richerfu/wry-example. It seems worked and doesn't crash, i think it can be merged. Could you help me to merge this work for OpenHamrony, thanks.

PR #1119, opened 2025-06-04
fix(windows): apply scroll lines per wheel setting

Legend-Master:
Fix `WindowEvent::MouseWheel` doesn't account for mouse wheel speed settings

Reference docs: https://learn.microsoft.com/en-us/windows/win32/inputdev/about-mouse-input#determining-the-number-of-scroll-lines
Reference winit PR: https://github.com/rust-windowing/winit/pull/4268

### DIFF
--- a/.changes/windows-scroll-line-settings.md
+++ b/.changes/windows-scroll-line-settings.md
@@ -2,4 +2,4 @@
 tao: patch
 ---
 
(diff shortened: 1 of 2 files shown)


lucasfernog (approved):

Legend-Master:
Merging as the winit PR just got accepted

PR #1120, opened 2025-06-04
feat: add focusable property and Window::set_focusable

robertrpf:
Adds the option to create non focusable windows and change that state with Window::set_focusable.

The feature was requested in https://github.com/tauri-apps/tauri/issues/11130

### DIFF
--- a/.changes/focusable.md
+++ b/.changes/focusable.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Added `WindowBuilder::with_focusable` to allow creating unfocusable windows.
--- a/.changes/set-focusable.md
+++ b/.changes/set-focusable.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Added `Window::set_focusable`.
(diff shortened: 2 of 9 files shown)


lucasfernog:
we should try to implement this for the other platforms too

FabianLars:
yeah, sooner or later at least. the author asked upfront if windows-only implementation is okay for a start. I said yes, cause it's a start at least.
Not sure if one of us can work on linux / macos anytime soon so maybe merge it windows-only for now and add the other platforms later?

gtk seems to be easy* so i can add that right to this pr but i can't spot anything for macos

\* if accept-focus is right, weird that electron doesn't support linux

edit: for macos, ~~electron sets `setDisableKeyOrMainWindow` but i don't find _any_ mention of this anywhere else on the internet.~~ nvm that's a custom setting they use as a return value for the overwritten canBecomeMain|KeyWindow methods.

lucasfernog:
CI error on Windows has been happening for a while, not related to this PR

lucasfernog (approved):

PR #1123, opened 2025-06-15
android: add intent handling

ambiso:
Demo implementation of https://github.com/tauri-apps/wry/issues/1563 

When the activity is first launched or its onNewIntent method is called, we check if there's a SEND intent, extract its data, and pass it on to the watch in [intent-context](https://github.com/ambiso/intent-context/blob/main/src/lib.rs)[^1].

The intent representation could also be improved; for example the intent jobject could be kept around and data could be retrieved only when needed.

[^1]: The dependency on tokio is maybe not ideal; this was just the most convenient way to make it work in my dioxus project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ bitflags = "2"
 crossbeam-channel = "0.5"
 url = "2"
 dpi = "0.1"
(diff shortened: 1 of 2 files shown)


FabianLars:
thanks for the pr, this was implemented as part of another feature :)

PR #1125, opened 2025-06-29
fix(windows): Cloak window initially to prevent white flash

FabianLars:
found this in an old tao copy from ages ago. i can't remember if i found out where `49235` comes from though 🤔


FabianLars:
can't repro the flash in tao anymore

PR #1128, opened 2025-08-11
feat: add openharmony support

richerfu:
This is an experiential support for OpenHarmony. Lots of code base with `winit`(https://github.com/rust-windowing/winit/pull/4117) which already has some exmaples to test it.

Now i have made some examples to test it. See details: https://github.com/richerfu/wry-example. It seems worked and doesn't crash, i think it can be merged. Could you help me to merge this work for OpenHamrony, thanks.


FabianLars:
Seeing the screenshots in the wry-example repo, can you maybe give me a quick rundown on how to run those examples in an emulator?

I've been searching on the internet for hours now and couldn't get it to work. The limited resources i found that looked like they may work were for real devices.

p.s. (i think i said that somewhere before but i can't find it right now) thanks for the PRs btw, no promises if and when they get merged though.

Edit: To be clear, i'm already failing to get an emulator running at all. latest deveco doesn't seem to have one and the few community emulators i've found don't work for one reason or another either.

richerfu:
So sorry that and i wrote a document to explain it: https://github.com/richerfu/wry-example/blob/main/Development.md

If you have any other questions you can ask me directly. Thanks @FabianLars

FabianLars:
Thank you very much! That is an awesome document :)

That said, it doesn't solve the main problem i have. On both Windows and macOS my DevEco installation doesn't come with a Device Manager nor do i see any way to install it or device images.
Could it be because i'm in Europe (with a european huawei account)?

richerfu:
I see, could you check the following file content?
Window: `C:\Users\xxx\AppData\Roaming\Huawei\${version}\options\country.region.xml`
Mac: `~/Library/Application Support/Huawei/${version}/options/country.region.xml`

Content should be:
```xml
<application>
    <component name="CountryRegionSetting">
        <countryregion name="CN"/>
    </component>
</application>
```

If not please copy those config to this file. @FabianLars

FabianLars:
Hah, that worked, awesome! Thank you so much! I will get to the testing then :)

src/platform_impl/ohos/mod.rs
```diff
@@ -0,0 +1,889 @@
...
+      }
+    }
+  }
+
+  pub fn run<F>(self, event_handler: F) -> ()
+  where
+    F: FnMut(event::Event<T>, &event_loop::EventLoopWindowTarget<T>, &mut ControlFlow),
+  {
+    let event_looper = Box::leak(Box::new(self));
+    event_looper.run_return(event_handler);
```
FabianLars, line 281: Would it cause any _major_ issues if we'd call `process::exit()` here similar to how we do it on Android? https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/android/mod.rs#L103-L104

I kinda hate that but i think it makes more sense to match the behavior on other platforms here. I only did a small test of this and it seemed to not cause issues for _me_.

richerfu: For openharmony, we can't block the eventloop and`run_return` will return immediately when it execute finish. So if we call `exit`, application will exit immediately.

> The OHOS of eventloop mode is similer with Node.js

Normally, i think we can see `run/run_app` as a hook, we can register callback that can dispatch event to tao eventloop. @FabianLars

FabianLars: ahh yeah, i take it back. I tested it incorrectly and didn't apply the local changes...

FabianLars:
would you be okay with me merging this and the wry pr into an `ohos` branch instead of `dev` ? I think having at least a base merged in our repo would make it a bit nicer for us to work on it as well (adding CI for example).
I have to focus on a poc in tauri for work first before i can take a real look at the PRs (that order would make sense either way so we can adapt the changes if needed for tauri) so alternatively we could also just keep the PRs open for the foreseeable future which is fine for me as well :)

p.s. official support in wry & tauri would probably wait until the winit integration progresses anyway.

src/platform_impl/ohos/mod.rs
FabianLars: i have a bad feeling about the tauri integration part with this behavior 🙃 it really does not fit in well with the other platforms.

richerfu:
No problem, i can provide any helps if you need.

richerfu:
Rebased dev branch. cc @frankplus @lucasfernog @FabianLars

Legend-Master:
@richerfu Following @FabianLars, I have opened 2 `feat/open-harmony` branches in tao and wry, do you mind I re-target the PRs to those branches and merge them there?

richerfu:
> @richerfu Following @FabianLars, I have opened 2 `feat/open-harmony` branches in tao and wry, do you mind I re-target the PRs to those branches and merge them there?

I'll do it later.

Legend-Master (approved):
Thanks! Awesome work!

PR #1132, opened 2025-08-20
chore: don't publish audit PDF to crates.io

Legend-Master:
Noticed this while doing an update

> Downloaded 17 crates (4.0MiB) in 5.92s (largest was `tao` at 2.3MiB)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ readme = "README.md"
 repository = "https://github.com/tauri-apps/tao"
 documentation = "https://docs.rs/tao"
 categories = [ "gui" ]
-exclude = ["audits/**"]
+include = ["/README.md", "src/**/*.rs", "examples/**/*.rs", "LICENSE*"]
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ readme = "README.md"
 repository = "https://github.com/tauri-apps/tao"
 documentation = "https://docs.rs/tao"
 categories = [ "gui" ]
+exclude = ["audits/**"]
```
FabianLars: we'll still end up with a lot of unnecessary stuff (though small in size), what about
```suggestion
include = ["/README.md", "src/**/*.rs", "examples/**/*.rs", "LICENSE*"]
```
(cargo.toml/.lock are included by cargo anyway)

Legend-Master: Does exclude examples impact `cargo doc` on crates.io? If not then we could go with this

FabianLars: ohh good question 🤔

FabianLars: i added it to the suggestion just in case

FabianLars (approved):

PR #1138, opened 2025-08-27
fix(macos): Fix color mappings in window.rs

### LINKED ISSUE #1127: Window background color is not properly set
**Describe the bug**
The background color of the window is not properly set as the window::RGBA (which is [u8;4]) is not properly converted to platform specific [0.0..=1.0; 4].

**Steps To Reproduce**
Set `with_background_color(RGBA::from[0x0f, 0x11, 0x16, 0xff])` on window builder. Optionally set the titlebar transparency to see the window color through it.

**Expected behavior**
Expected to have a nice dark background. Yet it becomes white.

**Screenshots**
Expected:
Image
Reality:
Image

OS: MacOS 15.5
Rustc: 1.88

**Additional context**
The issue is in the implementation of #995, that does not divide u8 values by 255.0. Does so for the alpha channel.

Image

Just adding the division manually into the tao source (`src/platform_impl/macos/window.rs`) and clean rebuild fixes the issue for me. I also see same issue in the other platforms src.

Why the alpha channel gets correctly normalized, but others do not?

aspcartman:
The color conversion from [u8;4] to [f64;4] was missing the division by 255.0. 

Fixes #1127.

### DIFF
--- a/src/platform_impl/android/ndk_glue.rs
+++ b/src/platform_impl/android/ndk_glue.rs
@@ -47,6 +47,7 @@ pub static PACKAGE: OnceCell<&str> = OnceCell::new();
 ///       - `private external fun focus(focus: Boolean)`
 /// 4. a one time setup function that will be ran once after tao has created its event loop in the `create` function above.
 /// 5. the main entry point of your android application.
+#[rustfmt::skip]
 #[macro_export]
 macro_rules! android_binding {
   ($domain:ident, $package:ident, $activity:ident, $setup:path, $main:ident) => {
--- a/src/platform_impl/android/ndk_glue.rs
+++ b/src/platform_impl/android/ndk_glue.rs
@@ -54,7 +54,7 @@ macro_rules! android_binding {
   };
   ($domain:ident, $package:ident, $activity:ident, $setup:path, $main:ident, $tao:path) => {{
     // NOTE: be careful when changing how this use statement is written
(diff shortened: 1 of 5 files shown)


src/platform_impl/android/ndk_glue.rs
```diff
@@ -54,7 +54,7 @@ macro_rules! android_binding {
   };
   ($domain:ident, $package:ident, $activity:ident, $setup:path, $main:ident, $tao:path) => {{
     // NOTE: be careful when changing how this use statement is written
-    use $tao::{platform::android::prelude::android_fn, platform::android::prelude::*};
+    use $tao::platform::android::prelude::{android_fn, *};
```
amrbashir: @FabianLars please revert this change, last time I changed it, it broke usage in `wry` iirc

FabianLars: stupid muscle memory, i didn't want to do `git add .` 🙃 

is there an attribute like ts-ignore we can use to tell cargo not to format that?

amrbashir: looks like you found `#[rustfmt::skip]`,

maybe it is worth to add more context in the comment above stating what exactly breaks

FabianLars: maybe yeah, but i have no idea what's going on 😂

FabianLars (approved):

PR #1139, opened 2025-08-27
feat(macos): add events for push notification registration

saurL:
feat(ios,macos): add event to send notification registration token

Implemented a new event on iOS and macOS to send the notification registration token.

Injected callbacks into AppDelegate to retrieve the token.

Added the notification registration call directly in the AppDelegate on macOS.


FabianLars:
The mix of verified and unverified (from different authors nonetheless) may get problematic 🤔

FabianLars:
Also can you try merging tauri-apps/tao@dev into your fork? That should clean up the diff. When i try to do it for you i get a 403 even though github says we're allowed to edit it.

saurL:
I think it should be done

saurL:
the non verified commit seems indeed to be problematic , not sure how to resolve it or why it is this way ( it comes from the tao repo)

FabianLars:
your commits aren't verified either but i really don't wanna drag this out any longer (in the tao repo at least) so i'll abuse my permissions for this.

saurL:
Sorry for that , I hope i haven't done any commit on the tauri repo yet , I'll configure so my commit are now verified

saurL:
Hello @FabianLars what's up with this PR ?

FabianLars:
ah sorry i forgot to push my local changes and didn't have access to the device the last few days.

FabianLars:
sorry for the commit spam, can't push to the branch directly idk

Issue #643, opened 2022-12-11
Use the color-scheme XDG portal on Linux for checking if the theme should be light/dark


X-m7:
Currently as I understand it TAO checks the current GTK theme to see if it contains "dark" in the name, however now there is a way to ask the desktop environment using XDG portals and D-Bus. The protocol was added [here](https://github.com/flatpak/xdg-desktop-portal/pull/633), and GNOME has provided documentation on how to use it [here](https://gitlab.gnome.org/GNOME/Initiatives/-/wikis/Dark-Style-Preference) (specifically see the "Other" section, which has example code written using C, SDL and GDBus/GIO, while the other sections assume libadwaita or libhandy is used).

The advantage of this method is that starting in GNOME 42 and elementaryOS 6 there is a global light/dark theme selector in their settings apps, which does not change the selected GTK theme, so only apps that check the portal setting will be able to follow that global light/dark preference. KDE Plasma also has such a light/dark setting, although ultimately the reported value from the portal is calculated using the selected colour scheme.

~~Also, using the portal to check the light/dark preference also means that apps will be able to respond immediately to the user switching between light and dark modes, while with the current method changing the GTK theme requires a restart of the apps.~~

declantsien:
> Also, using the portal to check the light/dark preference also means that apps will be able to respond immediately to the user switching between light and dark modes, while with the current method changing the GTK theme requires a restart of the apps.

This is not true... 
I believe Firefox/Chromium also uses only GTK themes.

If we decided to use XDG portal color-scheme, does that mean users have to install `xdg-desktop-portal-*` on their system. That is a overhead, IMO.

X-m7:
> I believe Firefox/Chromium also uses only GTK themes.

Firefox responds to my changing the colour scheme on KDE without changing the GTK theme (this effect is visible with the Alpenglow theme for example), so no it doesn't rely only on the GTK theme.

> If we decided to use XDG portal color-scheme, does that mean users have to install xdg-desktop-portal-* on their system.

That is already a necessity for anyone using Flatpak and Snap applications (so they can do things like open files without having to give them access to the entire system), as well as anyone wanting to do things like screen recording on Wayland (at least without apps needing to use private compositor specific APIs).

I think the ideal solution would be to prefer using the portal if it is present and a preference has been set (in case either the portal does not exist or the portal backend does not have support for that preference), and otherwise use the current method with the GTK theme name.

declantsien:
[...]

PR #1141, opened 2025-08-29
feat(linux): Add support for system color scheme using xdg-desktop-portal

haecker-felix:
Use the Linux XDG Desktop Portal to add support for the system color scheme. This works independently of the Linux distribution, desktop environment or toolkit used. If the system colour scheme changes to "dark mode", the window will adjust automatically.

For reference: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html

[Screencast From 2025-08-29 09-30-17.webm](https://github.com/user-attachments/assets/39eac1b8-2132-4bca-80fb-43264d750753)

Closes: 
- https://github.com/tauri-apps/tao/issues/643
- https://github.com/tauri-apps/tauri/issues/9427
- https://github.com/DioxusLabs/dioxus/issues/3700

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -412,7 +412,7 @@ pub enum WindowEvent<'a> {
   ///
   /// ## Platform-specific
   ///
-  /// - **Linux / Android / iOS:** Unsupported
+  /// - **Android / iOS:** Unsupported
   ThemeChanged(Theme),
 
   /// The window decorations has been clicked.
--- a/.changes/linux-system-color-scheme.md
+++ b/.changes/linux-system-color-scheme.md
@@ -2,4 +2,4 @@
 tao: minor
 ---
 
-Use the Linux XDG Desktop Portal to add support for the system color scheme. This works independently of the Linux distribution, desktop environment or toolkit used. If the system colour scheme changes to "dark mode", the window will adjust automatically.
+Use the Linux XDG Desktop Portal to add support for the system color scheme. Needs `dbus` feature flag.
--- a/.changes/linux-system-color-scheme.md
+++ b/.changes/linux-system-color-scheme.md
@@ -1,0 +1,5 @@
+---
+tao: minor
+---
+
+Use the Linux XDG Desktop Portal to add support for the system color scheme. This works independently of the Linux distribution, desktop environment or toolkit used. If the system colour scheme changes to "dark mode", the window will adjust automatically.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,6 +109,177 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "175571dd1d178ced59193a6fc02dde1b972eb0bc56c892cde9beeceac5bf0f6b"
 
 [[package]]
+name = "ashpd"
+version = "0.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da0986d5b4f0802160191ad75f8d33ada000558757db3defb70299ca95d9fcbd"
+dependencies = [
+ "async-fs",
+ "async-net",
+ "enumflags2",
+ "futures-channel",
+ "futures-util",
+ "rand 0.9.2",
+ "serde",
+ "serde_repr",
+ "url",
+ "zbus",
+]
+
+[[package]]
+name = "async-broadcast"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "435a87a52755b8f27fcf321ac4f04b2802e337c8c4872923137471ec39c37532"
+dependencies = [
+ "event-listener",
+ "event-listener-strategy",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-channel"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "924ed96dd52d1b75e9c1a3e6275715fd320f5f9439fb5a4a11fa51f4221158d2"
+dependencies = [
+ "concurrent-queue",
+ "event-listener-strategy",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-executor"
+version = "1.13.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "497c00e0fd83a72a79a39fcbd8e3e2f055d6f6c7e025f3b3d91f4f8e76527fb8"
+dependencies = [
+ "async-task",
+ "concurrent-queue",
+ "fastrand",
+ "futures-lite",
+ "pin-project-lite",
+ "slab",
+]
+
+[[package]]
+name = "async-fs"
+version = "2.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09f7e37c0ed80b2a977691c47dae8625cfb21e205827106c64f7c588766b2e50"
+dependencies = [
+ "async-lock",
+ "blocking",
+ "futures-lite",
+]
+
+[[package]]
+name = "async-io"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "19634d6336019ef220f09fd31168ce5c184b295cbf80345437cc36094ef223ca"
+dependencies = [
+ "async-lock",
+ "cfg-if",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite",
+ "parking",
+ "polling",
+ "rustix 1.0.8",
+ "slab",
+ "windows-sys 0.60.2",
+]
+
+[[package]]
+name = "async-lock"
+version = "3.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5fd03604047cee9b6ce9de9f70c6cd540a0520c813cbd49bae61f33ab80ed1dc"
+dependencies = [
+ "event-listener",
+ "event-listener-strategy",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-net"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b948000fad4873c1c9339d60f2623323a0cfd3816e5181033c6a5cb68b2accf7"
+dependencies = [
+ "async-io",
+ "blocking",
+ "futures-lite",
+]
+
+[[package]]
+name = "async-process"
+version = "2.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "65daa13722ad51e6ab1a1b9c01299142bc75135b337923cfa10e79bbbd669f00"
+dependencies = [
+ "async-channel",
+ "async-io",
+ "async-lock",
+ "async-signal",
+ "async-task",
+ "blocking",
+ "cfg-if",
+ "event-listener",
+ "futures-lite",
+ "rustix 1.0.8",
+]
+
+[[package]]
+name = "async-recursion"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.96",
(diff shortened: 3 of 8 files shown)


haecker-felix:
> error: package `zvariant v5.7.0` cannot be built because it requires rustc 1.77 or newer, while the currently active rustc version is 1.74.1

I downgraded `ashpd` from `0.12` to `0.7` with the hope that this will make it possible to compile with rustc 1.74. Alternatively, would it be an option to upgrade rustc to something newer?

FabianLars:
1.77 is fine since that's tauri's msrv

FabianLars:
the real problem though is the dependence on tokio or smol as we try or tried to keep tao and wry runtime agnostic and using dbus/zbus would also make it harder to get back to winit https://github.com/rust-windowing/winit/issues/1549 🤔 

really not sure how to proceed so i'll wait for feedback from the others.

does ashpd require the `tokio` or `async-std` feature flag to be enabled or could it work without either as well? (if it requires one i think tokio makes more sense since we use that downstream)

haecker-felix:
> 1.77 is fine since that's tauri's msrv

Ah, good to know! But let's wait for others feedback first before making any further changes.

> does ashpd require the tokio or async-std feature flag to be enabled or could it work without either as well? (if it requires one i think tokio makes more sense since we use that downstream)

I could switch to tokio, but that would likely require [async-compat](https://crates.io/crates/async-compat) then, as it's not possible to spawn tokio futures directly from glib. But yes, ashpd either requires the `async-std` or `tokio` feature to be enabled.

FabianLars:
hmm that sucks 😅

Hofer-Julian:
@FabianLars if the dependency on any async runtime is truly a blocker, we could also directly call the portal with dbus. That would suck though, since we'd basically re-implement parts of ashpd then.

You mentioned that you'll wait for feedback from others, who specifically do you mean?

FabianLars:
@amrbashir is the closest we have to a tao maintainer though idk how active he is at the moment. If he's unavail or has no opinion we could also ask @lucasfernog and @Legend-Master . If they don't have an opinion either i'll have to follow whatever winit decides/decided on 🤷

Legend-Master:
`ashpd` uses `zbus` underneath the hood and `zbus` seems to be an `async` based crate, so unless we use something like `dbus-rs` and implement the theme settings ourselves (which doesn't seem to be that hard from the `ashpd` implementation, not sure with `dbus-rs` though), we'll need an async runtime

If we are going with the async runtime route, I think it'll have to be in feature flags, defaults could be discussed though

Hofer-Julian:
What would be your preference @Legend-Master?

Reimplement with dbus-rs or using ashpd?

Legend-Master:
> What would be your preference @Legend-Master?
> 
> Reimplement with dbus-rs or using ashpd?

I don't know how much work would it take to re-implement it in `dbus-rs`, if that's manageable then I would prefer that as that makes the API simpler (no need to think about the async runtime when using `tao`)

haecker-felix:
Reimplemented using dbus-rs instead of ashpd/zbus :+1:

FabianLars:
sorry i was away over the weekend. we actually deliberately got rid of dbus-rs in the past eg https://github.com/tauri-apps/tauri/pull/6332 so idk how to feel about this. I think i'll pull myself out of this PR (i'm away for 2 weeks anyway).

haecker-felix:
I will wait with further changes until it is clear what is desired and what is not :/

To summarize, the current available options are:
1. `ashpd`/`zbus` + `smol` runtime (allows reusing of already existing glib main thread / context)
2. `ashpd`/`zbus` + `tokio` runtime
3. `dbus-rs` (no async runtime, but needs dedicated thread with message passing)

Legend-Master:
> we actually deliberately got rid of dbus-rs in the past

Do you remember the reason for that? (also it seems like tauri currently doesn't depend on dbus nor zbus)

> dbus-rs (no async runtime, but needs dedicated thread with message passing)

Personally I would prefer this, but take my opinion with a pinch of salt, I'm not the most familiar with the Linux ecosystem

Also I feel like we shouldn't be as strict as winit when considering these things, that's part of the reason why we branched out right?

And, I'm travelling from 11th to ~25th this month so I probably can't reply between this time period

FabianLars:
> Do you remember the reason for that? 

Linked in the PR. basically linking to a system library was a pain 🤷 

> (also it seems like tauri currently doesn't depend on dbus nor zbus)

correct, a few plugins use zbus (single-instance and notification come to mind)

> Also I feel like we shouldn't be as strict as winit when considering these things, that's part of the reason why we branched out right?

to a degree yes, but we're also looking to get back to winit. as long as we still use webkitgtk we still need our own linux backend though (much work on their side going on to make that possible next release(s))

> And, I'm travelling from 11th to ~25th this month so I probably can't reply between this time period

Safe travels and have fun :)

haecker-felix:
Hey folks, I hope you had a great vacation / time off! :)

I wanted to follow up to see if there is any new info on how we are proceeding with this PR so that Linux will get dark mode support as well. I summarized the possible options in https://github.com/tauri-apps/tao/pull/1141#issuecomment-3266723853 - let me know what you think.

FabianLars:
I believe re-introducing dbus-rs should be avoided if possible. Knowing the zbus community we probably would get a PR "fixing" that later on :D

Also whatever we'll use we should put the dependency behind a feature flag.

Speaking of, how about we mirror rfd's approach and offer both zbus-tokio and zbus-smol/async-std ? I'd really love to avoid the latter in tauri downstream (even considering the downsides of the tokio version you mentioned) but it may help other tao consumers not using tauri.

> [about winit] as long as we still use webkitgtk we still need our own linux backend though 

btw, without this i would have turned down this PR in favor of something downstream (easy to do but obv would keep ThemeChange noop on linux)

src/event.rs
```diff
@@ -412,7 +412,7 @@ pub enum WindowEvent<'a> {
   ///
   /// ## Platform-specific
   ///
-  /// - **Linux / Android / iOS:** Unsupported
+  /// - **Android / iOS:** Unsupported
```
FabianLars, line 415: judging by the win/mac impls you'll need to emit this event manually which i don't think you do yet.

Legend-Master:
> Speaking of, how about we mirror rfd's approach and offer both zbus-tokio and zbus-smol/async-std ?

If we can't use `dbus-rs` here, this would be the preferred way for me as well

haecker-felix:
> I believe re-introducing dbus-rs should be avoided if possible. Knowing the zbus community we probably would get a PR "fixing" that later on :D

Alright, `zbus` then, no `dbus-rs` :)

> Also whatever we'll use we should put the dependency behind a feature flag.

Just to make sure, if I understood you correctly, I would add a `async-std` and `tokio` feature flag, and enable the `async-std` one by default (to mirror `rfd` here)? Consumers would need to pick either `async-std` or `zbus` then.

FabianLars:
yep. though i don't care which flag will be the default, we could even have neither in the defaults 🤷

Hofer-Julian:
> yep. though i don't care which flag will be the default, we could even have neither in the defaults 🤷

If possible, it would be great, of course if the system color scheme would *just work* without developers manually opting in to it 😀

haecker-felix:
It's now possible to choose between `async-std` and `tokio`, defaulting to `async-std`. 

If you don't choose anything, or have `default-features = false`, you get a compile-time error: 

```
error: Either the `async-std` or the `tokio` feature has to be enabled
  --> /var/home/haeckerfelix/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ashpd-0.12.0/src/lib.rs:12:1
   |
12 | compile_error!("Either the `async-std` or the `tokio` feature has to be enabled");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This means that if consumers of `tao` set `default-features = false`, they must explicitly enable either the `async-std` or `tokio` feature.

src/event.rs
haecker-felix: Good catch, fixed it :+1:

FabianLars:
Thanks for implementing the feature flags!

> This means that if consumers of tao set default-features = false, they must explicitly enable either the async-std or tokio feature.

This is a bit of a no-go since we have to keep a tokio & asyncstd free option for a few users. Should be an easy fix though by marking the 3 deps as optional and having both feature flags enable it. plus a few `#[cfg(any(feature = "tokio", feature = "async-std"))]` flags in the code.

> If possible, it would be great, of course if the system color scheme would just work without developers manually opting in to it 😀

I get the sentiment but the lockfile change is a bit intense so we shouldn't take that decision lightly.

haecker-felix:
> This is a bit of a no-go since we have to keep a tokio & asyncstd free option for a few users.

Done 👍 

> I get the sentiment but the lockfile change is a bit intense so we shouldn't take that decision lightly.

I understand your concerns, but I fear that if this is not enabled by default on Tauri's side, hardly any developers will actively opt in - which would be a shame, because Tauri is one of the few remaining frameworks that does not yet support the new system “color-scheme” standard on Linux environments.

FabianLars:
it will be enabled on the tauri side (without opt-out probably) which is also why i wanted a tokio option cause that would make the most sense there. my concerns were only about tao (and wry) for those hardcore users (or those building frameworks on top of it like dioxus)

Hofer-Julian:
I am happy to hear that, seems like we misunderstood your intentions :)

Hofer-Julian:
Hey @FabianLars!

@haecker-felix made both tokio and async-std opt-in in https://github.com/tauri-apps/tao/pull/1141/commits/a7104e767070440ea9e7ef19a0ce31e75577a701

Is there anything else to be done before we can get this merged? :)

Andrew15-5:
One nitpick I have is that `Cargo.toml` now lost its EOL newline at EOF, which is not great for diffs. And it's not even consistent, as the new `src/platform_impl/linux/portal.rs` file has it. Though I haven't looked at all other files.

FabianLars:
sorry for the silence here. end of last year and beginning of the new year was an insanely busy time for me. Another reason was that i wanted to see how https://github.com/PolyMeilex/rfd/pull/286 plays out (they removed zbus again).
I'm still pretty busy this and probably next week but will get back to you then. We're not talking about months this time 😅 sorry again.

FabianLars:
I absolutely hate myself for it but i changed my mind, i'm so sorry for the time you wasted. Can we go back to dbus-rs? Perhaps in a follow-up PR we can do something like rfd and get rid of that as well (but really, in a follow-up! let's get it working first).

Since tao is specifically for tauri which depends on libdbus because of webkitgtk anyway it should be fine to use it here as well. Perhaps behind a default feature though just in case someone needs to disable it.

Hofer-Julian:
@FabianLars I switched back to dbus-rs. I really hope that this is now the last time we have to do this switcheroo

FabianLars:
no promises hehe

src/event.rs
FabianLars, line 415:
```suggestion
  /// - **Linux**: Needs `dbus` feature (enabled by default).
  /// - **Android / iOS:** Unsupported
```

Hofer-Julian:
@FabianLars can you please approve the workflow?

Hofer-Julian:
@FabianLars seems like it's difficult to get this to work without bumping the MSRV. Would you be fine with bumping it from `1.74` to `1.85`?

FabianLars:
sadly not, 1.77 is our current max. I tested felix' old commit https://github.com/tauri-apps/tao/commit/7c04b0687167aa39a2eb05e3a9e48882e9e9ff46 a few days ago and that seemed to work fine. Should we roll back to that? (missing the `fix: Emit ThemeChanged events` commit tho)  Seems like he also copied that to a backup branch https://github.com/haecker-felix/tao/tree/color-scheme-dbus-rs

i think i have a few minutes free some time this weekend so i could do that if you're okay with that approach - would also solve the problem with the unsigned commits 😅

Hofer-Julian:
> i think i have a few minutes free some time this weekend so i could do that if you're okay with that approach - would also solve the problem with the unsigned commits 😅

That would be much appreciated. Let me know if you need help with verifying that it actually works :)

FabianLars (approved):

Hofer-Julian:
Thank you @FabianLars!

PR #1146, opened 2025-10-11
fix(ios): Use WillEnterForeground instead of DidBecomeActive for Event::Resumed in iOS

JeffTsang:
This is a useful event to prepare your app before it enters the foreground on iOS.

### DIFF
--- a/.changes/will-enter-foreground.md
+++ b/.changes/will-enter-foreground.md
@@ -1,0 +1,5 @@
+---
+"tao": minor
+---
+
+Add WillEnterForeground event for iOS.
--- a/.changes/will-enter-foreground.md
+++ b/.changes/will-enter-foreground.md
@@ -1,5 +1,5 @@
 ---
-"tao": minor
+"tao": patch
 ---
 
 Add WillEnterForeground event for iOS.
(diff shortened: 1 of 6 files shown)


.changes/will-enter-foreground.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": minor
```
lucasfernog:
```suggestion
"tao": patch
```

lucasfernog:
I think we should use Event::Resumed for this.. it would also match the Android behavior (though that one needs a bug fix to not call the method on the first run)

JeffTsang:
Both events may have their own use cases, but if applicationWillEnterForeground is more closely aligned with Event::Resumed, it can be changed.

JeffTsang:
@lucasfernog I've updated the code to use Event::Resumed.

lucasfernog:
btw I think `Event::NewEvents(StartCause::Poll)` is a good replacement for Event::Resumed (that's what we use in tauri currently).
This change can be seen as a breaking change so we'll wait to release it on 0.35

FabianLars (approved):

PR #1148, opened 2025-10-15
refactor(android): call WindowEvent::Destroyed on activity destroy

### LINKED ISSUE #13568: [bug] Android crashes when re-launching after closing from app switcher
### Describe the bug

Steps to reproduce:
- Build an Android app with tauri-plugin-log
- Swipe away the app from the app switcher
- Open the app
- Observe a panic & crash

In the log you will see: `error while running tauri application: PluginInitialization("log", "attempted to set a logger after the logging system was already initialized")`

### Reproduction

_No response_

### Expected behavior

No crash should occur

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.3.2 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.87.0 (17067e9ac 2025-05-09)
    ✔ cargo: 1.87.0 (99624be96 2025-05-06)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 22.13.0
    - pnpm: 10.11.0
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.5.1
    - tauri-build 🦀: 2.2.0
    - wry 🦀: 0.51.2
    - tao 🦀: 0.33.0
    - @tauri-apps/api : 2.5.0
    - @tauri-apps/cli : 2.5.0

[-] Plugins
    - tauri-plugin-single-instance 🦀: 2.2.4
    - tauri-plugin-log 🦀: 2.4.0
    - @tauri-apps/plugin-log : not installed!
    - tauri-plugin-store 🦀: 2.2.0
    - @tauri-apps/plugin-store : 2.2.0
    - tauri-plugin-os 🦀: 2.2.1
    - @tauri-apps/plugin-os : 2.2.1
    - tauri-plugin-process 🦀: 2.2.1
    - @tauri-apps/plugin-process : 2.2.1
    - tauri-plugin-updater 🦀: 2.7.1
    - @tauri-apps/plugin-updater : 2.7.1
    - tauri-plugin-opener 🦀: 2.2.7
    - @tauri-apps/plugin-opener : 2.2.7
    - tauri-plugin-deep-link 🦀: 2.3.0
    - @tauri-apps/plugin-deep-link : 2.3.0
    - tauri-plugin-fs 🦀: 2.3.0
    - tauri-plugin-shell 🦀: 2.2.1
    - @tauri-apps/plugin-shell : 2.2.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:3001/
    - framework: React
    - bundler: Vite

[-] iOS
    - Developer Teams: [REDACTED]
```

### Stack trace

```text

```

### Additional context

_No response_

lucasfernog:
currently the Rust lifecycle is partially tied to the Android activity lifecycle:
- when the activity is created we initialize the event loop: https://github.com/tauri-apps/wry/blob/84b0ec2e35236e0b43150e37b55c7b6d2e5cf7e9/src/android/kotlin/WryActivity.kt#L61 https://github.com/tauri-apps/tao/blob/fda48e80ef0600079666c19146eb01bdfc4dd3c4/src/platform_impl/android/ndk_glue.rs#L196
- on destroy we are not doing anything in particular now, but we should let the app exit the event loop so it can properly restart when the activity gets recreated

to reproduce this, open a Tauri app and press the back button - it will destroy the activity, and reopening the app will crash due to some global state trying to reinitialize (like a log plugin initialization or any other global singleton values)

closes https://github.com/tauri-apps/tauri/issues/13568

### DIFF
--- a/.changes/android-destroy.md
+++ b/.changes/android-destroy.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Trigger `WindowEvent::Destroyed` when the Android activity is destroyed. In this case, the app should either exit by setting the control flow to `ControlFlow::ExitWithCode` or NOT call the create() external function when the activity is recreated and `onCreate` is called, handling how to recreate the app window via a separate hook that can leverage the existing tao event loop.
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -184,6 +184,17 @@ impl<T: 'static> EventLoop<T> {
               }
             );
           }
+          Event::Destroy => {
+            call_event_handler!(
+              event_handler,
+              self.window_target(),
+              control_flow,
+              event::Event::WindowEvent {
+                window_id: window::WindowId(WindowId),
+                event: event::WindowEvent::Destroyed,
+              }
+            );
+          }
           _ => {}
         },
         Some(EventSource::InputQueue) => {


lucasfernog:
note that Tauri will notice the only window is getting destroyed and automatically exit the event loop; if we want to keep the Rust app alive while the activity is destroyed, we'll need to enhance the lifecycle management on wry and tauri (this includes adding more hooks to detect app restart, currently we only have App::setup to initialize the app window and any other global state, which can't really be reused for app restarts - since you might do some one-off tasks such as initializing the log plugin).

FabianLars (approved):
i think i was never able to reproduce the issue so i didn't test this but it sounds reasonable.

lucasfernog:
> i think i was never able to reproduce the issue so i didn't test this but it sounds reasonable.

@FabianLars back button to close app and then reopen

FabianLars:
yeah, never had problems with that. it worked fine earlier today when i tested the back button default exit behavior in the context of https://github.com/tauri-apps/tauri/pull/14133 (in an emulator)

FabianLars:
actually nvm, that test app today was too basic for this

lucasfernog:
i mean you might not see the crash if you do not initialize the log plugin or any other singleton thingy, you might just notice two entry point calls in the same app instance

FabianLars:
yeah idk i think i got that once just now? (versus none with the PR 🤷 )

PR #1154, opened 2025-11-17
feat(mobile): add multi window support

lucasfernog:
Leverages [scenes](https://developer.apple.com/documentation/uikit/scenes) on iOS and [Activity embedding](https://developer.android.com/develop/ui/views/layout/activity-embedding) on Android.

# iOS
- Added Event::SceneRequested (on iPad the user can request a new window to be open - e.g. by long pressing the app icon and selecting "New window")
- Request new scene to be created on Window::new (if needed, main scene is detected automatically) and assign the window instance later when it gets connected

# Android
- Create new activity on Window::new (if needed, main activity is detected automatically)
- Added builder methods to determine the activity to be created
- System determines what to do with the activity (new stack, next to another one.. based on the embedding rules)

### DIFF
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -32,17 +32,6 @@ lazy_static! {
 pub enum OsError {
   JniError(jni::errors::Error),
   NoAvailableActivity,
-  ActivityNotFound {
-    activity_id: ActivityId,
-  },
-  ActivityAlreadyCreated {
-    activity_id: ActivityId,
-  },
-  ActivityClassMismatch {
-    activity_id: ActivityId,
-    expected_class_name: String,
-    actual_class_name: String,
-  },
 }
 
 impl std::error::Error for OsError {}
@@ -51,20 +40,6 @@ impl std::fmt::Display for OsError {
     match self {
       OsError::JniError(e) => write!(f, "JNI error: {e}"),
       OsError::NoAvailableActivity => write!(f, "no available activity"),
-      OsError::ActivityNotFound { activity_id } => write!(f, "activity not found: {activity_id}"),
-      OsError::ActivityAlreadyCreated { activity_id } => {
-        write!(f, "activity already created: {activity_id}")
-      }
-      OsError::ActivityClassMismatch {
-        activity_id,
-        expected_class_name,
-        actual_class_name,
-      } => {
-        write!(
-          f,
-          "activity class mismatch: {activity_id}, expected {expected_class_name} but got {actual_class_name}"
-        )
-      }
     }
   }
 }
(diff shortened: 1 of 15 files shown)


lucasfernog:
needs https://github.com/tauri-apps/wry/pull/1633

src/platform_impl/ios/window.rs
```diff
@@ -68,17 +92,44 @@ impl Inner {
...
+        window.makeKeyAndVisible();
+        return;
+      };
+      let mtm = MainThreadMarker::new().unwrap();
+      let application = UIApplication::sharedApplication(mtm);
+
+      // when we support multiple scenes, request the activation of this window's scene
+      if application.supportsMultipleScenes() {
+        // alternative function is iOS 17+
+        #[allow(deprecated)]
```
velocitysystems: Since this API is deprecated and will be removed in the future we should use [activateSceneSession(for:errorHandler:)](https://developer.apple.com/documentation/uikit/uiapplication/activatescenesession(for:errorhandler:)).

Can we use conditional logic based on availability? e.g.

```rust
if #available(iOS 17, *) {
   // Use supported API
} else {
   #[allow(deprecated)]
   // Fallback to deprecated API
}
```

Also - no error handler is set which means scene activation failures will be _invisible_.
Should we add logging and an optional fallback path?

src/platform_impl/ios/view.rs
```diff
@@ -526,6 +533,55 @@ pub unsafe fn create_window(
...
+          .requesting_scene_identifier
+          .as_ref()
+          .and_then(|id| app_state::scene_by_id(id))
+        {
+          options.setRequestingScene(Some(&scene));
+        }
+
+        // alternative function is iOS 17+
+        #[allow(deprecated)]
+        application.requestSceneSessionActivation_userActivity_options_errorHandler(
```
velocitysystems, line 596: Using deprecated API, no error handler is set. See this [comment](https://github.com/tauri-apps/tao/pull/1154/files#r2560518031).

src/platform_impl/ios/window.rs
```diff
@@ -449,6 +499,18 @@ impl Inner {
   pub fn set_badge_count(&self, count: i32) {
     set_badge_count(count);
   }
+
+  // instead of returning an Option here, we default to an empty string
+  // scene lifecycle will be enforced anyway soon (iOS 27)
+  pub fn scene_identifier(&self) -> String {
```
velocitysystems, line 517: Multiple windows without scenes get the same `""` identifier.
Should we instead return `Option<String>` and handle downstream?

src/platform_impl/android/mod.rs
```diff
@@ -143,155 +185,125 @@ impl<T: 'static> EventLoop<T> {
           }
           Event::Stop => self.running = false,
           Event::Start => self.running = true,
```
velocitysystems, line 186: Handlers for `Event::Start` and `EventSource::InputQueue` have been commented out. Is additional work required?

src/platform_impl/ios/window.rs
lucasfernog: if you're not using scenes, doesn't make much sense to use this getter
returning an Option here is inconvenient - scenes will be enforced next year, and Apple is already giving warnings

src/platform_impl/android/mod.rs
lucasfernog: those variants simply weren't being used

src/platform_impl/ios/view.rs
lucasfernog: [`f45fe1b`](https://github.com/tauri-apps/tao/pull/1154/commits/f45fe1bc3610e27f11c0701e43d969a6223e3e0c)

src/platform_impl/ios/window.rs
lucasfernog: [`f45fe1b`](https://github.com/tauri-apps/tao/pull/1154/commits/f45fe1bc3610e27f11c0701e43d969a6223e3e0c)

FabianLars (approved):

PR #1155, opened 2025-11-17
feat(android): trigger Event::Opened

lucasfernog:


lucasfernog:
needs #1154 to be merged first

src/platform_impl/android/ndk_glue.rs
```diff
@@ -436,6 +473,198 @@ pub unsafe fn onWindowFocusChanged(
   wake(event);
 }
 
+#[allow(non_snake_case)]
+pub unsafe fn onNewIntent(env: JNIEnv, _: JClass, intent: JObject) {
```
velocitysystems, line 485: There are a number of silent failure paths in this function e.g.

```rust
let Ok(action) = env.get_string(&action)
  .map(|action| action.to_string_lossy().to_string())
else {
  return; // Silent failure
};
```

Should these errors be propagated or logged?

src/platform_impl/android/ndk_glue.rs
```diff
@@ -188,6 +210,7 @@ lazy_static::lazy_static! {
   pub(crate) static ref CONTEXTS: Mutex<BTreeMap<ActivityId, AndroidContext>> = Mutex::new(Default::default());
   static ref WINDOW_MANAGER: Mutex<BTreeMap<ActivityId, GlobalRef>> = Mutex::new(Default::default());
   pub(crate) static ref ACTIVITY_CREATED_SENDERS: Mutex<BTreeMap<ActivityId, Sender<()>>> = Mutex::new(Default::default());
+  static ref INTENT_URLS: Mutex<Vec<url::Url>> = Mutex::new(Default::default());
```
velocitysystems: Using global mutable state for intent URLs. How does this interact with multi-window support? What if multiple windows exist? How does the intent get delivered to the correct window?

src/platform_impl/android/ndk_glue.rs
```diff
@@ -33,6 +34,26 @@ use std::{
 /// in the android project.
 pub static PACKAGE: OnceCell<&str> = OnceCell::new();
 
+/// Character set for encoding text content in data URLs.
+/// Encodes all control characters and special characters that might cause issues in URLs.
+const DATA_URL_ENCODING_SET: &AsciiSet = &CONTROLS
```
velocitysystems, line 42: Tauri converts shared text to data URLs to provide a unified `Event::Opened { urls: Vec<Url> }` API across all content types (deep links, files, and text). Is there a specific reason why this pattern is used?

Other cross-platforms frameworks (such as React Native) handle this by splitting out the different content types e.g.

```rust
pub enum Event {
    Opened { urls: Vec<Url> },        // Deep links & file URIs only
    TextShared { text: String },       // Raw shared text
    // ...
}
```

Alternatively a tagged union could be used to handle the different content types:

```rust
pub enum OpenedContent {
    Url(Url),
    Text(String),
}

pub enum Event {
    Opened { content: Vec<OpenedContent> },
    // ...
}
```

src/platform_impl/android/ndk_glue.rs
```diff
@@ -436,6 +473,198 @@ pub unsafe fn onWindowFocusChanged(
...
+    && action != "android.intent.action.VIEW"
+    && action != "android.intent.action.SEND_MULTIPLE"
+  {
+    return;
+  }
+
+  let mut urls = HashSet::new();
+
+  // Get intent type (may be null)
+  let intent_type: JString = env
```
velocitysystems: This defaults to an empty string if null. Would it be better to use a `Option<string>` here?

src/platform_impl/android/ndk_glue.rs
lucasfernog: the intent URL is not delivered to a particular window - it is instead sent as a global event Event::Opened
i wanted to include the URLs in the android Event enum but I can't since it has to implement Copy - that's why I'm using this global variable.

src/platform_impl/android/ndk_glue.rs
lucasfernog: > Is there a specific reason why this pattern is used?

backwards compatibility

tho the new variant is probably ok, would just be a little weird to have two different paths for the same interface (intent data)

src/platform_impl/android/ndk_glue.rs
lucasfernog: doesn't matter, the type is not exposed, we just check it to see the data format, but i'll change it anyway

src/platform_impl/android/ndk_glue.rs
lucasfernog: [`2268fac`](https://github.com/tauri-apps/tao/pull/1155/commits/2268fac61116813c2f8c4b04f30a050b37803c84)

FabianLars (approved):

PR #1157, opened 2025-11-21
fix(windows): exit process on `WM_ENDSESSION`

Legend-Master:
Reference: https://github.com/tauri-apps/tauri/pull/14479#issuecomment-3562121016

Open to suggestion on this one, we can detect `lParam` here to do special handling to Restart Manager but I honestly feel like the Windows shutdown behavior itself is a bit of a myth so might just to the exit ourselves regardlesss


Legend-Master:
After the testing in https://github.com/rust-windowing/winit/pull/4425#discussion_r2558437389, I think it should be possible to exit normally without the `std::process:exit` here, but that requires us to adopt https://github.com/rust-windowing/winit/pull/2767 or our current way of pulling the message through `GetMessageW` (`GetMessageW` dispatches none queued messages by itself and only returns on queued messages) will not return to us (or maybe we can post a message to the message queue?)

src/platform_impl/windows/event_loop.rs
```diff
@@ -2414,8 +2414,14 @@ unsafe extern "system" fn thread_event_target_callback<T: 'static>(
       // and we don't need to handle that case since we didn't do anything prior in response to `WM_QUERYENDSESSION`
       if wparam.0 == TRUE.0 as usize {
         subclass_input.event_loop_runner.loop_destroyed();
+        // `WM_ENDSESSION` can be sent by Windows during shutdown or by Restart Manager
+        //
+        // - From Windows shutdown: Windows will shut us down after we return `0` here
+        // - From Restart Manager: Restart Manager only sends the message but will wait for us to shutdown ourselves
+        //
+        // So, we call `exit` here directly to align this behavior
+        std::process::exit(0);
```
thomaseizinger, line 2337: This does also happen in the `run_return` case? I would really like to not have this then as we want to do some graceful cleanup before exiting.

Legend-Master: > This does also happen in the run_return case?

This happens in all cases when you spin the event loop, and yes, including `run_return`. One thing to note here is no matter what we do here, during shutdown, Windows will terminate us anyways

This does trigger `Exit` event for you to do some clean ups though

thomaseizinger: > This does trigger `Exit` event for you to do some clean ups though

I don't quite understand that. If you call process exit here, no more code can execute after this, right?

I don't know the details of `WM_ENDSESSION` but why can't we just gracefully exit the event-loop here? What would be the point of a signal from Windows if we can't run a few more instructions to shut down?

Exiting the process should IMO be reserved for the binary that embeds Tauri and not happen in Tauri itself.

Legend-Master: > I don't quite understand that. If you call process exit here, no more code can execute after this, right?

Yes, the event handler is called in the line above

> subclass_input.event_loop_runner.loop_destroyed();

---

> I don't know the details of WM_ENDSESSION but why can't we just gracefully exit the event-loop here? What would be the point of a signal from Windows if we can't run a few more instructions to shut down?

> Exiting the process should IMO be reserved for the binary that embeds Tauri and not happen in Tauri itself.

In the Windows shutdown case, you're supposed to block the event loop (not returning from the window procedure) to do the clean up, and once you return, Windows terminates the app if you don't do it yourselves

For the restart manager case, see my comment above https://github.com/tauri-apps/tao/pull/1157#issuecomment-3578870725

FabianLars (approved):
blind approval - maybe wait for @amrbashir idk

amrbashir (approved):

PR #1158, opened 2025-11-21
chore: add missing licenses to tao-macros

### LINKED ISSUE #1156: Question about tao-macros crate license
The tao-macros crate is license under MIT or Apache-2.0 but there is no MIT license file in the repo. And I noticed that the tao project is licensed under Apache-2.0 only. Is the MIT file missing or is the tao-macros crate actually supposed to be released under Apache-2.0 only just like tao crate?

FabianLars:
closes #1156

### DIFF
--- a/tao-macros/LICENSE-MIT
+++ b/tao-macros/LICENSE-MIT
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020-2023 Ngo Iok Ui & Tauri Programme within The Commons Conservancy
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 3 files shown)


tao-macros/LICENSE.spdx
```diff
@@ -0,0 +1,19 @@
+SPDXVersion: SPDX-2.1
+DataLicense: CC0-1.0
+PackageName: tao
```
Legend-Master: Not sure but should this be `tao-macros`? (also the descriptions)

tao-macros/LICENSE-MIT
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020-2023 Ngo Iok Ui & Tauri Programme within The Commons Conservancy
```
Legend-Master: Should this be something like

> Copyright (c) 2017 - Present Tauri Apps Contributors

in https://github.com/tauri-apps/tauri/blob/dev/LICENSE_MIT ?

FabianLars: guess so, but perhaps just 2022 - Present to match the crates lifetime?

tao-macros/LICENSE.spdx
FabianLars: missed that one but kept the description by choice. since idk where that is used though i'll update it

Legend-Master (approved):

PR #1173, opened 2026-01-19
Unsound synchronization on Linux

sftse:
Setting the theme from multiple threads will result in undefined behavior on Linux.

`Window` is `Sync` but `RefCell` is not. The methods that set the theme and fullscreen property take a `&Window` and will internally launder it into a `&RefCell` accessible from multiple threads.

@Legend-Master

### DIFF
--- a/.changes/change-pr-1173.md
+++ b/.changes/change-pr-1173.md
@@ -1,0 +1,7 @@
+---
+"tao": patch
+---
+
+`fn set_min_inner_size`, `fn set_max_inner_size`, `fn set_inner_size_constraints`,
+`fn set_fullscreen` and `fn set_theme` on `Window` were not properly thread-safe
+on Linux.
--- a/src/platform_impl/linux/window.rs
+++ b/src/platform_impl/linux/window.rs
@@ -62,10 +62,16 @@ pub struct Window {
   maximized: Rc<AtomicBool>,
   is_always_on_top: Rc<AtomicBool>,
   minimized: Rc<AtomicBool>,
+  // `Window` is `Send` and `Sync`, need a `RwLock` not a `RefCell`
+  // otherwise unsynchronized &RefCell from multiple threads
   fullscreen: RwLock<Option<Fullscreen>>,
+  // `Window` is `Send` and `Sync`, need a `RwLock` not a `RefCell`
+  // otherwise unsynchronized &RefCell from multiple threads
   inner_size_constraints: RwLock<WindowSizeConstraints>,
   /// Draw event Sender
   draw_tx: crossbeam_channel::Sender<WindowId>,
+  // `Window` is `Send` and `Sync`, need a `RwLock` not a `RefCell`
+  // otherwise unsynchronized &RefCell from multiple threads
   preferred_theme: RwLock<Option<Theme>>,
   css_provider: CssProvider,
 }
--- a/src/platform_impl/linux/window.rs
+++ b/src/platform_impl/linux/window.rs
@@ -6,7 +6,10 @@ use std::{
   cell::RefCell,
   collections::VecDeque,
   rc::Rc,
-  sync::atomic::{AtomicBool, AtomicI32, Ordering},
+  sync::{
+    atomic::{AtomicBool, AtomicI32, Ordering},
+    RwLock,
+  },
 };
 
 use gtk::{
@@ -59,11 +62,11 @@ pub struct Window {
   maximized: Rc<AtomicBool>,
   is_always_on_top: Rc<AtomicBool>,
   minimized: Rc<AtomicBool>,
-  fullscreen: RefCell<Option<Fullscreen>>,
-  inner_size_constraints: RefCell<WindowSizeConstraints>,
+  fullscreen: RwLock<Option<Fullscreen>>,
+  inner_size_constraints: RwLock<WindowSizeConstraints>,
   /// Draw event Sender
   draw_tx: crossbeam_channel::Sender<WindowId>,
-  preferred_theme: RefCell<Option<Theme>>,
+  preferred_theme: RwLock<Option<Theme>>,
   css_provider: CssProvider,
 }
 
@@ -272,9 +275,9 @@ impl Window {
       maximized,
       minimized,
       is_always_on_top,
-      fullscreen: RefCell::new(attributes.fullscreen),
-      inner_size_constraints: RefCell::new(attributes.inner_size_constraints),
-      preferred_theme: RefCell::new(preferred_theme),
+      fullscreen: RwLock::new(attributes.fullscreen),
+      inner_size_constraints: RwLock::new(attributes.inner_size_constraints),
+      preferred_theme: RwLock::new(preferred_theme),
       css_provider: CssProvider::new(),
     };
 
@@ -415,9 +418,9 @@ impl Window {
       maximized,
       minimized,
       is_always_on_top,
-      fullscreen: RefCell::new(None),
-      inner_size_constraints: RefCell::new(WindowSizeConstraints::default()),
-      preferred_theme: RefCell::new(None),
+      fullscreen: RwLock::new(None),
+      inner_size_constraints: RwLock::new(WindowSizeConstraints::default()),
+      preferred_theme: RwLock::new(None),
       css_provider: CssProvider::new(),
     };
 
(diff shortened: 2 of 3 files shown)


sftse:
Fixed CI

FabianLars:
can't wait for the day we're finally back with winit. happy to see you getting rid of lazy_static :)

> refactor: remove one mem::transmute

are you sure that's not needed? i don't know what it does from the top of my head but winit also still has that code.

sftse:
> are you sure that's not needed? i don't know what it does from the top of my head but winit also still has that code.

It's not unsound, since [ToUnicode](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/Input/KeyboardAndMouse/fn.ToUnicode.html) takes a `&mut [u16]` (that's just dandy, so what is the safety contract then at all? docs say nothing) but it took me a serious double-take because of the `&mut` to uninitialized memory. I think this used to be UB but maybe got changed? Not sure, but doesn't matter that much to me, can revert if you want.

Edit: This doesn't totally explain that the current use of `mem::transmute` is sound, but I did check it with MIRI to be 100% sure that at least by itself the `mem::transmute` can't be unsound.

sftse:
Dropped the `mem::transmute` change

sftse:
The diffs migrating from `lazy_static` to `once_cell` are a bit tedious to verify, difftastic helps a lot here.

Legend-Master:
Genuine question, what's benefit of switching from `lazy_static` to `once_cell`?

And if we want to do it, I would prefer we split it in a separate PR so when we squash merge them, we get 2 different commits for checking with git blame later

Also, you'll need to sign your commits 🙃

sftse:
> Genuine question, what's benefit of switching from lazy_static to once_cell?

I've seen `lazy_static` macros obscure dead code detection, they are a nuisance to autoformat and the `once_cell` api is the one that got stabilized to std, so the future change to the stabilized api is just a switching of types.
Additionally, pervasively used crates in the ecosystem harm the ability of `cargo-depgraph` to layout graphs in an intelligible way.

Not really strong arguments, but grug dev see unnecessary dependency, grug dev remove.

FabianLars:
less macros = more good

sftse:
To not make the same mistake as in my other PR, here's a program that crashes

```rust
fn main() {
    use tao::window::Fullscreen;
    let event_loop = tao::event_loop::EventLoop::new();
    let win = tao::window::Window::new(&event_loop).unwrap();
    std::thread::scope(|s| {
        s.spawn(|| loop {
            win.set_fullscreen(None);
        });
        s.spawn(|| loop {
            win.set_fullscreen(Some(Fullscreen::Borderless(None)));
        });
    });
}
```
`Memory fault`

src/platform_impl/linux/window.rs
```diff
@@ -288,9 +291,9 @@ impl Window {
       maximized,
       minimized,
       is_always_on_top,
-      fullscreen: RefCell::new(attributes.fullscreen),
-      inner_size_constraints: RefCell::new(attributes.inner_size_constraints),
-      preferred_theme: RefCell::new(preferred_theme),
+      fullscreen: RwLock::new(attributes.fullscreen),
```
Legend-Master, line 284: I think the chances for read/write are not different by too much, maybe we should use `Mutex` here?

Legend-Master (commented):
I am not the most familiar with the Linux implementation, the PR's code looks good and correct to me though

For the theme part, I initially did it through `Mutex` in #937 and later @amrbashir changed it to `RefCell` https://github.com/tauri-apps/tao/pull/937/commits/e595125d96f80746cbd37ff3c22eaaa1f05ddc73, so maybe there's a reason for this?

src/platform_impl/linux/window.rs
sftse: I used a `RwLock` because that is the most direct port of a `RefCell` to the multithreaded case. A `Mutex` can't have multiple readers, `RwLock` and `RefCell` can.

But scopes here make it clear there can't be deadlocks regardless of `Mutex` or `RwLock`, so we can move to `Mutex`.

sftse:
Any chance to merge this? We can switch to `Mutex` if that is required, but such a trivial unsoundness needs a fix asap.

Legend-Master:
Looking at the code, it seems like we use `Rc` and `RefCell` in `Windows` and then implement `Send` and `Sync`, this doesn't look right to me 

> Any chance to merge this? We can switch to `Mutex` if that is required, but such a trivial unsoundness needs a fix asap.

I don't really know the reason for us to mark the `Window` as `Send + Sync` even though we use `Rc` and `RefCell` internally, so I want to reach out whoever knows it first before merging (must have a reason right?) cc @tauri-apps/wg-webview

sftse:
The code doesn't look very defensively written, but when I was taking a look at the `Rc` it didn't seem too problematic. It would be dangerous if we had a derived `Clone` on `Window`, but because the `Rc` are private and none of them seem cloneable from a public method, the boundaries seem to clearly distinguish outside from the inside. As long as no `Rc` crosses that boundary, we should be fine, on either side of it the mistakes should be caught by Rust's Send-Sync traits.

sftse:
Here's what I understood, we can only move the `Rc` handles as a unit across thread boundaries.
```rust
mod foo {
    use std::rc::Rc;
    unsafe impl Send for Foo {}

    // UNSOUND: #[derive(Clone)]
    pub struct Foo(Rc<()>);
    impl Foo {
        pub fn new() -> Foo {
            Foo(Rc::new(()))
        }
    }
}

use foo::Foo;

fn main() {
    let foo = Foo::new();
    // safe, can only move all Rc handles across thread boundary at the same time
    std::thread::spawn(move || foo);

    // UNSOUND: if we can clone, split Rc handles across threads
    // let handle = foo.clone();
    // std::thread::spawn(move || handle);
}
```

That said, the atomics inside the `Rc<AtomicI32>` seem unnecessary in that case.

Legend-Master:
It is only fine if the ones we initially cloned to the `ApplicationWindow` callback functions are also moved and dropped together which seems to be the case though, but now, `ApplicationWindow` is marked `!Sync` and `!Send`...

sftse:
I see there are some additional things to investigate, but if we can't resolve those in the near future, does it make sense to delay this?

`Window` on winit is Send and Sync, so we probably want to retain those if we want to move back to mainline.

Legend-Master:
After getting myself more familiar with tao, I think this is really just the results of the lack of internal documentations.

The `Window` is `Send + Sync` similar to how Tauri's `Window` works, it's for the convenient of the caller and our inner window should make sure it only get accessed on the main thread.

With that, since `Window` is not `Clone`, you should not be able to get multiple of it to cause problems with `RefCell` so that's why it's being used. But apparently nobody thought of it can be non mutably borrowed to multiple threads...

The atomic values were probably used just to prevent the use of `Cell`s which are a little more verbose

src/platform_impl/linux/window.rs
```diff
@@ -59,11 +62,11 @@ pub struct Window {
   maximized: Rc<AtomicBool>,
   is_always_on_top: Rc<AtomicBool>,
   minimized: Rc<AtomicBool>,
-  fullscreen: RefCell<Option<Fullscreen>>,
-  inner_size_constraints: RefCell<WindowSizeConstraints>,
+  fullscreen: RwLock<Option<Fullscreen>>,
```
Legend-Master, line 67: Let's add some notes on why we did this so the next person seeing this won't be as confused as us

Legend-Master (approved):
I want to wait a review from @amrbashir if he responds this week or we can just merge it since it shouldn't cause much of a performance issue anyways.

Also let's add a change file

.changes/change-pr-1173.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Use thread safe primitives to prevent unsynchronized access to data
```
Legend-Master: Since this is for the end user, could we just use the description you wrote in this PR's description?

> Setting the theme from multiple threads will result in undefined behavior on Linux.

And maybe add in a mention for `inner_size_constraints` and `fullscreen`

sftse: I've added a different suggestion, mentioning the public api that was unsafe, WDYT?

Legend-Master: I would add the context **Linux** and **Window**, looks good otherwise

Legend-Master:
Since amr didn't respond in time, I'm going to just merge it now

Issue #641, opened 2022-12-09
Mac window freezes


TheButlah:
I have a tauri application that works properly on windows and linux. Passing RUST_LOG=trace to them doesn't show anything abnormal.

On Mac however, I get this:
<img width="693" alt="Screen_Shot_2022-12-09_at_1 10 18_AM" src="https://user-images.githubusercontent.com/6969415/206677411-6caf3499-821b-48ea-82b0-3024e247f265.png">

The application appears to display correctly - I see what I would expect to see when launching the app. But I cannot click on it or interact with it in any way. I have to force quit the app to get it to exit.

Source code is [here](https://github.com/SlimeVR/SlimeVR-Server/tree/f04f833939f707a414480a4599d6c6d090d594bc/gui)

I havent tried with any of the example Tauri apps yet, I'll try that soon.

My guess was that the log spam is blocking the app and there is something that is spamming extraneous windowing events.

amrbashir:
Please make a minimal repro app, with `tao` only if you think it is related to tao only.

ImUrX:
This is a [tauri-plugin-window-state](https://github.com/tauri-apps/tauri-plugin-window-state) issue, just verified by disabling it.

ImUrX:
Mainly what gets triggered is resize events, so we will have to investigate how to detect this or just disable when using macOS

PR #1181, opened 2026-02-02
fix(macos): update is_maximized method to use ns_window's isZoomed

Tunglies:
It calls `set_style_mask_sync` twice, forces synchronous IPC. Temporarily adding and removing that triggers unnecessary window layout recalculations and repaints. In multi-thread, modifing the `styleMask` while view is till initializing leads to the first responder errors and precondition failure crashes. Or lead to unexpected resize behavior.

https://github.com/tauri-apps/tao/blob/648161769c35ddd2bffd939f4d4d6a027325326c/src/platform_impl/macos/window.rs#L976-L995

This PR related to:
https://github.com/tauri-apps/plugins-workspace/pull/3242
https://github.com/tauri-apps/tauri/issues/5812
https://github.com/agmmnn/tauri-controls/issues/10
https://github.com/tauri-apps/tauri/issues/10261
https://github.com/tauri-apps/tao/issues/641

The `is_zoomed` is setting a temporary, then reading. Leads to

> This seems to trigger an infinite loop of sorts https://github.com/tauri-apps/tauri/issues/5812

I will give another PR to change `is_zoomed` latter.

CC @Legend-Master, @amrbashir

### DIFF
--- a/.changes/fix-macos-is_maximized.md
+++ b/.changes/fix-macos-is_maximized.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Update is_maximized method to use ns_window's isZoomed
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -1077,7 +1077,7 @@ impl UnownedWindow {
 
   #[inline]
   pub fn is_maximized(&self) -> bool {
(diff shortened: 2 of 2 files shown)


FabianLars:
I think undecorated windows are too common in tauri for this to be an acceptable approach.

PR #1188, opened 2026-02-19
fix(windows): do not panic when transitioning from `Destroyed`

### LINKED ISSUE #1180: [Windows] Panic in EventLoopThreadExecutor::execute_in_thread (Re-entrancy / Destroyed state)
**Describe the bug**
We are experiencing panics in our production environment (Tauri application) originating from `tao::platform_impl::platform::event_loop::EventLoopThreadExecutor::execute_in_thread` on the Windows backend.

Based on the panic messages, it appears to be related to internal state management within the Windows event loop runner, specifically involving re-entrancy checks or accessing the event loop state after it has been destroyed.

**Steps To Reproduce**
Steps to reproduce the behavior. It **must** use tao directly and not tauri.

1.  This issue is observed in a production Tauri application (v2.8.5).
2.  Due to the nature of the error (likely a race condition or specific timing during window destruction/heavy load), we have not been able to isolate a minimal reproduction using `tao` directly yet.
3.  However, the stack traces collected point directly to `tao`'s internal logic on Windows, suggesting unsafe state transitions during the event loop processing.

**Expected behavior**
The event loop should handle state transitions gracefully. Even if an event handler is re-entrant or the state is destroyed, it should ideally not cause a hard panic that crashes the application.

**Screenshots**
N/A (See stack trace details in "Additional context")

OS: Windows 10.0.26100 x86_64
Rustc: 1.89.0
Tao version: 0.34.3

**Additional context**
We have identified three distinct panic messages/locations in the stack traces that seem to stem from the same root cause:

1.  **Re-entrancy / No Handler:**
    `"either event handler is re-entrant (likely), or no event handler is registered (very unlikely)"`
    Location: `src/platform_impl/windows/event_loop/runner.rs:244`

2.  **Assertion Failure:**
    `"assertion failed: flush_paint_messages(None, &subclass_input.event_loop_runner)"`
    Location: `src/platform_impl/windows/event_loop.rs:2344`

3.  **Invalid State Transition:**
    `"cannot move state from Destroyed"`
    Location: `src/platform_impl/windows/event_loop/runner.rs:368`

It appears that the event loop is attempting to access state or flush messages after the runner state has arguably become invalid or destroyed.

---

**Full `tauri info` output:**

```text
[✔] Environment
    - OS: Windows 10.0.26100 x86_64 (X64)
    ✔ WebView2: 144.0.3719.93
    ✔ MSVC:
        - Visual Studio Professional 2017
        - Visual Studio Professional 2019
    ✔ rustc: 1.89.0 (29483883e 2025-08-04)
    ✔ cargo: 1.89.0 (c24e10642 2025-06-23)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 22.14.0
    - yarn: 1.22.22
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.8.5
    - tauri-build 🦀: 2.4.1
    - wry 🦀: 0.53.3
    - tao 🦀: 0.34.3
    - @tauri-apps/api : 2.2.0
    - @tauri-apps/cli : 2.2.2

[-] Plugins
    - tauri-plugin-updater 🦀: 2.9.0
    - @tauri-apps/plugin-updater : not installed!
    - tauri-plugin-fs 🦀: 2.4.2
    - tauri-plugin-log 🦀: 2.7.0
    - @tauri-apps/plugin-log : not installed!
    - tauri-plugin-http 🦀: 2.5.2
    - @tauri-apps/plugin-http : not installed!
    - tauri-plugin-dialog 🦀: 2.4.0
    - @tauri-apps/plugin-dialog : 2.0.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: Vue.js
    - bundler: Vite

### PR DESCRIPTION
**Pull request by thomaseizinger**
We are seeing crash reports from Tauri applications caused by this particular panic. I am not sure what they are caused by but it seems save to just remove the panic, log a message and leave the state in `Destroyed` instead.

Resolves: #1180

### DIFF
--- a/src/platform_impl/windows/event_loop/runner.rs
+++ b/src/platform_impl/windows/event_loop/runner.rs
@@ -365,7 +365,9 @@ impl<T> EventLoopRunner<T> {
         self.call_event_handler(Event::LoopDestroyed);
       }
 
-      (Destroyed, _) => panic!("cannot move state from Destroyed"),
+      (Destroyed, state) => {
+        debug!("Eventloop is already destroyed, cannot move to {state}")
+      }
     }
   }
 
--- a/src/platform_impl/windows/event_loop/runner.rs
+++ b/src/platform_impl/windows/event_loop/runner.rs
@@ -6,7 +6,7 @@ use std::{
   any::Any,
   cell::{Cell, RefCell},
   collections::{HashSet, VecDeque},
-  mem, panic,
+  fmt, mem, panic,
   rc::Rc,
   time::Instant,
 };
@@ -59,6 +59,18 @@ enum RunnerState {
   HandlingRedrawEvents,
   /// The event loop has been destroyed. No other events will be emitted.
   Destroyed,
+}
+
+impl fmt::Display for RunnerState {
+  fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    match self {
+      RunnerState::Uninitialized => write!(f, "Uninitialized"),
+      RunnerState::Idle => write!(f, "Idle"),
+      RunnerState::HandlingMainEvents => write!(f, "HandlingMainEvents"),
+      RunnerState::HandlingRedrawEvents => write!(f, "HandlingRedrawEvents"),
+      RunnerState::Destroyed => write!(f, "Destroyed"),
+    }
+  }
 }
 
 enum BufferedEvent<T: 'static> {
--- a/src/platform_impl/windows/event_loop/runner.rs
+++ b/src/platform_impl/windows/event_loop/runner.rs
@@ -6,7 +6,7 @@ use std::{
   any::Any,
   cell::{Cell, RefCell},
   collections::{HashSet, VecDeque},
-  fmt, mem, panic,
+  mem, panic,
   rc::Rc,
   time::Instant,
 };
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by Legend-Master — APPROVED** [review]
I am in favor of this, maybe we could discuss with the winit maintainers about the reasoning behind this before merging though (or we should identify the reason we reached this line if this is an internal state error that should not happen and the panic is the right thing to do)

File: src/platform_impl/windows/event_loop/runner.rs [hunk 1]
```diff
@@ -365,7 +377,9 @@ impl<T> EventLoopRunner<T> {
         self.call_event_handler(Event::LoopDestroyed);
       }
 
-      (Destroyed, _) => panic!("cannot move state from Destroyed"),
+      (Destroyed, state) => {
+        debug!("Eventloop is already destroyed, cannot move to {state}")
```
Legend-Master: I think we can just use `:?` here for the debug print

thomaseizinger: We can, it just won't be consistent with the other log messages where we only print the variant name. Your call!

Legend-Master: Sounds fine to me, let's do it

Legend-Master:
After a closer look at the current code, we only ever move the state to destroyed on `run_return` ends (which has a reset call follows it immediately) and on `WM_ENDSESSION`

For the `run_return` case, it seems to me that it's impossible to trigger this panic since we only receive events on the main thread and since we reset the states immediately, this should not happen
For the `WM_ENDSESSION` case, the fix should be #1157 instead of this

thomaseizinger:
> For the `run_return` case, it seems to me that it's impossible to trigger this panic since we only receive events on the main thread and since we reset the states immediately, this should not happen

We are using `run_return` in our application FWIW.

Legend-Master:
> We are using run_return in our application FWIW.

`run` also uses `run_return` internally, since the state is reset to `Uninitialized` in literally next line, I don't think it's possible to hit (unless we did some other violations like calling `move_state_to` in another thread)

PR #1190, opened 2026-02-24
fix(linux): exit cleanly when GtkApplication is a remote GIO instance

damadczar:
## Problem

When two processes share the same `GtkApplication` app_id (e.g. a GUI application and a background webserver using the same binary), the second process to register on D-Bus becomes a **remote GIO instance**. `g_application_register()` succeeds but `priv->impl` is never initialized — the `startup` and `activate` signals only fire on the primary instance.

If the remote process then proceeds to create a `GtkApplicationWindow`, GTK calls:

```
gtk_window_set_application()
  → gtk_application_add_window()
    → g_signal_emit(WINDOW_ADDED)
      → dispatches into priv->impl  ← NULL
        → SIGSEGV
```

## Fix

After `app.activate()`, check `app.is_remote()` (which wraps `g_application_get_is_remote()`). If true, return immediately with exit code `0`.

GIO has already forwarded the activate signal to the primary instance via D-Bus at this point, so the primary can respond — for example, by handling the second `StartCause::Init` that tao's `connect_activate` closure emits and bringing its existing window to the front.

## Behavior

- **Before:** Second launch hangs forever (remote instance never receives `StartCause::Init` because `connect_activate` only fires on the primary), then crashes when it tries to create a window.
- **After:** Second launch exits cleanly with code 0. The primary receives a D-Bus activate, fires `connect_activate` again, and the application can respond by focusing its existing window.

## Testing

Reproduce by building any tao app with a fixed `app_id`, running two instances simultaneously on Linux. Without this fix the second instance crashes; with it, the second exits cleanly.

### DIFF
--- a/.changes/change-pr-1190.md
+++ b/.changes/change-pr-1190.md
@@ -1,5 +1,4 @@
 ---
-"tao-macros": patch
 "tao": patch
 ---
 
--- a/.changes/change-pr-1190.md
+++ b/.changes/change-pr-1190.md
@@ -1,0 +1,6 @@
+---
+"tao-macros": patch
+"tao": patch
+---
+
+fix(linux): exit cleanly when GtkApplication is a remote GIO instance
--- a/src/platform_impl/linux/event_loop.rs
+++ b/src/platform_impl/linux/event_loop.rs
@@ -1035,6 +1035,15 @@ impl<T: 'static> EventLoop<T> {
 
         window_target.p.app.activate();
 
+        // If this is a secondary (remote) GIO instance, the activate signal
+        // was forwarded to the primary instance via D-Bus. Exit immediately so
+        // the primary can handle focus (e.g. bring its window to front).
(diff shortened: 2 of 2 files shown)


damadczar:
This happened to me writing a busybox style application where you could launch it in a gui mode from the gnome desktop, and it also ran as a service via systemd user service.  If the webserver was running, the gui would seg fault.

Part of my fix was to no longer set app_id on the application when running as webserver, but in gui mode, without this fix it's impossible to have the gui application work as you'd expect.  Without it, when running the gui, you cannot have single instance behavior.

FabianLars:
thanks for the pr! can you also add a changefile like https://github.com/thomaseizinger/tao/blob/8962da69655b57881fbef07d47e40ab94e9785e3/.changes/no-panic.md please?

damadczar:
changes file added.

FabianLars:
thanks

FabianLars (approved):

Issue #1191, opened 2026-02-26
macOS 15 Sequoia: 100% CPU with borderless+resizable windows (NSThemeZoomWidget leak)


EddieRydell:
## Description

On macOS 15 Sequoia, windows created with `decorations: false` + `resizable: true` (i.e., `NSWindowStyleMask::Borderless | NSWindowStyleMask::Resizable`) consume 100% CPU even when idle, due to `_NSThemeZoomWidget` objects being created on every layout pass.

## Root Cause

macOS 15 introduced a new window tiling system. As part of this, `NSThemeFrame.newZoomButton` is called on every layout pass for windows with `NSWindowStyleMask::Resizable` that don't have a title bar. Each call creates temporary `_NSThemeZoomWidget` objects that accumulate in the autorelease pool. Their dealloc calls `+[NSObject cancelPreviousPerformRequestsWithTarget:selector:object:]`, which iterates all run loop timers — an O(n²) operation.

## Reproduction

1. Create a Tauri v2 app (or any tao-based app) with `decorations: false` and `resizable: true`
2. Run on macOS 15.x Sequoia (tested on 15.6.1)
3. Observe 100% CPU usage even when the window is idle

## Process Sample Evidence

```
100.9% CPU, ~900MB RAM

Thread 0 (main):
  1960 AutoreleasePoolPage::releaseUntil
    678 -[_NSThemeZoomWidget dealloc]
      605 -[_NSThemeZoomWidgetCell dealloc]
        392 +[NSObject cancelPreviousPerformRequestsWithTarget:selector:object:]
```

## Relevant Code

**Window creation** (`src/platform_impl/macos/window.rs` lines 216-221):
```rust
let mut masks = if !attrs.decorations && screen.is_none() || pl_attrs.titlebar_hidden {
    NSWindowStyleMask::Borderless
        | NSWindowStyleMask::Resizable
        | NSWindowStyleMask::Miniaturizable
```

**`set_decorations(false)`** (lines 1362-1363):
```rust
NSWindowStyleMask::Borderless | NSWindowStyleMask::Resizable
```

## Tested Fixes

| Approach | CPU Fixed? | Side Effects |
|----------|-----------|--------------|
| `setCollectionBehavior(FullScreenNone)` + hide zoom button | No | Widgets still created by NSThemeFrame |
| Remove `NSWindowStyleMask::Resizable` | Yes (0% CPU) | Window cannot be resized |
[...]

PR #1182, opened 2026-02-02
fix(macos): remove synchronous styleMask mutations in is_zoomed

Tunglies:
Mentioned on https://github.com/tauri-apps/tao/pull/1181

fixes https://github.com/tauri-apps/plugins-workspace/issues/3240
fixes https://github.com/tauri-apps/plugins-workspace/pull/3242
fixes https://github.com/tauri-apps/tao/issues/641
fixes https://github.com/tauri-apps/tauri/issues/5812
fixes https://github.com/tauri-apps/tauri/issues/10261
fixes https://github.com/tauri-apps/tauri/issues/14822

CC @Legend-Master, @amrbashir

PR #486, opened 2022-07-19
Window::is_minimized()

henry40408:
closes #257 

### What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary

### Other information

N/A

PR #1192, opened 2026-02-26
fix(macos): prevent 100% CPU on macOS 15 for borderless+resizable windows

EddieRydell:
## Summary

On macOS 15 Sequoia, any window with `decorations: false` + `resizable: true` (i.e. `Borderless | Resizable` style mask) pegs the CPU at 100% even when idle. This has been a longstanding pain point on macOS — the underlying `is_zoomed()` style mask hack has caused freezes and infinite resize loops since at least 2022 — but macOS 15's new tiling system made it catastrophic by adding `_NSThemeZoomWidget` creation to every layout pass.

### Root cause

`is_zoomed()` temporarily flips the window's style mask from `Borderless | Resizable` to `Titled | Resizable`, calls `isZoomed()`, then flips it back. On macOS 15, each `setStyleMask:` call triggers a full AppKit layout pass, and `NSThemeFrame.newZoomButton` now creates `_NSThemeZoomWidget` objects on every such pass. These accumulate in the autorelease pool, and their dealloc calls `+[NSObject cancelPreviousPerformRequestsWithTarget:selector:object:]` — an O(n²) scan of all run loop timers.

Process sample from an idle window:
```
100.9% CPU, ~900MB RAM

Thread 0 (main):
  1960 AutoreleasePoolPage::releaseUntil
    678 -[_NSThemeZoomWidget dealloc]
      605 -[_NSThemeZoomWidgetCell dealloc]
        392 +[NSObject cancelPreviousPerformRequestsWithTarget:selector:object:]
```

### Changes

- **Remove the `is_zoomed()` style mask hack** — call `isZoomed()` directly instead of flipping the mask to `Titled | Resizable` and back. `isZoomed()` may return inaccurate results for borderless windows, but the performance tradeoff is clearly worthwhile. This eliminates the `setStyleMask:` storm that is the primary cause of the CPU spin.

- **Wrap `set_style_mask` in an autoreleasepool** — ensures any temporary objects created by `setStyleMask:` during legitimate calls (fullscreen transitions, decoration changes) are drained immediately rather than accumulating.

### Related issues and PRs

- #1191 — macOS 15 Sequoia: 100% CPU with borderless+resizable windows
- #1181 — update is_maximized method to use ns_window's isZoomed (same root cause, fixes only one call site)
- #1182 — remove synchronous styleMask mutations in is_zoomed (same root cause, uses frame-size comparison)
- tauri-apps/tauri#5812 — isMaximized triggers infinite resize loop (Dec 2022)
- tauri-apps/tauri#10261 — is_maximized + get_focused_window freezes app (Jul 2024)
- #641 — Mac window freezes (Dec 2022)

This PR supersedes #1181 and overlaps with #1182. The key difference from #1182 is the addition of the autoreleasepool safety net in `set_style_mask`, which neither existing PR includes.

## Test plan

- [x] Build a Tauri v2 app with `decorations: false`, `resizable: true` on macOS 15 Sequoia
- [x] Verify idle CPU stays near 0%
- [x] Verify window resizing still works
- [x] Verify `is_maximized()` / `is_zoomed()` calls do not cause freezes or resize loops

### DIFF
--- a/src/platform_impl/macos/util/async.rs
+++ b/src/platform_impl/macos/util/async.rs
@@ -49,10 +49,17 @@ fn run_on_main<F: Send + FnOnce()>(f: F) {
 }
 
 unsafe fn set_style_mask(ns_window: &NSWindow, ns_view: &NSView, mask: NSWindowStyleMask) {
-  ns_window.setStyleMask(mask);
-  // If we don't do this, key handling will break
-  // (at least until the window is clicked again/etc.)
-  ns_window.makeFirstResponder(Some(ns_view));
+  // Wrap in an autorelease pool so temporary objects created by
+  // setStyleMask: are drained immediately. On macOS 15 Sequoia,
+  // NSThemeFrame can create _NSThemeZoomWidget objects during layout
+  // of borderless+resizable windows; draining here prevents unbounded
+  // accumulation.
+  autoreleasepool(|_| {
+    ns_window.setStyleMask(mask);
+    // If we don't do this, key handling will break
+    // (at least until the window is clicked again/etc.)
+    ns_window.makeFirstResponder(Some(ns_view));
+  });
 }
 
 // Always use this function instead of trying to modify `styleMask` directly!
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -10,7 +10,7 @@ use std::{
   ffi::CStr,
   os::raw::c_void,
   sync::{
-    atomic::{AtomicBool, AtomicPtr, Ordering},
+    atomic::{AtomicBool, Ordering},
     Arc, Mutex, Weak,
   },
 };
@@ -213,11 +213,6 @@ fn create_window(
       }
     };
 
-    // Install the NSThemeFrame.newZoomButton swizzle (once) to prevent
-    // _NSThemeZoomWidget accumulation on macOS 15 Sequoia.
-    // See: https://github.com/tauri-apps/tao/issues/1191
-    Lazy::force(&SWIZZLE_THEME_FRAME);
-
     let mut masks = if !attrs.decorations && screen.is_none() || pl_attrs.titlebar_hidden {
       // Resizable UnownedWindow without a titlebar or borders
       // if decorations is set to false, ignore pl_attrs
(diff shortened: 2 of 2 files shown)


afonsojramos:
While investigating this issue independently (#1196), I found that the guard infrastructure to prevent the feedback loop already exists but was never wired up:

  - `window_delegate.rs` defines `is_checking_zoomed_in`: bool on WindowDelegateState (line 59)
  - `markIsCheckingZoomedIn` / `clearIsCheckingZoomedIn` ObjC methods are registered on the delegate class (lines 175-181)
  - `windowDidResize`: already checks this flag and suppresses events when true (line 349)
  - But is_zoomed() never calls these methods before toggling the style mask

So the original author clearly intended to suppress the spurious events, they just never connected the last mile.

That said, both this PR and #1182 are better long-term fixes since they eliminate the `setStyleMask` toggle entirely rather than just suppressing its side effects. The frame comparison approach in #1182 preserves correctness for borderless windows (where isZoomed() can return wrong results), while this PR's `autoreleasepool` addition is a good safety net for other set_style_mask callers.

One additional finding: `windowDidMove`: (line 358) does not check `is_checking_zoomed_in`, even though `windowDidResize` does. If the guard approach is ever used elsewhere, that gap should be closed too.

Reproduction branches for testing: [before](https://github.com/afonsojramos/tao/tree/repro/is-maximized-cpu-spike-before)/ [after](https://github.com/afonsojramos/tao/tree/repro/is-maximized-cpu-spike-after) (`cargo run --example maximize_check` on a borderless window calling `is_maximized()` in the resize handler).

Of the three approaches, #1182's frame comparison seems like the strongest, since it fully eliminates the `setStyleMask` toggle (fixing both the feedback loop and the Sequoia widget leak) while preserving correct maximized-state detection for borderless windows.

FabianLars:
> I found that the guard infrastructure to prevent the feedback loop already exists but was never wired up:

It was until it was accidentally removed in https://github.com/tauri-apps/tao/pull/486/changes because we didn't have a macOS reviewer at that time. 

I'm really not a fan of any of the current approaches but for the time being i agree that #1182 is the least bad for now.

PR #1196, opened 2026-03-19
fix(macos): wire up is_checking_zoomed_in guard in is_zoomed()

afonsojramos:
## Summary

- **Wire up the existing `is_checking_zoomed_in` guard** that was designed to suppress spurious `windowDidResize:`/`windowDidMove:` events during `is_zoomed()`'s temporary style mask toggle, but was never connected
- **Add RAII `ZoomCheckGuard`** struct to ensure the flag is always cleared, even on panic/unwind
- **Guard `windowDidMove:`** with the same `is_checking_zoomed_in` check that `windowDidResize:` already has
- **Add `debug_assert!`** for main-thread-only access in `is_zoomed()`

## Problem

Calling `is_maximized()` inside a window event listener on macOS causes 100% CPU and memory leaks. `is_zoomed()` temporarily toggles the window's style mask to `Titled | Resizable` (because `NSWindow.isZoomed()` doesn't work on borderless windows), which fires `windowDidResize:` on the delegate. If user code calls `is_maximized()` in the resize handler, it re-enters `is_zoomed()` — creating an infinite feedback loop.

On macOS 15 Sequoia, this also triggers `NSThemeFrame.newZoomButton` to create `_NSThemeZoomWidget` objects on every layout pass, compounding the CPU issue with memory leaks.

## Root Cause

The guard infrastructure already existed but was never wired up:
- `window_delegate.rs` defines `is_checking_zoomed_in: bool` and `markIsCheckingZoomedIn`/`clearIsCheckingZoomedIn` methods
- `windowDidResize:` already checks this flag and suppresses events when `true`
- **But `is_zoomed()` in `window.rs` never called these methods** before toggling the style mask

## Testing

Reproduction branches with a `maximize_check` example (borderless window calling `is_maximized()` in resize handler):

| Branch | What it shows |
|--------|---------------|
| [`repro/is-maximized-cpu-spike-before`](https://github.com/afonsojramos/tao/tree/repro/is-maximized-cpu-spike-before) | **Bug**: CPU spikes to 100% on resize |
| [`repro/is-maximized-cpu-spike-after`](https://github.com/afonsojramos/tao/tree/repro/is-maximized-cpu-spike-after) | **Fixed**: CPU stays idle on resize |

```bash
# Before fix — CPU spikes to 100%
git checkout repro/is-maximized-cpu-spike-before
cargo run --example maximize_check

# After fix — CPU stays idle
git checkout repro/is-maximized-cpu-spike-after
cargo run --example maximize_check
```

All existing tests pass (`cargo test --all`).

## Related Issues

- https://github.com/tauri-apps/tauri/issues/13199 — isMaximized() causes infinite loop on macOS
- https://github.com/tauri-apps/tauri/issues/6577 — CSS performance on macOS
- https://github.com/hoppscotch/hoppscotch/issues/3990 — 200% CPU usage on macOS when idle
- https://github.com/tauri-apps/tauri/issues/6754 — Window resize causes memory spike and crash

### DIFF
--- a/src/platform_impl/macos/window.rs
+++ b/src/platform_impl/macos/window.rs
@@ -499,6 +499,44 @@ impl From<WindowAttributes> for SharedState {
   }
 }
 
+/// RAII guard that sets `is_checking_zoomed_in` on the window delegate while
+/// the style mask is being temporarily toggled in `is_zoomed()`. This suppresses
+/// spurious `windowDidResize:`/`windowDidMove:` events that would otherwise
+/// create an infinite feedback loop when user code calls `is_maximized()` inside
+/// an event handler. The flag is always cleared on drop, even during unwind.
+struct ZoomCheckGuard {
+  delegate: id,
+}
+
+impl ZoomCheckGuard {
+  fn new(ns_window: &NSWindow) -> Self {
+    // SAFETY: The delegate is retained by ns_window for the duration of this
+    // synchronous, main-thread-only call. The guard does not outlive is_zoomed(),
+    // and the window (and thus its delegate) is kept alive by the caller's
+    // Retained<NSWindow>. The delegate pointer is !Send/!Sync by default (raw
+    // pointer), which correctly prevents use from other threads.
+    unsafe {
+      let delegate: id = msg_send![ns_window, delegate];
+      if !delegate.is_null() {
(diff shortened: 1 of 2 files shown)


afonsojramos:
Closing in favor of #1182

PR #1202, opened 2026-03-29
fix(windows): paint window background_color in WM_PAINT; add borderless_fixed example

tuyangJs:
### What this PR does

- Add `borderless_fixed` example: 500×600, undecorated, non-resizable, drag to move (with gray `background_color`).
- Windows: paint `background_color` in `WM_PAINT` so the solid color shows reliably on **undecorated** windows (`WM_ERASEBKGND` alone was often insufficient).

### Motivation

Undecorated windows frequently did not show `with_background_color` / `set_background_color` as users expect; handling the fill in `WM_PAINT` matches common Win32 practice for custom client backgrounds.
QQ20260329-202331

### Test plan

- [ ] `cargo test` (full suite)
- [ ] `cargo run --example borderless_fixed` on Windows — gray client area, drag-to-move, close
- [ ] (optional) Decorated window with `with_background_color` still correct

### Checklist

- [x] Change file added under `.changes/` (patch)
- [ ] Commits signed (per [signing commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits))

Made-with: Cursor

### DIFF
--- a/.changes/fix-windows-background-color-undecorated.md
+++ b/.changes/fix-windows-background-color-undecorated.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fix `WindowBuilder::with_background_color` on Windows for undecorated windows by painting the configured color during `WM_PAINT` (in addition to `WM_ERASEBKGND`). Add `borderless_fixed` example (500×600, undecorated, non-resizable, gray background, drag to move).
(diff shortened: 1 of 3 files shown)


Legend-Master:
Seems like this is because the subclassing happened too late that we missed the first `WM_ERASEBKGND` message, and we should look into getting #140 re-landed to fix this

I'll close this one for now since this is not the approach how we want this to be fixed, thanks anyways!

PR #1204, opened 2026-04-18
fix(linux): don't unwrap None in CursorIgnoreEvents handler

### LINKED ISSUE #1178: Linux Wayland: EventLoop::new_gtk panics on unwrap() of None value
**Describe the bug**
The application crashes immediately on startup with a panic in `tao`'s event loop initialization on Linux with Wayland. The crash occurs at line 449 in `event_loop.rs` where `Option::unwrap()` is called on a `None` value during GTK event loop creation.

**Steps To Reproduce**
1. Create a minimal Tauri/tao application on Ubuntu with Wayland
2. Run the application
3. Application crashes immediately during startup

**Expected behavior**
The application should initialize the GTK event loop successfully and start normally without crashing.

**Screenshots**
N/A - This is a crash during startup

OS: Ubuntu (Wayland session)
Rustc: 1.88.0 (from stack trace)
tao: 0.34.3
GTK: 3.x

**Additional context**
The crash occurs specifically in `tao::platform_impl::platform::event_loop::EventLoop<T>::new_gtk` at line 449 where `unwrap()` is called on a `None` value. This appears to be related to GTK initialization on Wayland systems. The application builds successfully but fails at runtime during event loop creation.

Stack trace:
```
thread 'main' panicked at /home/youssef/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tao-0.34.3/src/platform_impl/linux/event_loop.rs:449:18:
called `Option::unwrap()` on a `None` value
```

The crash happens in the GTK main context dispatch during event loop initialization, suggesting an issue with GTK/Wayland compatibility in the current tao version.

JGSimi:
## Which issue(s) this PR fixes:

Closes #1178

## What kind of change does this PR introduce?

Bug fix — replaces a bare `.unwrap()` that crosses a non-unwinding GTK main-loop boundary and aborts the whole process.

## Scenario

Creating a window with `transparent(true)` and `visible(false)` and later dispatching `CursorIgnoreEvents(true)` on it (which Tauri/wry does during setup for hidden popover-style windows) panics at `src/platform_impl/linux/event_loop.rs:448`:

```
thread 'main' panicked at tao-0.34.8/src/platform_impl/linux/event_loop.rs:448:18:
called `Option::unwrap()` on a `None` value
panic in a function that cannot unwind
```

`gtk::Window::window()` returns `None` until the underlying `GdkWindow` is realized. When the request is processed before realize, the `.unwrap()` blows up, and because the call lives inside the GTK main-context channel dispatcher (a non-unwinding C boundary), the panic becomes a hard abort and takes the whole event loop down. Users of downstream apps (Tauri v2 on modern Linux distros: Fedora 43+/Wayland, Arch, etc.) see the window flash white and die.

Reported repeatedly on Wayland — #1178 and various Tauri issues.

## How was the bug fixed?

`if let Some(gdk_window) = window.window()` instead of the bare unwrap. If the GdkWindow isn't available yet, skip — the event loop keeps running and the caller can re-send `CursorIgnoreEvents(true)` after the window is mapped (this is idempotent). The behaviour matches the `else` branch right below, which already operates on `gtk::Window` directly without touching the GdkWindow.

No behaviour change for callers that wait for realize before calling `set_ignore_cursor_events`. Callers that don't previously crashed the app — now they silently no-op until realize.

## Reproduction

Minimal Tauri v2 app on Fedora 43 / Wayland / aarch64 with a transparent, hidden window created via `WebviewWindowBuilder::new(...).transparent(true).visible(false).build()`. Same setup used by the Hat-Cross project where this was caught.

## Checklist

- [x] Ran cargo fmt locally (pure replacement of unwrap with if-let, no formatting drift)
- [x] Added a comment explaining the why
- [x] Tested locally — downstream app that used to panic on startup now runs cleanly on Fedora 43 aarch64 / Wayland with the fork applied via `[patch.crates-io]`

### DIFF
--- a/src/platform_impl/linux/event_loop.rs
+++ b/src/platform_impl/linux/event_loop.rs
@@ -442,11 +442,15 @@ impl<T: 'static> EventLoop<T> {
           }
           WindowRequest::CursorIgnoreEvents(ignore) => {
             if ignore {
-              let empty_region = Region::create_rectangle(&RectangleInt::new(0, 0, 1, 1));
-              window
-                .window()
-                .unwrap()
-                .input_shape_combine_region(&empty_region, 0, 0);
+              // `window.window()` is None until the GdkWindow is realized.
+              // Tauri/wry can send CursorIgnoreEvents before a transparent
+              // or initially-hidden window is mapped, and the bare unwrap
+              // aborts the whole GTK main loop with a non-unwinding panic.
+              // Silently skipping keeps the event loop alive.
+              if let Some(gdk_window) = window.window() {
+                let empty_region = Region::create_rectangle(&RectangleInt::new(0, 0, 1, 1));
+                gdk_window.input_shape_combine_region(&empty_region, 0, 0);
+              }
             } else {
               window.input_shape_combine_region(None)
             };


src/platform_impl/linux/event_loop.rs
```diff
@@ -451,11 +451,15 @@ impl<T: 'static> EventLoop<T> {
...
-                .input_shape_combine_region(&empty_region, 0, 0);
+              // `window.window()` is None until the GdkWindow is realized.
+              // Tauri/wry can send CursorIgnoreEvents before a transparent
+              // or initially-hidden window is mapped, and the bare unwrap
+              // aborts the whole GTK main loop with a non-unwinding panic.
+              // Silently skipping keeps the event loop alive.
+              if let Some(gdk_window) = window.window() {
+                let empty_region = Region::create_rectangle(&RectangleInt::new(0, 0, 1, 1));
+                gdk_window.input_shape_combine_region(&empty_region, 0, 0);
+              }
```
sftse, line 462: I think this would be equivalent to calling
```rust
window.input_shape_combine_region(Some(&empty_region));
```
directly on the `gtk::Window`.


Is the previous code here even correct, `empty_region.is_empty()` does not hold, probably
```rust
let empty_region = Region::create();
```
was intended.

---------------------
Is `wry` guaranteed to reschedule the missed event? Seems a bit non-local to have to reason about this. Is there a way to know the setup has completed before passing these events?

sftse (changes requested):

mogorman:
this patch resolved my crash on nixos. thanks.

PR #1207, opened 2026-04-27
fix(macos): keep webview synced during window zoom

Tunglies:
Related issues:
https://github.com/tauri-apps/tauri/issues/13898
https://github.com/tauri-apps/tauri/issues/14288
https://github.com/tauri-apps/tauri/issues/15092

Possible related issues(not tested):
https://github.com/tauri-apps/tauri/issues/14843
https://github.com/tauri-apps/tauri/issues/5226

---

  The `ns_window.zoom(None)` path is used for normal macOS windows because Tao previously called it
  whenever the window was both `Resizable` and `Titled`. That means most regular Tauri windows used
  AppKit's native smart-zoom implementation for `set_maximized`.

  The problem is that `NSWindow::zoom` performs an internal AppKit zoom animation instead of resizing the
  window through Tao's own frame path. During that animation, the `NSWindow` frame can advance before the
  embedded `WKWebView`'s native bounds and JavaScript viewport are updated. As a result, WebView content
  temporarily stays at the old size while the window has already grown, which shows up as white/blank
  borders or delayed content resizing during maximize/restore.

  The same issue can also be triggered by titlebar double-click, because AppKit invokes the same native
  zoom behavior there. Therefore the fix avoids `ns_window.zoom(None)` for maximize and intercepts
  `windowShouldZoom:toFrame:` so titlebar zoom also goes through Tao's controlled maximize path.

---

## Basline

https://github.com/user-attachments/assets/5b34f564-8453-4f47-9f43-a72cb3043468

## With PR

https://github.com/user-attachments/assets/8d338d77-6070-47d6-aad5-b93980b6a4d5


src/platform_impl/macos/util/async.rs
```diff
@@ -172,25 +194,16 @@ pub unsafe fn set_maximized_async(
...
 
-      let curr_mask = ns_window.styleMask();
       if shared_state_lock.fullscreen.is_some() {
         // Handle it in window_did_exit_fullscreen
         return;
-      } else if curr_mask.contains(NSWindowStyleMask::Resizable)
-        && curr_mask.contains(NSWindowStyleMask::Titled)
-      {
-        // Just use the native zoom if resizable
-        ns_window.zoom(None);
```
Tunglies, line 183: This may be more than a patch-level change. Fixing the animation behavior requires removing the use of
  `ns_window.zoom(None)` so AppKit no longer takes over the zoom animation path.

Tunglies:
Hey @amrbashir , would you like to review this PR? :)

PR #1209, opened 2026-04-30
refactor(mobile)!: move Resumed and Suspended to WindowEvent

lucasfernog:
follow up for https://github.com/tauri-apps/tauri/pull/15199/


src/platform_impl/android/ndk_glue.rs
```diff
@@ -65,10 +62,10 @@ const DATA_URL_ENCODING_SET: &AsciiSet = &CONTROLS
 /// 2. android package anme (for ex: wryapp)
 /// 3. the android activity that has external linking for the following functions and calls them:
 ///       - `private external fun onActivityCreate(activity: WryActivity)``
-///       - `private external fun start()`
-///       - `private external fun resume()`
-///       - `private external fun pause()`
-///       - `private external fun stop()`
+///       - `private external fun start(activity: WryActivity)`
```
Legend-Master, line 65: The naming here is honestly so confusing, some are prefixed with `onActivity` while the others are not.

Legend-Master (approved):
Android part looks good to me

FabianLars (approved):
i think we can support started and stopped on ios (with scenes at least) as well but that's something to investigate in another pr

ios side looks good (though i can't actually test it rn)

PR #1211, opened 2026-05-03
fix(android): don't panic on getCurrentWindowMetrics on API<30

SAY-5:
Addresses the Android 10 startup panic reported in DioxusLabs/dioxus#5149.

`MonitorHandle::size()` JNI-calls `getCurrentWindowMetrics`, which only exists on Android R (API 30+). On older devices it raises `NoSuchMethodError` and the chained `.unwrap()`s turn the JavaException into a hard process panic at startup. Catch the call failures, clear any pending JNI exception so the next JNI call doesn't blow up, and fall back to `PhysicalSize(0,0)` — the same fallback the function already uses when `window_manager` or the Android context is missing.

### DIFF
--- a/.changes/change-pr-1211.md
+++ b/.changes/change-pr-1211.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+fix(android): don't panic on getCurrentWindowMetrics on API<30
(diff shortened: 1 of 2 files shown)


SAY-5:
Added .changes/change-pr-1211.md in 80b9812 per the covector bot.

lucasfernog:
thanks, i'll release this one today

lucasfernog (approved):

lucasfernog (approved):

PR #1215, opened 2026-05-05
fix(windows): avoid reentrant input lock deadlocks

### LINKED ISSUE #1194: Windows - RDP Unresponsive
**Describe the bug**
The application works fine, but after:

1. opening an RDP session on Windows Server,
2. launching the application,
3. closing the session or minimizing the RDP client,
4. reopening/restoring the session after several seconds,
5. the application window then appears as "not responding".

**Steps To Reproduce**
Example tested: examples/minimize.rs https://github.com/tauri-apps/tao/blob/dev/examples/minimize.rs
Dependency version: tao = 0.34.6
cargo build --target=i686-pc-windows-msvc  --release

**Screenshots**
Image

**Platform and Versions**
OS: Microsoft Windows Server 2016 Standard - 14393
Rustc: rustc 1.93.1 (01f6ddf75 2026-02-11)

**Additional context**
Hello, and thank you to the authors and maintainers of tao/tauri for your excellent work .

After debugging, I discovered the freeze originates from a deadlock in the keyboard_callback function in src/platform_impl/windows/event_loop.rs.

I asked GitHub Copilot to analyze what could be blocking at that level; here's the analysis:
When Windows closes a session, WM_KILLFOCUS is sent to the window.
With keyboard_callback active, process_message calls synthesize_kbd_state(Released) which generates a KeyboardInput(Released) event for each physically pressed key.
Each of these events is dispatched via send_event to the user handler synchronously.
During Windows session shutdown/minimize, certain Win32 operations triggered in this handler (SetWindowPos, internal SendMessageW calls, etc.) can block because Windows is terminating the session process — hence the "not responding" state.

By adding this guard clause after the if !is_keyboard_related block in keyboard_callback, the freeze no longer occurs:
`if !is_keyboard_related {
  // We return early to avoid a deadlock from locking the window state
  // when not appropriate.
  return;
}

if msg == win32wm::WM_SETFOCUS || msg == win32wm::WM_KILLFOCUS {
  return;
}`

Is an explicit filter of WM_SETFOCUS / WM_KILLFOCUS at this level the right approach in Tao ?

Thank you :)

### LINKED ISSUE #12531: [bug] App hanging on startup when switching to other windows
### Describe the bug

Tauri application stops responding when switching to or opening other windows while it is starting up. I have experienced this with my own project opening multiple other applications at the same time, but I have tested and this can happen on a default Tauri + Vite project as well, created following this guide: [https://v2.tauri.app/start/create-project/](https://v2.tauri.app/start/create-project/) . In my experience it never seems to recover from this state on its own.

### Reproduction

This can happen by opening the tauri app alongside some other applications at the same time. But I've found it can be induced more often by spamming alt+tab quickly and repeatedly while the app is launching, though it does not happen every time. Video demonstration attached. This seems more difficult to reproduce running in dev, probably because of the less predictable startup timing.

https://github.com/user-attachments/assets/ec06d4fd-96c4-4958-9977-643e4759c26f

### Expected behavior

I would expect the application to be able to launch normally

### Full `tauri info` output

```text
> testing@0.1.0 tauri
> tauri info


[✔] Environment
    - OS: Windows 10.0.19045 x86_64 (X64)
    ✔ WebView2: 132.0.2957.127
    ✔ MSVC: 
        - Visual Studio Community 2019
        - Visual Studio Community 2022
    ✔ rustc: 1.82.0 (f6e511eec 2024-10-15)
    ✔ cargo: 1.82.0 (8f40fc59f 2024-08-21)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.9.0
    - npm: 10.1.0

[-] Packages
    - tauri 🦀: 2.2.5
    - tauri-build 🦀: 2.0.5   
    - wry 🦀: 0.48.1
    - tao 🦀: 0.31.1
    - @tauri-apps/api : 2.2.0
    - @tauri-apps/cli : 2.2.7

[-] Plugins
    - tauri-plugin-opener 🦀: 2.2.5     
    - @tauri-apps/plugin-opener : 2.2.5

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

### Stack trace

```text
None
```

### Additional context

_No response_

chuwik:
## Summary

This PR fixes a Windows/RDP freeze where a Tao window could become unresponsive after disconnecting, minimizing, or restoring an RDP session.

The deadlock was caused by keyboard/IME message handling calling `PeekMessageW` while input-state or keyboard-layout locks were held. `PeekMessageW` can synchronously re-enter the window procedure; if the re-entered path tries to acquire the same non-reentrant lock, the UI thread can freeze.

I validated myself, and with several users, that the RDP unresponsive scenario that motivated this patch no longer reproduces.

## What changed

The fix moves the message peeks out of the locked sections, then passes the already-peeked state into the keyboard/IME handlers.

In practice:

- keyboard message peeking now happens before keyboard event construction
- IME "more chars pending?" detection now happens before locking `window_state`
- keyboard layout cache locks are held for shorter scopes
- current key state is read without holding the layout cache lock

No public API changes, dependency changes, or non-Windows behavior changes are intended.

## Validation

- RDP disconnect/minimize/restore scenario no longer reproduces (validated with several users)
- `cargo fmt --all -- --check`
- `cargo test --all`

Fix https://github.com/tauri-apps/tauri/issues/12531

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -867,6 +867,14 @@ fn update_modifiers<T>(window: HWND, subclass_input: &SubclassInput<T>) -> Modif
   modifiers
 }
 
+/// WARNING: Due to using PeekMessage, the event handler
+/// function may get called during this function.
+/// (Re-entrance to the event handler)
+///
+/// This can cause a deadlock if calling this function
+/// while having a mutex locked.
+///
+/// It can also cause code to get executed in a surprising order.
 fn peek_next_key_message(window: HWND) -> Option<MSG> {
   unsafe {
     let mut next_msg = mem::MaybeUninit::uninit();
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -55,7 +55,7 @@ use crate::{
     dark_mode::try_window_theme,
     dpi::{become_dpi_aware, dpi_to_scale_factor, enable_non_client_dpi_scaling},
     keyboard::is_msg_keyboard_related,
-    keyboard_layout::LAYOUT_CACHE,
+    keyboard_layout::get_agnostic_mods,
     minimal_ime::is_msg_ime_related,
     monitor::{self, MonitorHandle},
     raw_input, util,
@@ -849,7 +849,7 @@ unsafe fn process_control_flow<T: 'static>(runner: &EventLoopRunner<T>) {
 fn update_modifiers<T>(window: HWND, subclass_input: &SubclassInput<T>) -> ModifiersState {
   use crate::event::WindowEvent::ModifiersChanged;
 
-  let modifiers = LAYOUT_CACHE.lock().get_agnostic_mods();
+  let modifiers = get_agnostic_mods();
   let mut window_state = subclass_input.window_state.lock();
   if window_state.modifiers_state != modifiers {
     window_state.modifiers_state = modifiers;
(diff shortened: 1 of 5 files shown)


Legend-Master:
Would it be easier for us to just copy in the new one from winit? That being said, I don't see where the `MinimalIme` is used in the new winit code anymore

> https://github.com/rust-windowing/winit/blob/master/winit-win32/src/minimal_ime.rs

chuwik:
> Would it be easier for us to just copy in the new one from winit? That being said, I don't see where the `MinimalIme` is used in the new winit code anymore
> 
> > https://github.com/rust-windowing/winit/blob/master/winit-win32/src/minimal_ime.rs

Yeah I think we _could_ pull over the one from winit, but that feels like a larger/riskier change for this PR. The current patch is fairly minimal and has already been tested, so my preference would be to merge this as the smaller safe fix first.

Since `MinimalIme` also doesn’t appear to be used in the newer code anymore, I think reworking that would probably deserve a separate follow-up PR where we can evaluate and test the behavior more carefully.

Legend-Master:
Just from looking at it, all the differences from the winit one is that it uses atomic values in MinimalIme instead of locking the entire MinimalIme

src/platform_impl/windows/event_loop.rs
```diff
@@ -867,6 +867,59 @@ fn update_modifiers<T>(window: HWND, subclass_input: &SubclassInput<T>) -> Modif
...
+    .as_bool()
+    {
+      Some(next_msg.assume_init())
+    } else {
+      None
+    }
+  }
+}
+
+fn next_key_message_for_keyboard(window: HWND, msg: u32, wparam: WPARAM) -> Option<MSG> {
```
Legend-Master, line 897: I still think ideally should copy in the winit related code which is not that much, just a few drops before calling `PeekMessageW` and moving the locks to inside `MinimalIme`

We want to switch back to winit one day so it would be nice to keep the changes small whenever possible

Legend-Master (commented):
I get the desire of landing this already tested code here, but we have very limited maintainer time here that this would be harder to maintain as it diverges from the winit code that isn't strictly necessary

Legend-Master (approved):
> Since `MinimalIme` also doesn’t appear to be used in the newer code anymore, I think reworking that would probably deserve a separate follow-up PR where we can evaluate and test the behavior more carefully.

Changed my mind, agree with this, I will get the winit change in through another PR

Thanks! Awesome work!

PR #1218, opened 2026-05-07
fix(wayland): fix client-side decorations and apply them only when necessary

### LINKED ISSUE #899: WindowBuilder::with_decorations(false) doesn't work on wayland kde.
**Describe the bug**
decoration set false doesn't work on Wayland KDE.
When running the example "decorations" (example/decorations.rs), press "space", the window decoration shluld disappear, then press "space" again, the window decoration should appear. But it only works on X11 or wayland gnome.
When using KDE Wayland, the window decoration (including titlebar) always shown, even press "space" again and again.
**Steps To Reproduce**
1. clone tao repo.
2. cd tao directory.
3. 
```bash
GDK_BACKEND=wayland cargo run --example decorations
```
set "GDK_BACKEND=wayland" environment variable and run decorations example
then press "space" key.

**Expected behavior**
When press "space" key, the window decorations (including the titlebar) disappear. Press the "space" key again the window decorations appear.

**Screenshots**

OS: Archlinux x86_64 KDE plasma 6
Rustc: 
1.77.1 (7cf61ebde 2024-03-27)
**Additional context**

#https://github.com/tauri-apps/tauri/issues/6562
#https://gitlab.gnome.org/GNOME/gtk/-/issues/5479

### LINKED ISSUE #1046: [request] Remove client decorations (linux/wayland)
This is mostly about https://github.com/tauri-apps/tao/pull/979 / https://github.com/tauri-apps/tao/issues/977 https://github.com/tauri-apps/tauri/issues/10702

From what i've heard so far the change caused more harm then good because now tao/tauri windows behave very non-wayland like.

First of all, the Titlebar buttons shouldn't have been a reason for this PR. If wayland doesn't show min/max buttons by default that's not our problem and if users want those buttons they should use the system settings to get them back.

Then the the more important problem, the 71 error code / resizing in general.

On wayland it's a very well known issue that app windows don't restore their state, _but_ this is not our problem to fix if wayland doesn't _want_ us to be able to do this.
I can't remember about Gnome, but KDE for example is working on this https://bugs.kde.org/show_bug.cgi?id=436318 etc

imho the better solution would have been to a) document wayland specific behavior as much as we can and b) : 
- Make set_size methods fail non-panicking if they had issues (didn't find mentions of this)
- webkitgtk internal errors mentioned in #10702 either shouldn't be our concern, or we should finally add the mentioned workarounds (env vars) that also fix other issues.
- moving and resizing with the native decorations not working should really have some other fix. i can't believe that this requires client decorations because this used to work fine without them for such a long time before.

Extra refs of issues people face after #979 (will add more over time - edit: that was a lie, but check out the issues linked by github below)
https://discord.com/channels/616186924390023171/1328842890637607044
https://discord.com/channels/616186924390023171/1305624260479418451

### LINKED ISSUE #1243: Linux X11: Window cannot be resized/moved when GTK_CSD=1 is set
**Describe the bug**
When I force GTK applications to run with client-side decoration with the `GTK_CSD=1` environment variable, Tauri apps cannot be resized/moved by the pointer device on X11.

**Steps To Reproduce**
Login into an X11 session,  export `GTK_CSD=1` environment variable, then run any Tauri apps that use system window decorations, and try to move the window by dragging the title bar or resize the window by dragging its border.

**Expected behavior**
Resizing the window and dragging the window by the title bar should work the same way  when client-side decoration is disabled with `GTK_CSD=0`.

OS: Arch Linux
Rustc: 1.96.0

**Additional context**
It happens only on X11. The resize/move operation with client-side decoration on Wayland works as expected.

### LINKED ISSUE #13440: [bug] native window bar's buttons not working (on wayland)
### Describe the bug

Hi, when the app is running on some Linux distros the close, minimize or maximize buttons in native window bar, are unresponsive. Clicking or hovering them does nothing. User has to double-click the window bar then somehow they start working as expected, but it's clunky and not very reliable.
Note: it seems it only happens on Wayland. I tested the same app on a bunch of distros, and only
Ubuntu24 (wayland) seems to be impacted. Eg, debian12 XFCE or PopOs22 which are both on X11 don't have the issue.

No error in logs (at least tauri app process).
The window config is declared in Rust, nothing fancy:

```
let window = WebviewWindowBuilder::new(app, "main", "index.html")
        .title("my_app")
        .fullscreen(false)
        .resizable(true)
        .closable(true)
        .minimizable(true)
        .maximizable(true)
        .center()
        .focused(true)
        .inner_size(328.0, 710.0)
        .build()
        .inspect_err(|e| error!("failed to create main window: {e}"))?;
```

It seems this has been previously reported but not fixed: https://github.com/tauri-apps/tauri/issues/11631
Could be related to https://github.com/tauri-apps/tauri/issues/11945

### Reproduction

create a simple app using the window config shared above and testing the window buttons after the window open

testing on Ubuntu24 default flavor (wayland)

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Arch Linux Unknown x86_64 (X64) (xmonad on X11)
    ✔ webkit2gtk-4.1: 2.48.1
    ✔ rsvg2: 2.60.0
    ✔ rustc: 1.86.0 (05f9846f8 2025-03-31)
    ✔ cargo: 1.86.0 (adf9b6ad1 2025-02-28)
    ✔ rustup: 1.28.1 (f9edccde0 2025-03-05)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (default)
    - node: 22.13.1
    - npm: 10.9.2
    - bun: 1.1.43

[-] Packages
    - tauri 🦀: 2.5.1
    - tauri-build 🦀: 2.2.0
    - wry 🦀: 0.51.2
    - tao 🦀: 0.33.0
    - tauri-cli 🦀: 2.3.1 (outdated, latest: 2.5.0)
    - @tauri-apps/api : 2.5.0
    - @tauri-apps/cli : 2.5.0

[-] Plugins
    - tauri-plugin-window-state 🦀: 2.2.2
    - @tauri-apps/plugin-window-state : 2.2.2
    - tauri-plugin-notification 🦀: 2.2.2
    - @tauri-apps/plugin-notification : 2.2.2
    - tauri-plugin-autostart 🦀: 2.3.0
    - @tauri-apps/plugin-autostart : 2.3.0
    - tauri-plugin-dialog 🦀: 2.2.1
    - @tauri-apps/plugin-dialog : 2.2.1
    - tauri-plugin-updater 🦀: 2.7.1
    - @tauri-apps/plugin-updater : 2.7.1
    - tauri-plugin-opener 🦀: 2.2.6
    - @tauri-apps/plugin-opener : 2.2.6
    - tauri-plugin-single-instance 🦀: 2.2.3
    - tauri-plugin-process 🦀: 2.2.1
    - @tauri-apps/plugin-process : 2.2.1
    - tauri-plugin-fs 🦀: 2.2.1
    - tauri-plugin-os 🦀: 2.2.1
    - @tauri-apps/plugin-os : 2.2.1
    - tauri-plugin-clipboard-manager 🦀: 2.2.2
    - @tauri-apps/plugin-clipboard-manager : 2.2.2

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

dgerhardt:
This PR fixes multiple issues regarding window decoration handling on Wayland:

1. The default client side decoration handling of GTK is now used once again. Custom CSD don't make any sense when no custom widgets are added to the title bar. Futhermore, the custom CSD didn't work correctly and introduced multiple bugs regarding the titlebar and server-side rendering on Wayland.

    This is mostly a revert of #979.

    Fixes #1046, tauri-apps/tauri#13749, tauri-apps/tauri#14251, tauri-apps/tauri#14748  
    Supersedes #1055
2. Window mouse events are once again propagated.

    This reverts #941, which broke CSD event handling on Wayland. A new fix for #939 might now be needed.
    
    Fixes tauri-apps/tauri#13440
    Might fix (unconfirmed) tauri-apps/tauri#11856
4. A workaround for a GTK bug has been added so compositors don't force decorations. This enables CSD for undecorated windows to ensure that compositors with server-side decoration support (e.g. KDE Plasma) do not apply them when decorations are disabled for the window. Changing the decoration state for existing windows is however still not supported.
    
      This is a workaround for a GTK bug: https://gitlab.gnome.org/GNOME/gtk/-/work_items/5479
    
      Fixes #899, tauri-apps/tauri#6562

While these are individual fixes, they depend on each other, so I submit them in a single PR. With the latter changes, regressions, which only reverting #979 would cause, can be avoided.


FabianLars:
Thanks for the PR! Really happy to see this progressing. There's one major problem though, your PR now re-introduces the startup crash you can see in #979 PR description/video. Simply add `.with_maximized(true)` to the resizable example and it should crash straight away. At least it does in my WSL instance.

dgerhardt:
Thanks for the quick feedback. While I wasn't able to reproduce the crash on my end (tested on Ubuntu 26.04, GNOME and KDE), the window did no longer correctly maximize in this situation for me.

I've added a commit to reapply the fix regarding maximize handling from #941. It fixes maximizing with disabled resizing for me and hopefully also the crash.

FabianLars:
Still crashes for me but the question would be how WSL-specific that is. If it doesn't happen on normal Linux systems i could actually live with that. I'll ask on discord for testers as well.

FabianLars:
> Simply add .with_maximized(true) to the resizable example and it should crash straight away. At least it does in my WSL instance.

idk if anything changed or if my comment was wrong, but it'll crash now if i add `.with_maximized(true).with_decorations(false)` - i assume that's what i meant back then as well, not sure.

.changes/fix-wayland-csd.md
```diff
@@ -0,0 +1,10 @@
+---
+"tao": patch
```
FabianLars, line 2: No need to change this, just fyi: I'm aiming for a new tao minor release and a tauri minor (2.12) just so we have more of an obvious point in time where this was changed and no hidden patch updates on `cargo update` or whatever.
(The dev branch is already for the next tao minor release so nothing actually changes)

FabianLars:
Sooo, i am okay with merging it as is but ideally we fix stuff before we actually release the changes
1) The wsl crash may be fixed by applying the same workaround https://github.com/tauri-apps/tao/pull/1218/changes/098cfb363f9c268f0d906a33ac91469330302ae5 but couldn't try that yet - i wouldn't block the release on a wsl specific crash that needs resizable, maximized and !decorations
2) The duplicate press/release events: i think the simplest fix is to compare new event's timestamp with the old one but i'm not sure if this is something for tao or tauri, but probably tao

@Legend-Master any thoughts or concerns?

Legend-Master:
It seems like the minimize button and maximize buttons are now gone after this?

Before

image

After

image

FabianLars:
yes and as stupid as it sounds, this is actually correct. stupid gnome 😂 you should see the same button "issue" in all native gnome apps unless you use Gnome tweaks to change that.

I'm a bit torn about this as well though, I always hated this and don't know anyone who doesn't jump straight into gnome tweaks to fix it, but i think it's a tradeoff well worth to have working titlebars again.
If we have more resources for Linux we could take another look at our own CSD but probably makes sense to wait for the gtk4 migration

Legend-Master:
Yeah I see... 'fixed' it by `gsettings set org.gnome.desktop.wm.preferences button-layout ":minimize,maximize,close"` 😂

Honestly, I didn't experience any problems just testing the dev branch in WSL, and this PR does break that very combination (in WSL). I won't block this though seeing how many linked issues 😂

dgerhardt:
I tested again on my Ubuntu 24.04.1 VM and can't reproduce this issue. All window buttons appear when using Gnome Desktop and a Wayland session.

This seems to be in issue with the WSL environment. Running the same version of Ubuntu, `gsettings get org.gnome.desktop.wm.preferences button-layout` returns different results on a native/VM Ubuntu (`:minimize,maximize,close`) and on WSL (`appmenu:close`). This isn't an issue with TAO but affects all GTK-based apps. When I run gnome-calculator in WSL, the buttons are also missing.

`gsettings set org.gnome.desktop.wm.preferences button-layout` allows end users to customize the button layout globally for all apps (e.g. `close,minimize,maximize:` for a macOS-like layout). When it is overridden, TAO-based apps would be inconsistent for users that changed this setting intentionally.

I would suggest to just add a note to the documentation for developers on WSL mentioning the need to run the `gsettings` command once.

FabianLars:
> All window buttons appear when using Gnome Desktop and a Wayland session.

Yeah, i pushed a change for that after Tony's comment that also works in wsl for me. Like i said, i am a bit torn about this but i believe most tauri devs would expect the buttons to be there so i added them back. I've not seen any errors caused by that yet.

I didn't merge this yet because i was trying a few things to get rid of the protocol error i mentioned but all solutions suck

dgerhardt:
> Yeah, i pushed a change for that after Tony's comment that also works in wsl for me.

I should have mentioned: I retested with and without the button layout commit. It works fine outside of WSL and on WSL all GTK-based apps have a broken button layout without changing the setting.

FabianLars:
> I tested again on my Ubuntu 24.04.1 VM and can't reproduce this issue. All window buttons appear when using Gnome Desktop and a Wayland session.

> This seems to be in issue with the WSL environment. 

More like this is one of the few things Ubuntu did right. They changed gnome's default. If you use fedora gnome for example it should behave similar-ish to wsl as it's vanilla gnome.

> When it is overridden, TAO-based apps would be inconsistent for users that changed this setting intentionally.

Which is fine imo. If gnome wants CSD we might as well do CSD you know. Also, we've had the buttons with the now removed custom titlebar so keeping it for now sounds reasonable to me. We can re-iterate on that later though if we get negative feedback.

---

> I didn't merge this yet because i was trying a few things to get rid of the protocol error i mentioned but all solutions suck

This is not wsl specific btw but at least weston specific (wsl uses weston) so i'll keep trying things but i'm very rusty with gtk so if somebody has ideas i'm all ears x)

FabianLars:
okayyy, any thoughts on the last 2 commits?

The first one to fix the crash is super duper ugly and i hate it but it's the best thing stupid me could come up with right now. I hope it holds it together until our gtk4 migration where we can set more of the required things on the window builder which hopefully doesn't even cause the issues we're dealing with here.

The second one to prevent duplicate events i think is the best approach we have - meaning the timestamp comparison, the implementation itself may be shit too, please tell me if so. Also, other events like MouseLeave are triggered twice as well but i did not see anyone ever complaining about this so i only touched the 3 events that were reverted here.

Legend-Master:
What works works I guess 😂, at least it doesn't crash in WSL now

dgerhardt:
I retested with the latest two commits on KDE Plasma. The crash fix actually also fixes maximize on startup behavior here, which was already broken on the dev branch before. Previously, the window would then often be in a weird maximized state without actually filling the full screen.

> Which is fine imo. If gnome wants CSD we might as well do CSD you know. Also, we've had the buttons with the now removed custom titlebar so keeping it for now sounds reasonable to me. We can re-iterate on that later though if we get negative feedback.

Even if I might not agree with [GNOME's design philosophy](https://afaikblog.wordpress.com/2011/03/01/where-did-the-buttons-go/) of not having minimize/maximize buttons at all, if a distro decides to go with it, I don't think there is a reason to actively work against that on the app side. But I do agree that it is reasonable to enforce the buttons for now to avoid a change of existing behavior.

FabianLars:
yeah i'm thinking of selling it as a gtk4 change once we update or something. but i'll leave that decision to whoever merges this pr

FabianLars (approved):

src/platform_impl/linux/event_loop.rs
```diff
@@ -253,6 +253,11 @@ impl<T: 'static> EventLoop<T> {
       _marker: std::marker::PhantomData,
     };
 
+    // Enable Min/Max titlebar buttons for all windows. This is an app-wide setting so it does not make sense to set it in Window::new.
+    if let Some(settings) = Settings::default() {
+      settings.set_gtk_decoration_layout(Some("menu:minimize,maximize,close"));
```
City-busz: I would prefer not to override the window decoration layout, so it would be consistent with native GTK windows and allow users to customize it to their needs.

FabianLars:
fuck it we ball - we can always add it back, users can even copy the same code into their app.

PR #1212, opened 2026-05-04
feat(android): describe exceptions

lucasfernog:
ref https://github.com/tauri-apps/tauri/issues/15311

PR #1219, opened 2026-05-11
fix(android): don't panic when VIEW intent has null MIME type

Sumberss:
Intent.getType() returns null for typical VIEW deep links (e.g. capsule://...), which causes env.get_string() to return Err. The previous .map(...unwrap()) chain panicked from the onActivityCreate JNI callback, aborting the process via SIGABRT.

Switch to and_then + .ok() to mirror the pattern already used a few lines above for the action string.

### DIFF
--- a/src/platform_impl/android/ndk_glue.rs
+++ b/src/platform_impl/android/ndk_glue.rs
@@ -505,12 +505,11 @@ pub unsafe fn handle_intent(mut env: JNIEnv, intent: JObject) {
   let intent_type = jni_call_method!(env, &intent, "getType", "()Ljava/lang/String;", l)
     .ok()
     .map(|jstr| jstr.into())
-    .map(|intent_type| {
+    .and_then(|intent_type: JString| {
       env
         .get_string(&intent_type)
-        .unwrap()
-        .to_string_lossy()
-        .to_string()
+        .ok()
+        .map(|s| s.to_string_lossy().to_string())
     });
 
   // Handle text/plain intents (EXTRA_TEXT)


FabianLars:
Thanks for the PR! Can you add a small changefile like this? <https://github.com/tauri-apps/tao/pull/1212/changes#diff-ec046f868e52182b99c3b7471600b0dfb827cc913c9162014bc967f4aa5aca82>
And more importantly, we require _all_ commits to be signed. I'd appreciate it if you could set that up quickly https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits 

Also, just for completeness sake, you wouldn't have a strack trace for this crash lying around still, right?

Sumberss:
Here is the stack trace : 

```
05-11 10:50:08.137  ActivityTaskManager: START u0 {act=android.intent.action.VIEW
                      cat=[android.intent.category.BROWSABLE] dat=capsule://auth/...
                      flg=0x14000000 cmp=com.capsule.love/.MainActivity}
                      with LAUNCH_SINGLE_TASK

  05-11 10:50:08.295  nativeloader: Load .../lib/arm64-v8a/libapp_lib.so ... ok

  05-11 10:50:08.298  RustStdoutStderr: thread '<unnamed>' (7608) panicked at
     /Users/capsule/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/
     tao-0.35.2/src/platform_impl/android/ndk_glue.rs:511:10:

  05-11 10:50:08.312  RustStdoutStderr: 12:  0x771444b268
                      - Java_com_capsule_love_Rust_onActivityCreate

  05-11 10:50:08.313  libc: Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE)
                      in tid 7608 (om.capsule.love), pid 7608 (om.capsule.love)
```

Legend-Master:
#1224 superseded this since it had the change file and commits signed, I have added you to the co-author and thanks anyways!

PR #1228, opened 2026-05-30
Fix(linux): JIS key mapping

ushinohama966:
### Description

This PR adds missing mappings for Japanese (JIS) keyboard specific keys (`Zenkaku_Hankaku`, `Hiragana_Katakana`, `Henkan`, and `Muhenkan`) in `raw_key_to_key` for the GTK backend.

Without this mapping, pressing these keys on Linux environments causes them to fall back to `Key::Unidentified`, leading to unexpected behaviors or warning logs in frontend event handlers.

### DIFF
--- a/.changes/fix-jis-key-mapping.md
+++ b/.changes/fix-jis-key-mapping.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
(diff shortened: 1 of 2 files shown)


pewsheen (approved):
Tested and looks good! Thanks!

pewsheen:
@ushinohama966 Could you follow https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits to sign your commit? otherwise we can't merge this PR.

ushinohama966:
@pewsheen Thank you for the feedback!
I have updated the commits to be GPG/SSH signed as requested. I have also verified that all commits in this PR now have the "Verified" status.

PR #1231, opened 2026-06-02
refactor(windows): re-land remove window subclassing

### LINKED ISSUE #1230: [Windows] `with_background_color` doesn't work on initial load
Seems like this is because the subclassing happened too late that we missed the first `WM_ERASEBKGND` message, and we should look into getting #140 re-landed to fix this

_Originally posted by @Legend-Master in https://github.com/tauri-apps/tao/issues/1202#issuecomment-4599515416_

Legend-Master:
Re-land of #140

The original reason for the revert in https://github.com/tauri-apps/tao/pull/156#issuecomment-884843341 was that transparent was broken, this change made sure that still works, tested with the transparent example locally

Fix #1230

The main reason for this change is to fix `with_background_color` doesn't work on initial load since the subclass is attached too late to receive the `WM_ERASEBKGND` message. It might also catch some other similar type of bugs.

This would also pull tao towards winit a bit more so when we switch back, there're lesser behavior change we need to go through.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -657,7 +657,7 @@ fn create_event_target_window<T: 'static>() -> HWND {
       // `explorer.exe` and then starting the process back up.
       // It is unclear why the bug is triggered by waiting for several hours.
       WS_EX_TOOLWINDOW,
-      PCWSTR::from_raw(thread_event_target_window_class.clone().as_ptr()),
+      thread_event_target_window_class,
       PCWSTR::null(),
       WS_OVERLAPPED,
       0,
@@ -2279,8 +2279,8 @@ unsafe extern "system" fn thread_event_target_callback<T: 'static>(
 ) -> LRESULT {
   let userdata_ptr = util::GetWindowLongPtrW(window, GWL_USERDATA) as *mut ThreadMsgTargetData<T>;
   if userdata_ptr.is_null() {
-    // `userdata_ptr` will always be null for the first `WM_GETMINMAXINFO`, as well as `WM_NCCREATE` and
-    // `WM_CREATE`.
+    // `userdata_ptr` will always be null for the first `WM_GETMINMAXINFO`, as well as
+    // `WM_NCCREATE` and `WM_CREATE`.
     return DefWindowProcW(window, msg, wparam, lparam);
   }
   let userdata = Box::from_raw(userdata_ptr);
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -36,13 +36,14 @@ use windows::{
       Controls::{self as win32c, HOVER_DEFAULT},
       Input::{KeyboardAndMouse::*, Pointer::*, Touch::*, *},
       Shell::{
-        DefSubclassProc, RemoveWindowSubclass, SHAppBarMessage, SetWindowSubclass, ABE_BOTTOM,
-        ABE_LEFT, ABE_RIGHT, ABE_TOP, ABM_GETAUTOHIDEBAR, APPBARDATA,
+        DefSubclassProc, SHAppBarMessage, ABE_BOTTOM, ABE_LEFT, ABE_RIGHT, ABE_TOP,
+        ABM_GETAUTOHIDEBAR, APPBARDATA,
       },
       WindowsAndMessaging::{self as win32wm, *},
     },
   },
 };
+use windows_core::w;
 
 use crate::{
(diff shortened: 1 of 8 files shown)


src/platform_impl/windows/window.rs
```diff
@@ -1379,62 +1432,6 @@ unsafe fn register_window_class(window_classname: &str) -> Vec<u16> {
   class_name
 }
 
-unsafe extern "system" fn window_proc(
```
Legend-Master, line 1382: This is now removed, the subclass is now the main window_proc

src/platform_impl/windows/window.rs
```diff
@@ -1166,199 +1275,143 @@ unsafe fn init<T: 'static>(
...
-    let window_state = Arc::new(Mutex::new(window_state));
-    WindowState::set_window_flags(window_state.lock(), real_window.0, |f| *f = window_flags);
-    window_state
-  };
-
-  let win = Window {
-    window: real_window,
-    window_state,
-    thread_executor: event_loop.create_thread_executor(),
-  };
```
Legend-Master, line 1328: This is moved to `InitData::create_window`

src/platform_impl/windows/event_loop.rs
```diff
@@ -1043,24 +1018,21 @@ unsafe fn public_window_callback_inner<T: 'static>(
       result = ProcResult::Value(LRESULT(0));
     }
 
-    win32wm::WM_NCCREATE => {
```
Legend-Master, line 1101: This is moved to `InitData::create_window`

src/platform_impl/windows/window.rs
```diff
@@ -1166,199 +1275,143 @@ unsafe fn init<T: 'static>(
...
+  // size.
+  if fullscreen.is_some() {
+    window.set_fullscreen(fullscreen);
+    force_window_active(window.hwnd());
+  } else if maximized {
+    window.set_maximized(true);
   }
 
-  win.set_visible(attributes.visible);
-  win.set_closable(attributes.closable);
```
Legend-Master, line 1350: This is moved to `InitData::on_create`, called from inside `event_loop::public_window_callback`, inside `CreateWindowExW`

except

```rust
  if fullscreen.is_some() {
    window.set_fullscreen(fullscreen);
    force_window_active(window.hwnd());
  } else if maximized {
    window.set_maximized(true);
  }
```

src/platform_impl/windows/window.rs
```diff
@@ -95,56 +96,7 @@ impl Window {
...
-              }
-            }),
-          )
-          .into();
-
-          assert!(RegisterDragDrop(win.window.0, &file_drop_handler).is_ok());
-          Some(file_drop_handler)
-        } else {
-          None
-        };
```
Legend-Master, line 133: This is moved to `InitData::create_window_data`

Legend-Master (commented):
Some notes to help with the review, the changes are mainly renames and indentations, so best viewed on a diff editor that makes it easier to view these

amrbashir (approved):

FabianLars:
I think this introduced a flicker of the titlebar in undecorated windows.
Try running `cargo run --example transparent` and it should flicker very briefly. checkout `a89d75aabbadc2f919acc820e1350a11f44e8758` (the commit before the merge), rebuild the example and it should not flicker anymore.

Legend-Master:
I couldn't really see the flicker on Windows 10 here, could you catch it in a screen recording or something? (wondering how exactly it looks like)

FabianLars:
https://github.com/user-attachments/assets/af7c733d-79d5-4fb0-aeef-5c3241dbbdd3

forgot to turn off hdr (and the music) but you can hopefully still see it.

Legend-Master:
I see, since I don't have a Windows 11 machine at the moment, and it doesn't seem to happen on my Windows 10 machine at all, could you give winit a quick test see if it has the same problems?

Also does moving 

https://github.com/tauri-apps/tao/blob/14ba5e9661bb20424123bbefa65f5fef1d61c5ca/src/platform_impl/windows/window.rs#L1210-L1223

to 

https://github.com/tauri-apps/tao/blob/14ba5e9661bb20424123bbefa65f5fef1d61c5ca/src/platform_impl/windows/window.rs#L1407

fixes it?

FabianLars:
Seems to happen on winit as well but the whole window kinda disappears after the flash lol

> Also does moving to fixes it

No. I can still see the titlebar flash but now the window is flashing white again on top of that

Legend-Master:
Anyhow, opened #1240 to track this, I could investigate more when I get my hands on a Windows 11 machine

PR #1238, opened 2026-06-10
refactor(windows): sync keyboard code from winit

Legend-Master:
Needs #1231
Reference: https://github.com/tauri-apps/tao/pull/1215#pullrequestreview-4256402119

Mainly getting in https://github.com/rust-windowing/winit/pull/2662

`MinimalIme` is removed in favor of `ImmGetCompositionStringW`, see https://github.com/tauri-apps/tao/pull/1238#discussion_r3395326462


src/platform_impl/windows/keyboard.rs
```diff
@@ -44,9 +48,6 @@ pub struct MessageAsKeyEvent {
   pub is_synthetic: bool,
 }
 
-pub(crate) static KEY_EVENT_BUILDERS: Lazy<Mutex<HashMap<WindowId, KeyEventBuilder>>> =
```
Legend-Master, line 47: This is moved to `WindowData::key_event_builder`

src/platform_impl/windows/event_loop.rs
```diff
@@ -994,22 +942,10 @@ unsafe fn public_window_callback_inner<T: 'static>(
...
-    let events = {
-      let mut key_event_builders =
-        crate::platform_impl::platform::keyboard::KEY_EVENT_BUILDERS.lock();
-      if let Some(key_event_builder) = key_event_builders.get_mut(&WindowId(window.0 as _)) {
-        key_event_builder.process_message(msg, wparam, lparam, next_key_message, &mut result)
-      } else {
-        Vec::new()
-      }
-    };
+    let events =
```
Legend-Master, line 942: The locks are all inside `key_event_builder` so there's no locks here anymore

src/platform_impl/windows/minimal_ime.rs
```diff
@@ -20,43 +23,46 @@ pub fn is_msg_ime_related(msg_kind: u32) -> bool {
 
 pub struct MinimalIme {
   // True if we're currently receiving messages belonging to a finished IME session.
-  getting_ime_text: bool,
+  getting_ime_text: AtomicBool,
```
Legend-Master: This alone probably doesn't do anything, we're still having an outer lock `userdata.window_state.lock()`

Legend-Master (commented):
Searching through winit's history, I can't seem to find any time where `MinimalIme` is used

> `git log --full-history -S MinimalIme`

And there were never an `ReceivedImeText` event either, it seems like this would be the `WindowEvent::Ime(Ime::Commit(text))` in today's winit, getting that one in is a bit hard though

src/platform_impl/windows/minimal_ime.rs
Legend-Master: Switched to use `ImmGetContext` and `ImmGetCompositionStringW` in 0e84ef3

amrbashir (approved):

PR #1245, opened 2026-06-25
fix(ios): use autorelease_ptr to fix UISceneConfiguration crash

### LINKED ISSUE #1244: [bug] iOS release build crashes on launch with UIApplicationSceneManifest enabled
**Describe the bug**

`configuration_for_connecting_scene_session` uses `Retained::as_ptr(&config)` to return a raw pointer to a
`UISceneConfiguration`. The `Retained<UISceneConfiguration>` is dropped at the end of the block, calling `objc_release` and deallocating the object before UIKit has a chance to retain it.

In release builds the optimizer drops `config` immediately after `as_ptr`, causing a reliable crash in
`-[UIApplication _connectUISceneFromFBSScene:transitionContext:]`.

The fix is to replace `Retained::as_ptr(&config)` with `Retained::autorelease_ptr(config)`, which transfers ownership to the Objective-C autorelease pool and keeps the object alive until UIKit retains it.

```diff
- Retained::as_ptr(&config)
+ Retained::autorelease_ptr(config)
```

**Steps To Reproduce**

1. Create a tao `EventLoop` and add `UIApplicationSceneManifest` to the app's
   `Info.plist` with `UIApplicationSupportsMultipleScenes` set to `true` and an
   empty `UISceneConfigurations` dictionary:
   ```xml
   <key>UIApplicationSceneManifest</key>
   <dict>
       <key>UIApplicationSupportsMultipleScenes</key>
       <true/>
       <key>UISceneConfigurations</key>
       <dict/>
   </dict>
   ```
2. Build for iOS in **release mode** (`cargo build --release --target aarch64-apple-ios`).
3. Deploy to a physical device or simulator.
4. Launch the app — it crashes immediately.

The crash does **not** occur in debug builds because the optimizer does not reorder the drop of `config` relative to the return.

**Expected behavior**

The app should launch without crashing. UIKit should receive a valid `UISceneConfiguration` from the delegate method.

**Screenshots**

Crash occurs in `objc_retain` called from
`-[UIApplication _connectUISceneFromFBSScene:transitionContext:]`, consistent with a use-after-free on the returned `UISceneConfiguration`.

- OS: iOS 26 (tested on physical iPad)
- Rustc: 1.94.0 (4a4ef493e 2026-03-02)
- tao: 0.35.3

velocitysystems:
Resolves #1244

## Description

[configuration_for_connecting_scene_session](cci:1://file:///Users/matthew/Bethel/Projects/jw-library/src-tauri/patches/tao/src/platform_impl/ios/view.rs:628:2-647:3) uses `Retained::as_ptr(&config)` to return a `UISceneConfiguration` to UIKit. The `Retained` is dropped at the end of the block, calling `objc_release` and deallocating the object before UIKit retains it. In release builds the optimizer drops `config` immediately after `as_ptr`, causing a reliable crash in `-[UIApplication _connectUISceneFromFBSScene:transitionContext:]`.

## Fix

Replace `Retained::as_ptr(&config)` with `Retained::autorelease_ptr(config)`, which transfers ownership to the autorelease pool and keeps the object alive until UIKit retains it.

## Testing

Verified on a physical iPad running iOS 26:

1. Unpatched release build — crashes on every launch
2. Patched release build — launches successfully
3. Reverted to unpatched — crash returns
4. Re-applied patch — launches successfully

### DIFF
--- a/.changes/ios-multiwindow-crash.md
+++ b/.changes/ios-multiwindow-crash.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Prevent use-after-free in iOS `configurationForConnectingSceneSession` that crashes release builds on launch.
--- a/src/platform_impl/ios/view.rs
+++ b/src/platform_impl/ios/view.rs
@@ -643,7 +643,7 @@ pub fn create_delegate_class() {
 
       // Dynamically set the delegate class name
       config.setDelegateClass(Some(super::scene::TaoSceneDelegate::class()));
-      Retained::as_ptr(&config) as _
+      Retained::autorelease_ptr(config) as _
     }
   }
 


velocitysystems:
Verified this fix resolves the latent Tao bug. While further work is recommended to the Tao state machine to handle multiple-scene support this resolves the crash on startup.

velocitysystems:
@FabianLars Can we get this in the next release? It fixes a crash when enabling multi-window support on iOS 26.

FabianLars (approved):

PR #1250, opened 2026-06-29
fix(ios): use minimal windowing control style on iPadOS 26

### LINKED ISSUE #15521: [bug] iPadOS 26 window controls overlap content with no way to offset or avoid them
### Describe the bug

iPadOS 26 introduces a new windowing system in which every app scene gets macOS-style window controls rendered by the system in the top-left corner of the window. For Tauri apps on iPadOS 26, these controls are drawn **on top of the webview content**, overlapping any page UI positioned near the top-left (headers, navigation buttons, menu toggles).

Image

This is the iOS counterpart to #4789, which was resolved for macOS by PR #12366. The same class of problem now exists on iPadOS 26, but the macOS solution cannot be ported directly:

- On macOS, AppKit exposes the actual button views (`NSWindow.standardWindowButton(...)`), so tao/wry can simply move them (`inset_traffic_lights()` in `wry/src/wkwebview/class/wry_web_view_parent.rs`).
- On iPadOS 26, the window controls are **system-owned chrome with no view handle**. There is no API to reposition them freely.

Compounding the problem for webview-based apps:
- The controls do **not** contribute to the safe area, so `env(safe-area-inset-top)` / `env(safe-area-inset-left)` in CSS report `0` and there is no pure-CSS workaround.
- There is no event or API surfaced to the web layer to detect whether the controls are visible (they disappear in fullscreen) or how much space they occupy, so static padding hacks are unreliable.

Capacitor has the identical problem open as [ionic-team/capacitor#8172](https://github.com/ionic-team/capacitor/issues/8172).

### Reproduction

1. Create a Tauri app: `npm create tauri-app@latest`
2. Add the iOS target and init: `cargo tauri ios init`
3. Run on an iPad (or iPad simulator) running iPadOS 26: `cargo tauri ios dev`
4. Put the app into windowed mode (drag the bottom-right resize handle, or use the window arrangement menu).
5. Observe the system window controls in the top-left corner overlapping the page content. Any UI element rendered in that corner is partially or fully obscured and may be unreachable.

### Expected behavior

Tauri should provide a way to adopt the native iPadOS 26 windowing behavior so the window controls integrate into the status bar / title bar region like native apps, instead of overlapping content.

Apple's model is *avoidance and style selection*, not free positioning, so the expected API surface differs from macOS's `trafficLightPosition`:

1. **Window control style selection** — expose `preferredWindowingControlStyle(for:)` (`UIWindowScene.WindowingControlStyle`: `.automatic` / `.minimal` / `.unified`) via an iOS-specific config option, e.g.:

   ```json
   {
     "app": {
       "windows": [
         {
           "windowControlsStyle": "unified"
         }
       ]
     }
   }
   ```

`.unified` is the native iPadOS 26 behavior where the controls integrate with the title bar region and system components (`UINavigationBar`, status bar strip) arrange around them — i.e. "in the status bar like native apps".

2. **Inset reporting to the web layer** — bridge the region occupied by the controls recommended by Apple in
   [developer forums thread 787227](https://developer.apple.com/forums/thread/787227)) into the webview as CSS custom properties (e.g. `--window-controls-inset-left`, `--window-controls-inset-top`). This mirrors he CSS Window Controls Overlay `titlebar-area-*` env() variables that WKWebView does not implement, and matches how Tauri apps already handle the macOS overlay title bar.

### Workaround

Set `UIDesignRequiresCompatibility = YES` as documented in [ionic-team/capacitor#8172](https://github.com/ionic-team/capacitor/issues/8172). However per the [Apple Docs](https://developer.apple.com/documentation/bundleresources/information-property-list/uidesignrequirescompatibility) the system ignores this key when you build for iOS 27 or later.

Image

### Additional context

**Related** 
- #4789
- tauri-apps/tauri#12366 — macOS implementation (`trafficLightPosition`)
- [ionic-team/capacitor#8172](https://github.com/ionic-team/capacitor/issues/8172)
- [Apple Developer Forums thread 787227](https://developer.apple.com/forums/thread/787227)
- [What's new in iPadOS 26](https://developer.apple.com/ipados/whats-new/)
- [UIWindowScene documentation](https://developer.apple.com/documentation/uikit/uiwindowscene)

velocitysystems:
Resolves tauri-apps/tauri#15521

## Description

iPadOS 26 adds macOS-style system window controls to the top-left corner of every windowed scene. Because a Tao window hosts a fullscreen `WKWebView` with no native toolbar, these controls do not contribute to the safe area and float on top of, overlapping, the web content. There is no native toolbar for them to integrate with.

## Fix

Implement `UIWindowSceneDelegate`'s optional `preferredWindowingControlStyle(for:)` method on `TaoSceneDelegate`, returning the `minimal` style. The `minimal` style places the controls in a safe-area strip above the content instead of floating over it. This is the same approach taken by Flutter and Capacitor.

No runtime version check is needed: `preferredWindowingControlStyleForScene:` is an optional protocol method the system only sends on iPadOS 26+, and the `UISceneWindowingControlStyle` class is resolved lazily at call time. On earlier iOS versions the conformance is inert and nothing changes.

This also enables the `UISceneWindowingControlStyle` and `UIWindowScene` features on `objc2-ui-kit` and bumps the `objc2` crate family `0.3.0` -> `0.3.2`, since the `UISceneWindowingControlStyle` type does not exist in `0.3.0`. The bump is patch-level and the macOS build is unaffected.

## Testing

- `cargo check --target aarch64-apple-ios -p tao` compiles with no errors.
- `cargo check -p tao` (macOS host) still compiles, confirming the dependency bump does not regress the desktop build.
- Verified on iPadOS 26 on a physical device

## Related

https://github.com/madsmtm/objc2/issues/849

### DIFF
--- a/.changes/ios-windowing-control-style.md
+++ b/.changes/ios-windowing-control-style.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fix iPadOS 26 system window controls overlapping `WKWebView` content by implementing `preferredWindowingControlStyleForScene:` on the scene delegate and returning the `minimal` style. The optional protocol method is a no-op on iOS versions earlier than 26.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1616,47 +1616,49 @@ dependencies = [
  "bitflags 2.7.0",
  "objc2 0.6.4",
  "objc2-core-foundation",
- "objc2-foundation 0.3.0",
+ "objc2-foundation 0.3.2",
 ]
 
 [[package]]
 name = "objc2-cloud-kit"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c1948a9be5f469deadbd6bcb86ad7ff9e47b4f632380139722f7d9840c0d42c"
+checksum = "73ad74d880bb43877038da939b7427bba67e9dd42004a18b809ba7d87cee241c"
 dependencies = [
  "bitflags 2.7.0",
  "objc2 0.6.4",
- "objc2-foundation 0.3.0",
+ "objc2-foundation 0.3.2",
 ]
 
 [[package]]
 name = "objc2-core-data"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1f860f8e841f6d32f754836f51e6bc7777cd7e7053cf18528233f6811d3eceb4"
+checksum = "0b402a653efbb5e82ce4df10683b6b28027616a2715e90009947d50b8dd298fa"
 dependencies = [
  "objc2 0.6.4",
- "objc2-foundation 0.3.0",
+ "objc2-foundation 0.3.2",
 ]
 
 [[package]]
 name = "objc2-core-foundation"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "daeaf60f25471d26948a1c2f840e3f7d86f4109e3af4e8e4b5cd70c39690d925"
+checksum = "2a180dd8642fa45cdb7dd721cd4c11b1cadd4929ce112ebd8b9f5803cc79d536"
 dependencies = [
  "bitflags 2.7.0",
+ "dispatch2",
  "objc2 0.6.4",
 ]
 
 [[package]]
 name = "objc2-core-graphics"
-version = "0.3.0"
+version = "0.3.2"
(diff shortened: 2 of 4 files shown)


velocitysystems:
@FabianLars This now forces Tao to use the correct windowing mode on iOS 26 (and is compatible with iOS 27).
Can we get this into the next release? Thanks!

image

FabianLars (approved):
i have no means to test this atm but looks safe enough

velocitysystems:
> i have no means to test this atm but looks safe enough

I've verified this with local debug + release builds and a CI/CD build pushed to TestFlight.

PR #1254, opened 2026-07-02
fix(macos): reapply traffic light inset on title change and fullscreen exit

tenderdeve:
Moving the traffic-light-reset fix down to tao, per the discussion on tauri-apps/tauri#15628.

### What
On macOS the custom traffic light inset (`set_traffic_light_inset`) is lost on two events:
- **`setTitle:`** — the buttons snap back to default (tauri-apps/tauri#13044)
- **leaving fullscreen** of a maximized window — same reset (tauri-apps/tauri#15451)

Both reporters noted resizing the window fixes it, which is the tell: `draw_rect` already re-insets on redraw (view.rs), but neither `setTitle:` nor `windowDidExitFullScreen:` triggers a `drawRect:`, so the position stays default until the next redraw.

### How
Reapply the inset **already stored on the view's `ViewState` (`traffic_light_inset`)** after both events:
- `set_title_async` re-insets right after `setTitle:`
- `window_did_exit_fullscreen` re-insets after the transition completes

No new state — the value is read back from the existing `taoState` ivar, so it can't drift from what `set_traffic_light_inset` wrote. New helper `reapply_traffic_light_inset(ns_window, ns_view)` in `view.rs`.

@FabianLars this is the ivar-based version we talked about — supersedes your `refresh-lights` branch (that one threaded a new `SharedState.traffic_lights_position` and only covered the title case). Also covers fullscreen exit here. `cargo check` is green; I don't have a mac to smoke-test the button positions, so a visual check would be appreciated.

### DIFF
--- a/.changes/fix-macos-traffic-light-reset.md
+++ b/.changes/fix-macos-traffic-light-reset.md
@@ -1,0 +1,7 @@
+---
+"tao": patch
+---
+
+On macOS, re-apply the custom traffic light inset after a title change and
+after leaving fullscreen, so the buttons no longer jump back to their default
+position on those events.
--- a/src/platform_impl/macos/util/async.rs
+++ b/src/platform_impl/macos/util/async.rs
@@ -17,6 +17,7 @@ use crate::{
   dpi::LogicalSize,
   platform_impl::platform::{
     ffi::{self, id, NO, YES},
+    view::reapply_traffic_light_inset,
     window::SharedState,
   },
 };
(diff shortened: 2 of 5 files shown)


FabianLars:
the title fix works good. couldn't reproduce the fullscreen one, waiting a few days for the issue author to respond https://github.com/tauri-apps/tauri/issues/15451

FabianLars (approved):
didn't hear back yet but the fullscreen change can't hurt i think so imo ready to go

PR #1257, opened 2026-07-03
fix(ios): emit `Opened` event for URLs on cold start

### LINKED ISSUE #1248: iOS: cold-start URL/document opens are dropped — connect_scene ignores connectionOptions.URLContexts
### Summary

Under the scene lifecycle, an iOS app launched **by** a URL or document open (custom scheme, universal link, or a `CFBundleDocumentTypes` "Open with" file) never receives the URL. tao emits `Event::Opened` only for **warm** opens (app already running); **cold-start** opens are silently dropped.

### Environment

- tao 0.35.2 (code path unchanged on `dev`)
- iOS, scene lifecycle (the default since #1154)

### Root cause

iOS delivers the launch URL two different ways under the scene lifecycle:

- App **already running** → `scene(_:openURLContexts:)`. tao handles this: `TaoSceneDelegate::scene_openURLContexts` parses the URLs and emits `Event::Opened` ✅ (`src/platform_impl/ios/scene.rs`).
- App **not running** (cold start) → the URLs arrive in `scene(_:willConnectTo:options:)` via `connectionOptions.URLContexts`. tao's `connect_scene` (`src/platform_impl/ios/app_state.rs`) sets up the window/scene but **never reads `options.URLContexts()`**, so no `Event::Opened` is emitted ❌.

There's no app-delegate fallback either: for scene-based apps the launch URL is **not** placed in `application(_:didFinishLaunchingWithOptions:)`'s `launchOptions`, so once the scene callback ignores it, it's unrecoverable.

### Expected vs actual

- **Expected:** a cold-start open emits `Event::Opened { urls }`, matching the warm path.
- **Actual:** nothing is emitted; the app launches with no indication a URL/file triggered it.

### Impact

- `tauri-plugin-deep-link`'s `getCurrent()` returns `null` on cold start and `onOpenUrl` never fires — deep links / universal links / document opens only work if the app is already running. Affects every Tauri iOS app using the plugin.
- Document-based apps (`CFBundleDocumentTypes` / "Open with") can't receive a file when launched by it.

### Repro

1. Scene-lifecycle iOS app that logs `Event::Opened`.
2. Register a custom scheme (or a document type).
3. Fully quit the app; trigger the scheme / "Open with" from another app.
4. App launches; `Event::Opened` never fires. (Warm — app already open — works.)

### Possible fix

Extract the URL-parsing/emit from `scene:openURLContexts:` into a shared helper, and call it from `connect_scene` on `options.URLContexts()` so cold and warm opens are symmetric. If the event loop isn't ready yet the event queues via `try_user_callback_transition`. (Could also be extended to `connectionOptions.userActivities` for universal links; this covers `URLContexts` — custom schemes + document opens.)

```diff
diff --git a/src/platform_impl/ios/app_state.rs b/src/platform_impl/ios/app_state.rs
@@ use crate::{
-    scene::multiple_scenes_enabled,
+    scene::{emit_opened_from_url_contexts, multiple_scenes_enabled},
@@ pub unsafe fn connect_scene(scene: &UIScene, options: &UISceneConnectionOptions)
     }
   }
+
+  // A cold-start URL/document open (the app was launched *by* it) is delivered
+  // here in `connectionOptions`, not via `scene:openURLContexts:` (which only
+  // fires while the app is already running). Without this, such opens are
+  // silently dropped under the scene lifecycle. Deliver them through the same
+  // `Event::Opened` path as the warm case; if the event loop isn't ready yet
+  // the event is queued (see `try_user_callback_transition`).
+  emit_opened_from_url_contexts(&options.URLContexts());
 }

diff --git a/src/platform_impl/ios/scene.rs b/src/platform_impl/ios/scene.rs
@@ pub unsafe fn multiple_scenes_enabled() -> bool {
   (*num).as_bool()
 }

+/// Parse the URLs out of a set of open-URL contexts and, if any, deliver them
+/// as a single `Event::Opened`. Shared by the two ways iOS surfaces a URL/
+/// document open to a scene-based app: `scene:openURLContexts:` (app already
+/// running) and the `connectionOptions` of `scene:willConnectToSession:`
+/// (app launched *by* the open — see `app_state::connect_scene`). Keeping
+/// both on this one path makes warm and cold opens symmetric.
+pub(crate) unsafe fn emit_opened_from_url_contexts(url_contexts: &NSSet<UIOpenURLContext>) {
+  let urls: Vec<url::Url> = url_contexts
+    .iter()
+    .filter_map(|ctx| {
+      ctx.URL().absoluteString().and_then(|url| {
+        let url = url.to_string();
+        url.parse().map_err(|e| {
+          log::error!("failed to parse URL {url} from open URL context: {e}");
+          e
+        }).ok()
+      })
+    })
+    .collect();
+  if !urls.is_empty() {
+    app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::Opened { urls }));
+  }
+}
+
 define_class!(
@@     #[unsafe(method(scene:openURLContexts:))]
     fn scene_openURLContexts(&self, _scene: &UIScene, url_contexts: &NSSet<UIOpenURLContext>) {
       unsafe {
-        let urls: Vec<url::Url> = url_contexts
-          .iter()
-          .filter_map(|ctx| { /* …same parsing… */ })
-          .collect();
-        if !urls.is_empty() {
-          app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::Opened { urls }));
-        }
+        emit_opened_from_url_contexts(url_contexts);
       }
     }
```

Verified on device (iPhone 15 Pro, tao 0.35.2): with this change a cold-start document open ("Open with") now fires `Event::Opened` exactly like the warm case.

balabiturembekov:
Closes #1248

## Problem

On a cold start, URLs delivered through `connectionOptions.URLContexts` in `scene:willConnectToSession:options:` were ignored, so `Event::Opened` was never emitted. Deep links (custom URL schemes and universal links, e.g. via `tauri-plugin-deep-link`) only worked while the app was already running, where `scene:openURLContexts:` handles them.

## Changes

- Extract the shared URL-contexts handling into `emit_opened_from_url_contexts` in `scene.rs` and call it from both `scene:openURLContexts:` (warm start) and `connect_scene` (cold start).
- Extract the pure `parse_url_strings` helper and reuse it for universal links in `scene:continueUserActivity:`, which previously panicked (`.unwrap()`) on URLs the `url` crate cannot parse; such URLs are now logged and skipped.
- Add unit tests for `parse_url_strings`: deep-link round-trips (custom schemes, query/fragment, file URLs), invalid input filtering, empty input, and WHATWG normalization behavior.

## Test plan

CI compiles the tests for the iOS target (`cargo test --no-run`) but does not execute them; all five tests were executed locally and pass. `cargo check --target aarch64-apple-ios --tests` is warning-free for the touched files and `cargo fmt --check` passes.

### DIFF
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -28,7 +28,7 @@ use crate::{
       CFRunLoopTimerRef, CFRunLoopTimerSetNextFireDate, CGRect, CGSize, NSInteger,
       NSOperatingSystemVersion, NSUInteger,
     },
-    scene::multiple_scenes_enabled,
+    scene::{emit_opened_from_url_contexts, multiple_scenes_enabled},
   },
   window::WindowId as RootWindowId,
 };
@@ -530,6 +530,8 @@ pub unsafe fn connect_scene(scene: &UIScene, options: &UISceneConnectionOptions)
       }));
     }
   }
+
+  emit_opened_from_url_contexts(&options.URLContexts());
 }
 
 pub unsafe fn register_window_for_scene(window: id) {
--- a/src/platform_impl/ios/app_state.rs
+++ b/src/platform_impl/ios/app_state.rs
@@ -28,7 +28,10 @@ use crate::{
       CFRunLoopTimerRef, CFRunLoopTimerSetNextFireDate, CGRect, CGSize, NSInteger,
       NSOperatingSystemVersion, NSUInteger,
     },
-    scene::{emit_opened_from_url_contexts, multiple_scenes_enabled},
+    scene::{
+      emit_opened, multiple_scenes_enabled, url_strings_from_url_contexts,
+      url_strings_from_user_activities,
+    },
   },
   window::WindowId as RootWindowId,
 };
@@ -531,7 +534,13 @@ pub unsafe fn connect_scene(scene: &UIScene, options: &UISceneConnectionOptions)
     }
   }
 
-  emit_opened_from_url_contexts(&options.URLContexts());
+  // on cold start every deep link arrives here instead of through the callbacks
+  // that handle them while the app is running: custom schemes and file URLs in
+  // the URL contexts, universal links in the user activities, since
+  // scene:continueUserActivity: is not called for the launch activity
+  let mut url_strings = url_strings_from_url_contexts(&options.URLContexts());
+  url_strings.extend(url_strings_from_user_activities(&options.userActivities()));
+  emit_opened(&url_strings, "scene:willConnectToSession:options:");
 }
 
 pub unsafe fn register_window_for_scene(window: id) {
--- a/src/platform_impl/ios/scene.rs
+++ b/src/platform_impl/ios/scene.rs
@@ -17,6 +17,20 @@ use crate::{
   window::WindowId as RootWindowId,
 };
 
+pub(crate) fn emit_opened_from_url_contexts(url_contexts: &NSSet<UIOpenURLContext>) {
+  let url_strings: Vec<String> = url_contexts
+    .iter()
+    .filter_map(|ctx| ctx.URL().absoluteString().map(|s| s.to_string()))
+    .collect();
+
+  let urls = parse_url_strings(&url_strings);
+  if !urls.is_empty() {
+    unsafe {
+      app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::Opened { urls }));
+    }
+  }
+}
+
 // true when the system allows the app to display multiple scenes and multiple_scenes_enabled() returns true
 // https://developer.apple.com/documentation/uikit/uiapplication/supportsmultiplescenes?language=objc
 pub unsafe fn app_supports_multiple_scenes() -> bool {
@@ -108,26 +122,7 @@ define_class!(
 
     #[unsafe(method(scene:openURLContexts:))]
     fn scene_openURLContexts(&self, _scene: &UIScene, url_contexts: &NSSet<UIOpenURLContext>) {
-      unsafe {
-        let urls: Vec<url::Url> = url_contexts
-          .iter()
-          .filter_map(|ctx| {
-            ctx.URL().absoluteString().and_then(|url| {
-              let url = url.to_string();
-              url
-                .parse()
-                .map_err(|e| {
-                  log::error!("failed to parse URL {url} from scene:openURLContexts: {e}");
-                  e
-                })
-                .ok()
-            })
-          })
-          .collect();
-        if !urls.is_empty() {
-          app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::Opened { urls }));
-        }
-      }
+      emit_opened_from_url_contexts(url_contexts);
     }
 
     #[unsafe(method(stateRestorationActivityForScene:))]
@@ -162,10 +157,10 @@ define_class!(
           .webpageURL()
           .and_then(|url| url.absoluteString())
         {
-          let url = url.to_string().parse::<url::Url>().unwrap();
-          app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::Opened {
-            urls: vec![url],
-          }));
+          let urls = parse_url_strings(&[url.to_string()]);
+          if !urls.is_empty() {
+            app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::Opened { urls }));
+          }
         }
       }
     }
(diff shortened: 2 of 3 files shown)


src/platform_impl/ios/app_state.rs
```diff
@@ -530,6 +530,8 @@ pub unsafe fn connect_scene(scene: &UIScene, options: &UISceneConnectionOptions)
       }));
     }
   }
+
+  emit_opened_from_url_contexts(&options.URLContexts());
```
velocitysystems: `connect_scene` reads options.URLContexts(), which covers custom URL schemes and file URLs on cold start — but _universal links_ (https://…) are delivered on cold start via `options.userActivities()`.

Since `scene:continueUserActivity:` isn't called for the launch activity, a universal link that cold-launches the app still won't emit `Event::Opened`. Was universal-link cold start intended to be in scope for #1248?

If so, `connect_scene` would also need to walk `options.userActivities()` — ideally via a second shared helper, the same way `parse_url_strings` was extracted. If not, could the changelog line be narrowed from "URLs" to "custom-scheme URLs" and a follow-up issue created?

src/platform_impl/ios/scene.rs
```diff
@@ -196,3 +191,93 @@ define_class!(
     }
   }
 );
+
+fn parse_url_strings(url_strings: &[String]) -> Vec<url::Url> {
```
velocitysystems: The shared error message dropped the original "from scene:openURLContexts:" qualifier in favor of a generic "failed to parse URL {s}". With three call sites now sharing this helper, logs no longer tell you which delivery path produced a bad URL.

velocitysystems (changes requested):
Good work @balabiturembekov. Just a few comments for your consideration.

velocitysystems:
Hi @balabiturembekov. Are you planning to make these changes? If not, I'd like to move forward with making them.

balabiturembekov:
Sorry for the delay — just pushed both changes.

src/platform_impl/ios/scene.rs
```diff
@@ -17,6 +17,45 @@ use crate::{
...
+pub(crate) fn url_strings_from_user_activities(
+  user_activities: &NSSet<NSUserActivity>,
+) -> Vec<String> {
+  user_activities
+    .iter()
+    .filter_map(|activity| webpage_url_string(&activity))
+    .collect()
+}
+
+fn webpage_url_string(user_activity: &NSUserActivity) -> Option<String> {
```
velocitysystems, line 41: `webpageURL` isn't exclusive to universal links — a handoff activity can carry one too, so this doesn't quite match the comment on `url_strings_from_user_activities` above, which has handoff among the activities being skipped.

Apple's universal-links guide guards on the activity type first; could `webpage_url_string` do the same? That would also cover `scene:continueUserActivity:`, which has the same gap.

```rust
if !user_activity
  .activityType()
  .isEqualToString(unsafe { NSUserActivityTypeBrowsingWeb })
{
  return None;
}
```

Both methods are safe in objc2-foundation 0.3.2; the constant just needs adding to the `objc2_foundation` import. Probably not hit in practice today, but it fails silently when it is.

velocitysystems (changes requested):
Thanks @balabiturembekov. One last comment for your consideration.

balabiturembekov:
Done

src/platform_impl/ios/app_state.rs
```diff
@@ -530,6 +533,14 @@ pub unsafe fn connect_scene(scene: &UIScene, options: &UISceneConnectionOptions)
       }));
     }
   }
+
+  // on cold start every deep link arrives here instead of through the callbacks
+  // that handle them while the app is running: custom schemes and file URLs in
+  // the URL contexts, universal links in the user activities, since
+  // scene:continueUserActivity: is not called for the launch activity
+  let mut url_strings = url_strings_from_url_contexts(&options.URLContexts());
+  url_strings.extend(url_strings_from_user_activities(&options.userActivities()));
```
velocitysystems, line 542: **Blocker: these two accessors abort on an ordinary launch.**

`UISceneConnectionOptions.URLContexts` is declared `nonnull`, so objc2 generates it as non-`Option` and panics when the message returns nil — which is exactly what happens on a cold launch with no deep link:

```
thread '<unnamed>' panicked at objc2-ui-kit-0.3.2/src/generated/UISceneOptions.rs:27:5:
unexpected NULL returned from -[UISceneConnectionOptions URLContexts]
```

Reproduced on iPhone 17 Pro / iOS 26.5 in a headless scene-enabled app that opens no URL at all. Since `connect_scene` runs for every scene connection, this is a startup crash for every consumer of the PR, not just deep-link users.

Reading both through `msg_send!` as `Option<Retained<…>>` and treating nil as "no launch payload" fixes it; I ran this exact change on both runtimes and both launch cleanly:

```rust
// The generated accessors assume Apple's nonnull annotations and panic if
// UIKit returns nil. Treat nil as no launch payload instead.
let url_contexts: Option<Retained<NSSet<UIOpenURLContext>>> = msg_send![options, URLContexts];
let user_activities: Option<Retained<NSSet<NSUserActivity>>> =
  msg_send![options, userActivities];

let mut url_strings = Vec::new();
if let Some(url_contexts) = url_contexts {
  url_strings.extend(url_strings_from_url_contexts(&url_contexts));
}
if let Some(user_activities) = user_activities {
  url_strings.extend(url_strings_from_user_activities(&user_activities));
}
```

(needs `use objc2_foundation::{NSSet, NSUserActivity};` and `UIOpenURLContext` added to the `objc2_ui_kit` import; `msg_send!` is already in scope crate-wide via the `#[macro_use]` in `lib.rs`, and the getter keeps the correct retain semantics.)

src/platform_impl/ios/scene.rs
```diff
@@ -17,6 +17,55 @@ use crate::{
   window::WindowId as RootWindowId,
 };
 
+// custom URL schemes and file URLs, delivered both on cold start
+// (UISceneConnectionOptions) and while the app is running (scene:openURLContexts:)
+pub(crate) fn url_strings_from_url_contexts(url_contexts: &NSSet<UIOpenURLContext>) -> Vec<String> {
+  url_contexts
+    .iter()
+    .filter_map(|ctx| ctx.URL().absoluteString().map(|s| s.to_string()))
```
velocitysystems, line 25: Minor, same class of assumption: `UIOpenURLContext.URL` is also `nonnull`-annotated and generated as non-`Option`, so a nil here would abort the app rather than skip one URL. I have not seen it return nil (unlike `URLContexts`), so I would not block on it — just noting that the surviving unconditional accessor in this file is this one.

src/platform_impl/ios/scene.rs
```diff
@@ -156,17 +189,9 @@ define_class!(
 
     #[unsafe(method(scene:continueUserActivity:))]
     fn scene_continueUserActivity(&self, _scene: &UIScene, user_activity: &NSUserActivity) {
```
velocitysystems: Have you been able to confirm that a universal link that cold-starts the app arrives *only* in `connectionOptions.userActivities` and not also through this callback shortly after `scene:willConnectToSession:options:`? Apple documents `scene(_:continue:)` as the running-app path, but double delivery on launch has been reported often enough that a duplicate `Event::Opened` seems worth ruling out — consumers like `tauri-plugin-deep-link` would fire their handler twice. If it can happen, dropping an activity already seen in the connection options would be the cheap guard.

src/platform_impl/ios/scene.rs
```diff
@@ -196,3 +221,112 @@ define_class!(
...
+          log::error!("failed to parse URL {s} from {source}: {e}");
+          e
+        })
+        .ok()
+    })
+    .collect()
+}
+
+#[cfg(test)]
+mod tests {
```
velocitysystems, line 240: These tests live in an iOS-only module, so `cargo test` on a host target never builds them and the iOS job only does `--no-run` — as the PR description says, they were run by hand. They also mostly assert `url` crate behaviour (normalization, `https://` rejection) rather than anything this file decides. Since the one piece of logic that is really yours is "unparseable URLs are logged and skipped, the rest are kept", I would keep just that case and drop the round-trip/normalization ones, or move `parse_url_strings` somewhere host-testable so CI actually runs them.

velocitysystems (changes requested):

PR #1104, opened 2025-03-25
fix(linux): Port to gtk4-rs

conradhale:
Closes #1051

You can find more info on most of the changes in the [GTK docs](https://docs.gtk.org/gtk4/migrating-3to4.html)

### Summary of changes
Most of the X11-specific features have been removed from GTK4. This includes getting/setting window position, setting cursor position, getting global coordinates of pointer, setting window level (always-on-top), making window visible on all workspaces, and setting the skip taskbar hint.

Some of these methods can still be found in gdk-x11, however, X11 support is set to be completely removed from GTK5, and has been deprecated in GTK4.18. I personally do not think it is worth the effort to add the extra code to continue supporting these features though. Most developers using TAO really shouldn't depend on X11-only features since the number of people using X11 is becoming increasingly small. Differences in behavior on X11 and Wayland is also in many cases a recipe for bugs.

I've also added some features that were previously missing from the Wayland backend, like getting the cursor position (relative to the surface), and setting the urgency hint for the window (uses xdg-activation like winit).

gtk-rs-core also removed MainContext::channel, so I've followed the reccomendation of using `context.spawn_local()` with `async-channel`.

Also, due to the `GtkWidget::configure_event` being removed, I had to subclass GtkWindow to allow connecting to changes in the Window's size allocation, since the `default-width` and `default-height` property change notification are fired too late to be used for making layout changes.

Finally, I switched to using gdk-x11 and gdk-wayland instead of gdk-x11-sys and gdk-wayland-sys. It's now possible to get the window xid and wl_surface id without unsafe code, making the raw_window_handle functions implementations closer to winit's.

### Todo
- Support creating window with Libadwaita (Enabled via cargo feature probably).
- Update .changes
- Test on X11

PR #1258, opened 2026-07-03
feat(linux): port to gtk4

### LINKED ISSUE #1051: Upgrade `tao` to `gtk4-rs`
We need to upgrade from [`gtk`](https://crates.io/crates/gtk) to [`gtk4-rs`](https://crates.io/crates/gtk4) to migrate a unsoundness [issue](https://github.com/advisories/GHSA-wrw7-89jp-8q8g) found in glib < 0.2.
Please claim this issue if you want to support or work on this and keep in mind this is a prioritized issue and we try to migrate as soon as possible. If the issue is already assigned to someone you can either reach out with comments or questions reach out for collaboration to the assigned person.

This transition includes:

- [ ]  Upgrading [`gtk`](https://crates.io/crates/gtk) to [`gtk4-rs`](https://crates.io/crates/gtk4) and handling all errors caused by the dependency change
- [ ]  Handle all changes in a backwards compatible way if possible
    - [ ]  Raise awareness to the working group if breaking changes are required
- [ ]  Document migration errors that may be issues in the future or which were complicated to handle
- [ ]  Reach out to the working group (alternatively @tweidinger first) if blockers or upstream issues are hit

johncarmack1984:
Closes #1051. Supersedes #1104. This takes @conradhale's GTK4 port and @rmakestrash-jpg's continuation of it (rebase to 0.34.5, the optional libadwaita feature, the `gtk::init` ordering fix, plus @aurelj's no-x11 compile fix), merges everything onto current dev, and bumps the stack to gtk4-rs 0.11 / glib 0.22. Their commits carry their authorship in the branch history; my work is the forward merge, the 0.11 bump, and CI.

Discussed in tauri-apps/wry#1474. gtk4 0.11 here matches @Ranrar's wry gtk4/webkit6 branch, so the two can land as a coherent pair.

What the forward merge had to reconcile, so nobody has to re-derive it in review:

- #1173's RwLock soundness fix applies cleanly to the port's `Window` fields.
- #1141 (xdg-desktop-portal color scheme): `portal.rs` is rewritten from glib channels (removed in glib 0.20+) to async-channel, matching how the port already moves events. The dummy-window `SetTheme` arm now emits `ThemeChanged` like upstream.
- #1228's JIS key mappings are translated to gdk4 `Key` constants.
- #1218's event dedup and `Propagation::Proceed` changes are gtk3-signal-specific. GTK4 event controllers neither swallow decoration events nor double-deliver, so the port keeps its plain controllers. The xdg_surface buffer-size workaround is likewise replaced by the port's existing maximize process.

Breaking, beyond the obvious type swap: public gtk types in the Unix extension traits are gtk4 0.11 types; the `x11` feature is backed by gdk4-x11; new optional `libadwaita` feature; MSRV rises to 1.92, gtk4-rs 0.11's floor. The MSRV bump is crate-wide, not Linux-only, and it's the one call here I most want maintainer eyes on.

Verified: `cargo check --all-features`, `clippy --all-targets` (lib clean; the two remaining lints are in examples unchanged from dev), and `cargo test`, on rust 1.96 / Debian trixie / GTK 4.18. Both things this draft was waiting on have since happened: tauri itself builds against this branch (wry on webkit6, muda on gtk4), and a real Tauri app runs on it under X11 -- window creation, the event loop, and keyboard input all through this port. The menu in this capture was opened by key events going through the gtk4 event controllers:

a Tauri app running on this tao branch

Wayland session time is what's left.


amrbashir:
Thanks for the PR but I should let you know that I am working on alternative that uses current winit code base at https://github.com/tauri-apps/winit-gtk4/tree/amr/feat/gtk4

I don't know yet but there is a good chance will be using winit-gtk4 instead of porting tao to gtk4

johncarmack1984:
winit-gtk4 on the winit codebase makes sense to me. If that's the direction, happy to help over there instead. point me at what needs doing. No attachment to this specific PR; whatever transfers (the muda side, the headless test setup), transfers. My only aim is resolving the security warnings.

.github/workflows/sync-upstream.yml
```diff
@@ -0,0 +1,53 @@
+name: Sync with Upstream
```
Legend-Master: let's remove this first so I could trigger ci runs

johncarmack1984: Done, removed in 4884738. Thanks for taking a look.

FabianLars:
if we use tao (in the beginning*) instead of winit then i wonder if it makes sense to put it behind a feature flag and have gtk3 and gtk4 side by side for a while. That would make migration in wry and tauri a bit easier i think and would potentially make a 2.x release possible.

Alternatively we could release tao gtk4 as 0.50 or something but having to deal with to release trains kinda sucks.

\* in the beginning: i'm not sure about winit-gtk4's, and more importantly winit 0.31's timelines so perhaps it makes sense to start the gtk4 migration with tao and then later on migrate to winit. That would allow us to move a bit quicker. If we don't even want to consider a 2.x tauri (pre)release than this doesn't matter i guess.

johncarmack1984:
@FabianLars The flag is realistic. Everything gtk-facing lives in platform_impl/linux plus the unix extension traits, so the shape would be: today's backend moves to linux/gtk3 unchanged, this port lands as linux/gtk4, mutually exclusive features with gtk3 as the default, compile_error! on both or neither. CI grows a second linux leg. Current users see zero change until they opt in.

Two caveats. The unix extension traits expose gtk types, so those signatures switch with the feature. wry and tauri cross that seam either way, the flag just makes it opt-in. And gtk4 dropped the global-position APIs, so outer_position and set_always_on_bottom go unsupported there. Side by side means nobody inherits that until they choose it.

The RUSTSEC warnings only clear for builds that opt in, so I'd call the flag the migration path rather than the fix. For tauri the same flag would need to travel through wry and muda (tauri-apps/muda#369 can take the same shape), plus tray-icon for apps that use it.

Happy to restructure this PR that way if that's the direction. The port is done, so what's left is layout and plumbing.

FabianLars:
disregard my feature flag idea. i was overruled.

johncarmack1984:
No worries on my end; the backend is identical either way, so nothing here is wasted. Feel free to take this PR whole, in parts, or not at all, whatever fits the v3 shape. Happy to grab whatever other gtk4 migrations are needed to get tauri past gtk3, whether it's tauri v2.x or v3+

HigherOrderLogic:
> Thanks for the PR but I should let you know that I am working on alternative that uses current winit code base at [tauri-apps/winit-gtk4@`amr`/feat/gtk4](https://github.com/tauri-apps/winit-gtk4/tree/amr/feat/gtk4?rgh-link-date=2026-07-03T22%3A46%3A38Z)

Hi, so what's the current status on this? Seems like the branch has been deleted, so I assume the current code base is on `master` branch?

Ranrar:
No it lives here at this point https://github.com/tauri-apps/wry/pull/1767 feel free to try it

PR #1261, opened 2026-07-09
fix(ios): emit WindowEvent::Destroyed when a scene disconnects

### LINKED ISSUE #1260: [bug] iOS does not implement sceneDidDisconnect
**Describe the bug**
When iOS disconnects a background scene via `sceneDidDisconnect:` (e.g., under memory pressure) or the user swipes a window away, the corresponding WebviewWindow is not cleaned up. This causes `app.webview_windows().len()` to never decrease and per-window resources aren't released.

**Steps To Reproduce**
Steps to reproduce the behavior. It **must** use tao directly and not Tauri.

**Expected behavior**
When the platform destroys a scene/window, the corresponding WebviewWindow should be automatically destroyed, appropriate events should be emitted, and the window count should decrement. This ensures per-window resources are cleaned up properly.

**Additional context**
Patching tao's `sceneDidDisconnect` to emit a custom event bypasses Tauri's window management, leaving `webview_windows()` out of sync. The issue manifests when the platform tears down a scene, not when the app explicitly closes a window.

polw1:
closes #1260

Implements `sceneDidDisconnect:` on the iOS scene delegate, which was previously a no-op.

When the system reclaims a scene, we now emit a `WindowEvent::Destroyed` for each of the scene's windows, so consumers are notified that the windows are gone.

### DIFF
--- a/src/platform_impl/ios/scene.rs
+++ b/src/platform_impl/ios/scene.rs
@@ -80,6 +80,10 @@ define_class!(
       }
     }
 
+    // iOS may disconnect a scene because the user closed it, or because
+    // the system discarded a backgrounded scene to free resources. A
+    // restored scene goes through `scene:willConnectToSession:` again
+    // and gets a new `UIWindow` and `WindowId`.
     #[unsafe(method(sceneDidDisconnect:))]
     fn sceneDidDisconnect(&self, scene: &UIScene) {
       unsafe {
--- a/src/platform_impl/ios/scene.rs
+++ b/src/platform_impl/ios/scene.rs
@@ -87,8 +87,19 @@ define_class!(
     #[unsafe(method(sceneDidDisconnect:))]
     fn sceneDidDisconnect(&self, scene: &UIScene) {
       unsafe {
+        if app_state::is_terminated() {
+          log::debug!("ignoring `sceneDidDisconnect` after application termination");
+          return;
+        }
+
         if let Some(window_scene) = scene.downcast_ref::<UIWindowScene>() {
-          for window in window_scene.windows() {
+          let windows = window_scene.windows();
+
+          if windows.is_empty() {
+            log::debug!("`sceneDidDisconnect` called for a `UIWindowScene` with no windows; no `WindowEvent::Destroyed` events were emitted");
+          }
+
+          for window in windows {
             app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::WindowEvent {
               window_id: RootWindowId(window.into()),
               event: WindowEvent::Destroyed,
--- a/src/platform_impl/ios/scene.rs
+++ b/src/platform_impl/ios/scene.rs
@@ -93,19 +93,11 @@ define_class!(
         }
 
         if let Some(window_scene) = scene.downcast_ref::<UIWindowScene>() {
-          let windows = window_scene.windows();
-
-          if windows.is_empty() {
+          if window_scene.windows().is_empty() {
             log::debug!("`sceneDidDisconnect` called for a `UIWindowScene` with no windows; no `WindowEvent::Destroyed` events were emitted");
           }
-
-          for window in windows {
-            app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::WindowEvent {
-              window_id: RootWindowId(window.into()),
-              event: WindowEvent::Destroyed,
-            }));
-          }
         }
+        handle_scene_window_events(scene, || WindowEvent::Destroyed);
       }
     }
 
--- a/src/platform_impl/ios/scene.rs
+++ b/src/platform_impl/ios/scene.rs
@@ -81,7 +81,18 @@ define_class!(
     }
 
     #[unsafe(method(sceneDidDisconnect:))]
-    fn sceneDidDisconnect(&self, _scene: &UIScene) {}
+    fn sceneDidDisconnect(&self, scene: &UIScene) {
+      unsafe {
+        if let Some(window_scene) = scene.downcast_ref::<UIWindowScene>() {
+          for window in window_scene.windows() {
+            app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::WindowEvent {
+              window_id: RootWindowId(window.into()),
+              event: WindowEvent::Destroyed,
+            }));
+          }
+        }
+      }
+    }
 
     #[unsafe(method(sceneDidBecomeActive:))]
     fn sceneDidBecomeActive(&self, scene: &UIScene) {
--- a/src/platform_impl/ios/scene.rs
+++ b/src/platform_impl/ios/scene.rs
@@ -49,7 +49,13 @@ pub unsafe fn multiple_scenes_enabled() -> bool {
 
 unsafe fn handle_scene_window_events(scene: &UIScene, event: impl Fn() -> WindowEvent<'static>) {
   if let Some(window_scene) = scene.downcast_ref::<UIWindowScene>() {
-    for window in window_scene.windows() {
+    let windows = window_scene.windows();
+
+    if windows.is_empty() {
+      log::debug!("scene has no windows; no window events were emitted");
+    }
+
+    for window in windows {
       app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::WindowEvent {
         window_id: RootWindowId(window.into()),
         event: event(),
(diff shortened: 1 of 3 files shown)


src/platform_impl/ios/scene.rs
```diff
@@ -70,7 +70,18 @@ define_class!(
     }
 
     #[unsafe(method(sceneDidDisconnect:))]
```
velocitysystems, line 93: `sceneDidDisconnect:` isn't only called when the user closes a window. iOS also calls it when it discards a backgrounded scene to free resources. If that scene comes back later, `scene:willConnectToSession:` creates a new `UIWindow` (and a new `WindowId`). So mapping this to `WindowEvent::Destroyed` makes sense, but unlike on desktop platforms it doesn't necessarily mean the user closed the window.

Could we add a comment explaining this? It would help future users understand that `Destroyed` on iOS doesn't imply user intent, and that a restored scene gets a new `WindowId` instead of reusing the old one.

velocitysystems (changes requested):

velocitysystems:
Thanks, @polw1. @FabianLars this would be a good candidate for 0.36.0 once the comment above is addressed.

src/platform_impl/ios/scene.rs
polw1: That makes sense, thanks.

src/platform_impl/ios/scene.rs
```diff
@@ -69,8 +69,23 @@ define_class!(
...
+    // iOS may disconnect a scene because the user closed it, or because
+    // the system discarded a backgrounded scene to free resources. A
+    // restored scene goes through `scene:willConnectToSession:` again
+    // and gets a new `UIWindow` and `WindowId`.
     #[unsafe(method(sceneDidDisconnect:))]
-    fn sceneDidDisconnect(&self, _scene: &UIScene) {}
+    fn sceneDidDisconnect(&self, scene: &UIScene) {
+      unsafe {
+        if let Some(window_scene) = scene.downcast_ref::<UIWindowScene>() {
+          for window in window_scene.windows() {
```
velocitysystems: Has this been verified on a real device for the system-reclaim case (not just user-initiated scene close)? `UIWindowScene.windows `isn't documented to remain populated by the time `sceneDidDisconnect:` runs — Apple's template stores `var window: UIWindow?` on the delegate for exactly this reason — and if the array is already empty here, the loop silently does nothing and #1260 wouldn't actually be fixed for that path.

If it does turn out to be reliable, could we add a debug log for the empty case so a regression is visible rather than silent?

velocitysystems: The termination path in view.rs filters to _Tao-owned_ windows (isKindOfClass: `TaoUIWindow`) before emitting `Destroyed`, but this loop emits for **every window** in the scene. UIKit can inject internal windows (e.g. UITextEffectsWindow) into a scene, which would produce `Destroyed` events with WindowIds Tao never handed out.

Since consumers may react to `Destroyed` (window-count/exit logic), should we apply the same filter here?

src/platform_impl/ios/scene.rs
```diff
@@ -69,8 +69,23 @@ define_class!(
...
+            app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::WindowEvent {
```
velocitysystems: `handle_nonuser_event` panics if the app state is already `Terminated`. It's likely UIKit never delivers `sceneDidDisconnect:` after `applicationWillTerminate`:, but since this is a teardown-path callback and the ordering isn't firmly documented, would an early-return guard be worth adding for robustness?

It would also make a duplicate `Destroyed` with the `will_terminate` path impossible regardless of ordering.

velocitysystems (changes requested):
Thanks @polw1. Took a more detailed look and a few observations for your consideration.

src/platform_impl/ios/scene.rs
polw1: Thanks for raising this, and sorry for the delay in replying.

I tested this on a physical iPad with multiple windows open and under memory pressure.

After `sceneDidEnterBackground`, it was common to see the WebView content being discarded and then reloaded when the window returned to the foreground.

In one test, while the app had several windows open and was running in non-full-screen mode alongside other applications, the system terminated the process and relaunched the app with only its default window.

I could not reproduce a case where the system disconnected an individual scene while keeping the app process alive. Therefore, I was unable to verify whether `UIWindowScene.windows` remains populated when `sceneDidDisconnect` is called.

However, I think we should add a debug log here. Thanks.

polw1: I wonder if this filtering should be left to consumers. 

For example, `tauri-runtime-wry` already ignores events for window IDs that are not registered in its `window_id_map`,
```rust
// crates/tauri-runtime-wry/src/lib.rs:4204
if let Some(window_id) = window_id_map.get(&window_id) {
...
```

 so UIKit-owned windows should not affect Tauri's window state.

What do you think ?

src/platform_impl/ios/scene.rs
polw1: I tried to reason about this, but I don't think we can observe this ordering naturally. 

I could only reproduce the panic by manually invoking `applicationWillTerminate:` followed by `sceneDidDisconnect:`.

But since we don't control UIKit, and as you mentioned, the ordering isn't perfectly documented, and Tao already tracks termination through `AppStateImpl::Terminated`, I agree that a local guard here is a small and reasonable improvement.

Thanks

src/platform_impl/ios/scene.rs
```diff
@@ -80,8 +80,34 @@ define_class!(
...
+
+        if let Some(window_scene) = scene.downcast_ref::<UIWindowScene>() {
+          let windows = window_scene.windows();
+
+          if windows.is_empty() {
+            log::debug!("`sceneDidDisconnect` called for a `UIWindowScene` with no windows; no `WindowEvent::Destroyed` events were emitted");
+          }
+
+          for window in windows {
+            app_state::handle_nonuser_event(EventWrapper::StaticEvent(Event::WindowEvent {
```
FabianLars: if you pull in the dev branch there's now a `handle_scene_window_events` helper function for this

velocitysystems: @polw1 Please see @FabianLars's comment. Can you please update to use the existing helper function?

velocitysystems (changes requested):

src/platform_impl/ios/scene.rs
```diff
@@ -80,8 +80,26 @@ define_class!(
...
+        if app_state::is_terminated() {
+          log::debug!("ignoring `sceneDidDisconnect` after application termination");
+          return;
+        }
+
+        if let Some(window_scene) = scene.downcast_ref::<UIWindowScene>() {
+          if window_scene.windows().is_empty() {
+            log::debug!("`sceneDidDisconnect` called for a `UIWindowScene` with no windows; no `WindowEvent::Destroyed` events were emitted");
+          }
+        }
```
velocitysystems: Small thing, non-blocking: this block does the `UIWindowScene` downcast and the `windows()` call purely to check emptiness, and then `handle_scene_window_events` immediately repeats both.

Would it be cleaner to fold the emptiness log into the helper, or to do a single downcast here and pass the array through?

src/platform_impl/ios/scene.rs
```diff
@@ -80,8 +80,26 @@ define_class!(
...
+        handle_scene_window_events(scene, || WindowEvent::Destroyed);
```
velocitysystems, line 101: @polw1 Apologies — I missed your reply when I last looked at this thread.

Your reasoning makes sense. I followed the reference you linked, and `tauri-runtime-wry`'s `Event::WindowEvent` dispatch already gates on `window_id_map.get(&window_id)`, so unknown window IDs are simply ignored rather than mishandled.

Also, on `dev`, the helper already emits unfiltered `Focused(false)`, `Suspended`, and `Resumed` events. Filtering on `TaoUIWindow` at this one call site would make it inconsistent with the other cases. If we decide we want that filter, it should be added in the helper as a separate change.

velocitysystems (changes requested):
Thanks @polw1. @FabianLars do you have any further observations?

src/platform_impl/ios/scene.rs
FabianLars: moving the log sounds reasonable. i guess that's also helpful for the other events using this helper function

polw1: I moved the log into `handle_scene_window_events`. I also went ahead and updated sceneDidBecomeActive, since there didn’t seem to be a reason for it not to use the helper as well

velocitysystems: Thanks @polw1. @FabianLars Are you happy with these changes?

src/platform_impl/ios/scene.rs
polw1: no worries, I agree

src/platform_impl/ios/scene.rs
FabianLars: yep, lgtm but no way to test ios stuff today but i don't think that's needed here.

FabianLars (approved):

PR #1264, opened 2026-07-13
fix(windows): release window state lock before taskbar update (fix: #1263)

### LINKED ISSUE #1263: Windows: permanent main-thread deadlock in TaskbarCreated handler — window_state mutex held across ITaskbarList COM call
## Environment
- tao 0.35.3 (latest), via Tauri v2
- Windows 11 (10.0.26200), tray-icon app with two windows (one with `skipTaskbar: true`)

## Symptom
The app permanently freezes ("Not Responding") whenever Explorer restarts — which
also reliably happens on **resume from sleep + re-login** (especially with an
external-monitor-only setup, where the taskbar is recreated during unlock).
Once frozen, the app never recovers and must be killed.

## Root cause (from a minidump of the hung process)
In `src/platform_impl/windows/event_loop.rs` (0.35.3, lines ~2250-2253):

```rust
} else if msg == *S_U_TASKBAR_RESTART {
    let window_state = subclass_input.window_state.lock();      // lock held...
    let _ = set_skip_taskbar(window, window_state.skip_taskbar); // ...across a COM call
}
```

`set_skip_taskbar` calls `ITaskbarList::AddTab/DeleteTab`, a **cross-process COM
call into Explorer** that internally pumps incoming sent messages
(`SendMessageW` → `NtUserMessageCall`). While Explorer is restarting it can
broadcast `TaskbarCreated` more than once; the second broadcast is dispatched
**re-entrantly on the same thread during that pump**, enters the same handler,
and calls `subclass_input.window_state.lock()` again. `parking_lot::Mutex` is
not re-entrant → the main thread deadlocks against itself, permanently.

Minidump stack of the main thread (abridged, symbolicated):

```
ntdll!ZwWaitForAlertByThreadId
parking_lot::raw_mutex::RawMutex::lock_slow
lock_api::mutex::Mutex<tao::...::WindowState>::lock          ← inner, blocked (event_loop.rs:2251)
tao::...::public_window_callback_inner (S_U_TASKBAR_RESTART)
user32!SendMessageW / win32u!NtUserMessageCall               ← re-entrant dispatch during pump
windows::Win32::UI::Shell::ITaskbarList::AddTab
tao::platform_impl::platform::window::set_skip_taskbar        (window.rs:1535)
lock_api::mutex::Mutex<tao::...::WindowState>::lock           ← outer, HOLDING the same lock (event_loop.rs:2251-2252)
tao::...::public_window_callback_inner (S_U_TASKBAR_RESTART)
shell32!Shell_NotifyIconW / SendMessageTimeoutW               (tray re-registration in flight)
```

## Reproduction
1. Run any tao/Tauri app that creates a window (a tray icon makes it easier to hit).
2. Restart `explorer.exe` from Task Manager (repeat a few times if needed —
   the race needs two `TaskbarCreated` deliveries close together).
3. App goes "Not Responding" forever. Resume-from-sleep + re-login triggers the
   same path in real usage.

## Suggested fix
Copy the flag out and release the guard before the COM call:

```rust
} else if msg == *S_U_TASKBAR_RESTART {
    let skip_taskbar = subclass_input.window_state.lock().skip_taskbar;
    let _ = set_skip_taskbar(window, skip_taskbar);
}
```

We have been running this one-line patch via `[patch.crates-io]` and can no
longer reproduce the hang (explorer restarts and sleep/wake are both clean).
Other handlers that hold `window_state` across calls that may pump messages
(e.g. `update_theme` → `try_window_theme`) may deserve the same audit.

seemoer:
## Summary

- Release the window state lock before calling `set_skip_taskbar` when handling the `TaskbarCreated` message.
- Add a patch change file for `tao`.

## Problem

The `TaskbarCreated` handler previously held `window_state` while calling `set_skip_taskbar`.

Updating the taskbar state can synchronously trigger reentrant window handling that attempts to acquire the same lock, potentially causing the application to deadlock and become unresponsive.

The issue is timing-dependent, so restarting Explorer or resuming from sleep does not reproduce it every time.

## Fix

Copy `skip_taskbar` while briefly holding the lock, then release the lock before calling `set_skip_taskbar`.

## Testing

- `cargo test`
- `git diff --check`

Fixes #1263

Related upstream fix: rust-windowing/winit#4563

### DIFF
--- a/.changes/taskbar-created-lock.md
+++ b/.changes/taskbar-created-lock.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Release the window state lock before updating taskbar visibility on Windows to avoid a reentrant `TaskbarCreated` deadlock.
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -2195,8 +2195,8 @@ unsafe fn public_window_callback_inner<T: 'static>(
         update_theme(userdata, window, false);
         result = ProcResult::Value(LRESULT(0));
       } else if msg == *S_U_TASKBAR_RESTART {
-        let window_state = userdata.window_state.lock();
-        let _ = set_skip_taskbar(window, window_state.skip_taskbar);
+        let skip_taskbar = userdata.window_state.lock().skip_taskbar;
+        let _ = set_skip_taskbar(window, skip_taskbar);
       }
     }
   };


Legend-Master (approved):
Thanks!

PR #1265, opened 2026-07-14
chore(deps): relax `windows` versions to `=0.61, <=0.62`

Legend-Master:
Since we have no public types containing `windows` types, this should be safe to do.

Alternatively, we wait for this minor released and going straight to 0.62 and bump MSRV in next minor

### DIFF
--- a/.changes/windows-0.61-or-0.62.md
+++ b/.changes/windows-0.61-or-0.62.md
@@ -1,0 +1,5 @@
+---
+"tao": minor
+---
+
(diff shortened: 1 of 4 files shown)


FabianLars (approved):
> Alternatively, we wait for this minor released and going straight to 0.62 and bump MSRV in next minor

i kinda prefer that. feels safer. but your call

Legend-Master:
Actually, since we have both `windows` and `windows-core`, although very unlikely, if cargo selects say 0.61 for `windows` and 0.62 for `windows-core`, it will break the compilation. Maybe locking to 0.62 next minor then.

PR #1266, opened 2026-07-14
feat(android): initialize ndk-context with application Context

### LINKED ISSUE #1220: Android: please initialize ndk-context
In releases prior to the current one, Tao initialized the `ndk-context` with the Android MainActivity's context. This initialization was relied on by many Tauri-based plugins and apps (e.g., any that use the [keyring ecosystem](https://github.com/open-source-cooperative/keyring-rs/wiki/Keyring), such as [the keyring-demo app](https://github.com/open-source-cooperative/keyring-demo/releases). With the latest release, this initialization has gone away, requiring apps to alter the Tauri auto-generated Android stubs in order to initialize it. It would be great to get it back.

I have been told that the reason for this change was Tauri's move to support multiple windows. Since there is a unique ndk-context, I can understand why the decision was made not to initialize it with a specific activity context. But there is a unique _application_ context in every Android app, and it's global to the app. I would like to request that Tao continue to initialize the ndk-context (as it did before) but to do so with the _application_ context rather than the MainActivity context.

An alternative approach would be to extend the ndk-context crate to support multiple named contexts, and for Tao to initialize a separate named ndk-context in each activity. Since there doesn't seem to be active support for the ndk-context crate at this time, that doesn't seem to be a viable alternative. In addition, it would require a lot of changes in client code to pick up the naming of contexts, whereas initializing the global application context is backward-compatible with existing code (or at least any code that only uses application context data and not activity-specific data).

See open-source-cooperative/android-native-keyring-store#21 for an example of how this lack of initialization has affected existing modules and what developers have to do to work around it.

FabianLars:
I am sadly not exactly an android expert so this may not be the best approach. I don't like that it's set in onActivityCreate but i didn't want to change any function arguments again and that was the earliest callback with a jobject we can use i found.

Closes #1220

### DIFF
--- a/src/platform_impl/android/ndk_glue.rs
+++ b/src/platform_impl/android/ndk_glue.rs
@@ -320,7 +320,24 @@ pub enum WindowEvent {
   Destroyed,
 }
 
-pub unsafe fn create(_env: JNIEnv, _: JClass, _: JObject, main: fn()) {
+pub unsafe fn create(mut env: JNIEnv, _: JClass, jobject: JObject, main: fn()) {
+  let vm = env.get_java_vm().unwrap();
+  let application = env
+    .call_method(
+      jobject,
+      "getApplication",
+      "()Landroid/app/Application;",
+      &[],
+    )
+    .unwrap()
+    .l()
+    .unwrap();
+  let application = env.new_global_ref(application).unwrap();
+  ndk_context::initialize_android_context(
+    vm.get_java_vm_pointer() as *mut _,
+    application.as_obj().as_raw() as *mut _,
+  );
+
   let logpipe = {
     let mut logpipe: [RawFd; 2] = Default::default();
     libc::pipe(logpipe.as_mut_ptr());
--- a/src/platform_impl/android/ndk_glue.rs
+++ b/src/platform_impl/android/ndk_glue.rs
@@ -215,7 +215,7 @@ pub type ActivityId = i32;
 pub(crate) static CONTEXTS: Lazy<Mutex<BTreeMap<ActivityId, AndroidContext>>> =
   Lazy::new(Default::default);
 // Keeping a reference so we can store it in ndk-context
-static APPLICATION_CONTEXT: Lazy<Mutex<Option<GlobalRef>>> = Lazy::new(Default::default);
+static APPLICATION_CONTEXT: OnceCell<GlobalRef> = OnceCell::new();
 static WINDOW_MANAGER: Lazy<Mutex<BTreeMap<ActivityId, GlobalRef>>> = Lazy::new(Default::default);
 pub(crate) static ACTIVITY_CREATED_SENDERS: Lazy<Mutex<BTreeMap<ActivityId, Sender<()>>>> =
   Lazy::new(Default::default);
@@ -392,8 +392,7 @@ pub unsafe fn onActivityCreate(
 
   let vm = env.get_java_vm().unwrap();
 
-  let mut app_context = APPLICATION_CONTEXT.lock().unwrap();
-  if app_context.is_none() {
+  APPLICATION_CONTEXT.get_or_init(|| {
     let application = jni_call_method!(
       env,
       &activity,
@@ -407,8 +406,8 @@ pub unsafe fn onActivityCreate(
       vm.get_java_vm_pointer() as *mut _,
       application.as_obj().as_raw() as *mut _,
     );
-    *app_context = Some(application);
-  }
+    application
+  });
 
   let activity_id = jni_call_method!(env, &activity, "getId", "()I", i).unwrap();
 
--- a/src/platform_impl/android/ndk_glue.rs
+++ b/src/platform_impl/android/ndk_glue.rs
@@ -214,6 +214,8 @@ pub type ActivityId = i32;
 
 pub(crate) static CONTEXTS: Lazy<Mutex<BTreeMap<ActivityId, AndroidContext>>> =
   Lazy::new(Default::default);
+// Keeping a reference so we can store it in ndk-context
+static APPLICATION_CONTEXT: Lazy<Mutex<Option<GlobalRef>>> = Lazy::new(Default::default);
 static WINDOW_MANAGER: Lazy<Mutex<BTreeMap<ActivityId, GlobalRef>>> = Lazy::new(Default::default);
 pub(crate) static ACTIVITY_CREATED_SENDERS: Lazy<Mutex<BTreeMap<ActivityId, Sender<()>>>> =
   Lazy::new(Default::default);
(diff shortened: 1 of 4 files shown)


wuwbobo2021:
Well, thank you for your support! I just thought about having a `tauri_plugin_ndk_context` that initializes `ndk_context` on loading before realizing this PR.

It looks like the application context's `GlobalRef` can be simply leaked by `std::mem:forget` instead of storing in `APPLICATION_CONTEXT`.

(Edited, deleted my wrong comments)

FabianLars:
The Option is just there so we can easily check if we ran this logic already since onActivityCreate may be invoked multiple times. Now that i write it out, i could have used Lazy::get instead.

> then the application context's GlobalRef can be simply leaked by std::mem:forget, or turned to an unmanaged, owned raw pointer by Global::into_raw in the new jni version 0.22 (which has fixed a few soundness issues in jni 0.21).

I started with an AtomicBool and tried to leak the GlobalRef but nothing worked, always got an error about invalid references or similar. It may have been another issue though, i didn't check again after i got this approach working.
I do not want to call ndk-context on every onActivityCreate call and add a panic handler so i'd keep the AtomicBool similar to `DID_RESUME`

We can also delay this PR to the next tao minor version where we can update the jni crate.

FabianLars:
Oh and thanks for the review of course!

wuwbobo2021:
> Oh and thanks for the review of course!

Not at all. My mind just got stuck that I didn't realize that the checking for `Option` prevents `onActivityCreate` from initializing `ndk_context` for multiple times, or, I just forgot about this while writing my comments.

> I started with an `AtomicBool` and tried to leak the `GlobalRef` but nothing worked, always got an error about invalid references or similar. It may have been another issue though, i didn't check again after i got this approach working.

Maybe a `OnceLock` or `once_cell::OnceCell` can be used for this purpose? Well, I know this is not a real problem, just ignore this.

wuwbobo2021:
> We can also delay this PR to the next tao minor version where we can update the jni crate.

Excuse me, I don't mean that soundness issues of `jni` 0.21 are related to global reference handling; they are probably untriggered in the current Tauri version. Bumping `jni` to 0.22 will break the current MSRV of Tauri crates.

FabianLars:
> Excuse me, I don't mean that soundness issues of jni 0.21 are related to global reference handling; they are probably untriggered in the current Tauri version.

ah got it, into_raw sounds interesting though

> Bumping jni to 0.22 will break the current MSRV of Tauri crates.

That's planned for tauri 2.12. We will release one more tao minor with current msrv and next one can raise it.

src/platform_impl/android/ndk_glue.rs
```diff
@@ -388,6 +390,26 @@ pub unsafe fn onActivityCreate(
   let intent =
     jni_call_method!(env, &activity, "getIntent", "()Landroid/content/Intent;", l).unwrap();
 
+  let vm = env.get_java_vm().unwrap();
+
+  let mut app_context = APPLICATION_CONTEXT.lock().unwrap();
+  if app_context.is_none() {
+    let application = jni_call_method!(
```
Legend-Master, line 396: Not an Android expert either, is the `application` we get from any activity the same inside a single process? This seems to be initializing the static with the first activity right?

> https://developer.android.com/reference/android/content/ContextWrapper#getApplicationContext()

It looks like it but I'm not sure

src/platform_impl/android/ndk_glue.rs
```diff
@@ -214,6 +214,8 @@ pub type ActivityId = i32;
 
 pub(crate) static CONTEXTS: Lazy<Mutex<BTreeMap<ActivityId, AndroidContext>>> =
   Lazy::new(Default::default);
+// Keeping a reference so we can store it in ndk-context
+static APPLICATION_CONTEXT: Lazy<Mutex<Option<GlobalRef>>> = Lazy::new(Default::default);
```
Legend-Master: As suggested by @wuwbobo2021, `OnceCell`/`OnceLock` might be a better fit here.

Legend-Master (commented):
The static you used seems better than the `std::mem:forget` suggested by @wuwbobo2021 that at least we know where did things go

src/platform_impl/android/ndk_glue.rs
FabianLars: yeah i mindlessly copied the other Lazy uses we have there :/

src/platform_impl/android/ndk_glue.rs
FabianLars: > Not an Android expert either, is the application we get from any activity the same inside a single process?

As far as i understand it, yes.

> This seems to be initializing the static with the first activity right?

If you mean that it's registered once the first activity is created, then yes. I couldn't make the other callbacks work without changing how they're called in kotlin (if at all)

If you mean whether we set an Application Context or Activity Context in the Lazy static then it's the Application Context. The text you linked is very confusingly phrased imo, but it gets clearer if you remove the single part so you end up with
> Return the context of the single, global Application object of the current process. This generally should only be used if you need a Context whose lifecycle is separate from the current context, that is tied to the lifetime of the process rather than the current component.
>
> Consider for example how this interacts with [registerReceiver(BroadcastReceiver,IntentFilter)](https://developer.android.com/reference/android/content/Context#registerReceiver(android.content.BroadcastReceiver,%20android.content.IntentFilter)):
>
>    If used from the Context returned here, the receiver is being registered with the global state associated with your application. Thus it will never be unregistered for you. This is necessary if the receiver is associated with static data, not a particular component. However using the ApplicationContext elsewhere can easily lead to serious leaks if you forget to unregister, unbind, etc.

No activity context at play here - those are a few lines further down btw.

Legend-Master (approved):

src/platform_impl/android/ndk_glue.rs
```diff
@@ -388,6 +390,25 @@ pub unsafe fn onActivityCreate(
...
+      env,
+      &activity,
+      "getApplicationContext",
+      "()Landroid/content/Context;",
+      l
+    )
+    .unwrap();
+    let application = env.new_global_ref(application).unwrap();
+    ndk_context::initialize_android_context(
+      vm.get_java_vm_pointer() as *mut _,
```
Legend-Master, line 406: Actually, the life time of this doesn't seem right as the pointer come from the outer `let vm = env.get_java_vm().unwrap();`, is this correct?

FabianLars: so as far as i understood it there's only one vm in jni-rs and it should be valid for the whole time once it was initialized once. The lifetime of get_java_vm is not bound to the env but to the global vm if i read it right.
I think it'll be a tiny bit clearer in jni 0.22 where it leans more into a singleton pattern in some places.

Anyway, i'll do some multi-window testing and see if i can spot any issues.

FabianLars: oof that's hard to test. seems like we never destroy the main activity unless we completely close the app. i assume that's intended cause we have a bunch of shit depend on that at first glance but not sure if that's really the intention or just ductape

FabianLars: having all the shit split over 3 crates (tauri wry tao) makes everything so much harder here

wuwbobo2021: > so as far as i understood it there's only one vm in jni-rs and it should be valid for the whole time once it was initialized once. The lifetime of get_java_vm is not bound to the env but to the global vm if i read it right.

- According to <https://docs.oracle.com/en/java/javase/21/docs/specs/jni/invocation.html> creation of multiple VMs in a single process is not supported even on PC-side runtimes;

- According to <https://developer.android.com/ndk/guides/jni-tips#javavm-and-jnienv>, it is safe to cache the JVM pointer in static storage in a native library. Some claims that the JVM is never destroyed in an Android process, while the process itself may be killed; the native library instance (along with the global static storages defined in crates like `ndk_context` and `tao`) is destroyed when the process running the JVM is killed.

Also, the description of "cached process" in <https://developer.android.com/guide/components/activities/process-lifecycle> might be relevant here.

brotskydotcom:
@FabianLars Do we know which Tauri release (if any) contains this change? When I build my app with Tauri 2.11.5, I don't see an initialized ndk-context (and have to initialize it myself).

FabianLars:
2.12 I think. No ETA but soon

brotskydotcom:
Thanks for replying!

PR #1270, opened 2026-07-17
fix(ios): handle layout while view is detached

### LINKED ISSUE #1269: [bug] iOS `layoutSubviews` aborts when Tao’s UIView is detached (on device rotation)
**Describe the bug**

On iOS, Tao’s `layoutSubviews` callback asserts that the root `UIView` always belongs to a `UIWindow`. In practice, UIKit can temporarily detach and lay out that view during device rotation or window/view-controller transitions, particularly when Tao is embedded in an existing UIKit application hierarchy. In that state, `UIView.window` is nil, causing the assertion to panic and terminate the process with `SIGABRT`.

**Steps To Reproduce**

1. Create an iOS executable depending `tao` and `objc = "0.2"`, package it as an iOS application without a scene manifest.
2. Use the following Tao event-loop application:
```rust
 use objc::{msg_send, runtime::Object, sel, sel_impl};
 use tao::{
   event::Event,
   platform::ios::WindowExtIOS,
   window::WindowBuilder,
 };

 fn main() {
   let window = WindowBuilder::new().build(&event_loop).unwrap();
   let mut triggered = false;

     if triggered || !matches!(event, Event::MainEventsCleared) {
       return;
     }

     unsafe {
       let view = window.ui_view().cast::<Object>();
       let attached_window: *mut Object = msg_send![view, window];
       if attached_window.is_null() {
         return;
       }

       triggered = true;
       let _: () = msg_send![view, removeFromSuperview];
       let _: () = msg_send![view, setNeedsLayout];
       let _: () = msg_send![view, layoutIfNeeded];
     }
   });
 }
```
3. The process aborts in Tao’s layout_subviews callback at assert!(!window.is_null()).

Sufficient plist:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN"
  "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
  <key>CFBundleExecutable</key>
  <string>tao-detached-layout</string>
  <key>CFBundleIdentifier</key>
  <string>org.example.tao-detached-layout</string>
  <key>CFBundleName</key>
  <string>TaoDetachedLayout</string>
  <key>CFBundlePackageType</key>
  <string>APPL</string>
  <key>CFBundleShortVersionString</key>
  <string>1.0</string>
  <key>CFBundleVersion</key>
  <string>1</string>
  <key>LSRequiresIPhoneOS</key>
  <true/>
  <key>MinimumOSVersion</key>
  <string>16.0</string>
  <key>UIDeviceFamily</key>
  <array>
    <integer>1</integer>
    <integer>2</integer>
  </array>
  <key>UILaunchScreen</key>
  <dict/>
</dict>
</plist>
```

**Expected behavior**

Laying out a detached Tao view should not terminate the application. When no window is available, Tao should finish the superclass layout and defer window-dependent frame calculation and resize-event delivery until the view is attached again.

**Screenshots**

OS: iPadOS 26.5.2
Rustc: 1.95.0
Tao: 0.34.8; also present on dev

**Additional context**

The panic originates from `layout_subviews` in the iOS view implementation (https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/ios/view.rs#L133-L140). The same implementation already handles a null window defensively in `set_content_scale_factor`.

Replacing the assertion with an early return when `window.is_null()` prevents the abort while preserving the existing behavior for attached views.

ahirner:
Avoid querying window geometry while UIKit is laying out detached view. View will report geometry after reattachment.
fixes #1269

### DIFF
--- a/src/platform_impl/ios/view.rs
+++ b/src/platform_impl/ios/view.rs
@@ -136,7 +136,11 @@ unsafe fn get_view_class(root_view_class: &'static Class) -> &'static Class {
         let () = msg_send![super(object, superclass), layoutSubviews];
 
         let window: id = msg_send![object, window];
-        assert!(!window.is_null());
+        // UIKit can lay out a view while it is detached during window and view-controller
+        // transitions. There is no window ID or screen geometry to report until reattachment.
+        if window.is_null() {
+          return;
+        }
         let window_bounds: CGRect = msg_send![window, bounds];
         let screen: id = msg_send![window, screen];
         let screen_space: id = msg_send![screen, coordinateSpace];
--- a/.changes/ios-detached-layout.md
+++ b/.changes/ios-detached-layout.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Prevent iOS apps from aborting when UIKit lays out a Tao view while it is temporarily detached from its window.


src/platform_impl/ios/view.rs
```diff
@@ -136,7 +136,11 @@ unsafe fn get_view_class(root_view_class: &'static Class) -> &'static Class {
         let () = msg_send![super(object, superclass), layoutSubviews];
 
         let window: id = msg_send![object, window];
-        assert!(!window.is_null());
+        // UIKit can lay out a view while it is detached during window and view-controller
+        // transitions. There is no window ID or screen geometry to report until reattachment.
+        if window.is_null() {
```
FabianLars, line 141: we really need some tracing or something in the ios code in the future. all these assert/assert_eq calls are giving me anxiety.

FabianLars:
Thanks :) could you also add a small changefile please? (check .changes dir)

FabianLars (approved):

PR #1274, opened 2026-07-19
fix(windows): use DefWindowProcW for thread event target

grig2277:
Commit 5f209bd8f6d087ecc1aad6610ca798e378a179cd registered `thread_event_target_callback` directly as `lpfnWndProc` but left three `DefSubclassProc` calls. A regular `WNDPROC` must delegate to `DefWindowProcW`.

The old default handling returned `FALSE` for `WM_QUERYENDSESSION` and vetoed Windows shutdown in https://github.com/cjpais/Handy/issues/1710.

Verification:
- `cargo check --lib --target x86_64-pc-windows-gnu` passed.
- A manual Windows 11 build 22631 probe changed the `Tao Thread Event Target` `WM_QUERYENDSESSION` response from `0x0` before the patch to `0x1` after it.

### DIFF
--- a/src/platform_impl/windows/event_loop.rs
+++ b/src/platform_impl/windows/event_loop.rs
@@ -33,8 +33,7 @@ use windows::{
       Controls::{self as win32c, HOVER_DEFAULT},
       Input::{KeyboardAndMouse::*, Pointer::*, Touch::*, *},
       Shell::{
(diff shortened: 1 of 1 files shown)


Legend-Master (approved):
Nice catch thanks! Please run `cargo fmt` and sign your commits and I'll merge this.

Legend-Master (approved):
Thanks!

PR #1288, opened 2026-07-28
refactor(android)!: rename lifecycle bindings

Legend-Master:
~~Waiting on #1209~~


FabianLars:
do we have a better name for https://github.com/Legend-Master/tao/blob/ca69bcc2fc310472beaf678701f7a72a1411ea62/src/platform_impl/android/ndk_glue.rs#L330 ? maybe a bit confusing to have onCreate and create. `createApp` or `init`?

Legend-Master:
Ah, I sorta see why it was renamed to `onActivityCreate` now that `create` was kept in #1154, let me take a closer look at why are we having so many create functions, looking at wry's `wryCreate`, `onCreate`, `create` 😂

Legend-Master:
Ok so basically wry is doing what tao is supposed to do on Android so it calls `wryCreate` for itself and also calls `onCreate` and `create` for tao. I'm still super confused about the differences between `onCreate` and `create` though that one is call inside `WryActivity` while the other is in `WryLifecycleObserver` attached in `WryActivity.onCreate`

FabianLars-crabnebula:
I assume the idea was to have create be called as early as possible (to register the logger whatever as early as possible) but i agree that that doesn't make much sense here if it's also simply called in activity onCreate

Legend-Master:
I don't understand why I can't retarget the branches to dev now, I remember it used to retarget dev after the targeted PR gets merged 😰

Edit: fixed by restoring the old branch first...

Legend-Master:
@FabianLars have you tried multi-window setup on Android before? The current code looks quite weird that `create` is called on every `WryActivity.onCreate` which calls the main function that maps to the tauri app's mobile entry point. Should this function be called only once?

src/platform_impl/android/ndk_glue.rs
```diff
@@ -61,16 +61,17 @@ const DATA_URL_ENCODING_SET: &AsciiSet = &CONTROLS
...
-///       - `private external fun stop(activity: WryActivity)`
-///       - `private external fun onActivitySaveInstanceState()`
-///       - `private external fun onActivityDestroy(activity: WryActivity)`
-///       - `private external fun onActivityLowMemory()`
+///       - `private external fun onCreate(activity: WryActivity)`
+///       - `private external fun onStart(activity: WryActivity)`
+///       - `private external fun onResume(activity: WryActivity)`
+///       - `private external fun onPause(activity: WryActivity)`
+///       - `private external fun onStop(activity: WryActivity)`
+///       - `private external fun onSaveInstanceState(activity: WryActivity)`
```
FabianLars: this getting the activity but not the instance is a bit awkward imo. since we're not using this callback, maybe remove it and only add it when we actually need it so we know what input we actually need?

similar situation for onLowMemory which isn't used anywhere

FabianLars:
yes i tried it before merging lucas' prs. the changes were a bit too much to approve without testing. I did not notice any issues with the current code but i totally agree it's weird.

still trying to follow lucas' train of thought on why he added that lifecycle thing

Legend-Master:
Also found that we're calling `startActivity` in tao that is required in wry which is also not documented...

src/platform_impl/android/ndk_glue.rs
Legend-Master: > this getting the activity but not the instance is a bit awkward imo. 

Sorry what instance?

I honestly don't know why `onSaveInstanceState` is even here that there's no `onRestoreInstanceState` for it to be useful. Also it doesn't even pass in the bundle for you to store the data which made it completely useless. Maybe removing it?

Legend-Master: > similar situation for onLowMemory which isn't used anywhere

That one seems to go through to the event loop but we never mapped it to anything... I don't mind removing the activity here since it's not tied to the activity.

FabianLars:
And i just re-tested the multiwindow project, seems like i misunderstood how the lifecycle observer works. It looks like once set it uses the same observer instance across all activities, meaning the observer's onCreate is only called once, not on every activity onCreate. So Rust.create and wryCreate are indeed only called once.
(that's kinda what i hoped it does when i added the ndk-context but gave up on using the observer for that before really understanding it all)

Edit: This is true for all of the observer methods we have defined here. They are all called app-wide, not per-activity. I also tested the onpause hook for example when going to background while the main activity is not onscreen.

Edit2: This may be because of https://github.com/tauri-apps/tao/pull/1266#discussion_r3594837986 actually

Legend-Master:
Thanks for testing, that really helped 😂 This is still super confusing I must say that `ProcessLifecycleOwner.get().lifecycle.addObserver(WryLifecycleObserver)` on the new activity doesn't trigger `WryLifecycleObserver.onCreate`?

Edit: I see, https://developer.android.com/reference/androidx/lifecycle/ProcessLifecycleOwner I though it was activity lifecycle owner

Legend-Master:
> Edit: This is true for all of the observer methods we have defined here. They are all called app-wide, not per-activity. I also tested the onpause hook for example when going to background while the main activity is not onscreen.

Which means all of them were wrong and should be fixed in https://github.com/tauri-apps/wry/pull/1720 ?

FabianLars:
>  This is still super confusing I must say that ProcessLifecycleOwner.get().lifecycle.addObserver(WryLifecycleObserver) on the new activity doesn't trigger WryLifecycleObserver.onCreate?

Yeah, that's the part i still don't get either. I don't see any docs saying that it gets ignored in any cases but i guess there's some "normal" java or android thing at play here that notices it's the same instance being added all the time idk. I'll stop thinking about this for now 😂 

> Which means all of them were wrong and should be fixed in https://github.com/tauri-apps/wry/pull/1720 ?

Not sure what exactly you mean. The PR moves Resumed/Suspended from app-wide to per-activity == from Lifecycle to Activity which makes sense now. But i think i just don't get what you mean.

----

What's also fun is that the lifecycle onCreate fires after activity onCreate so this https://github.com/tauri-apps/tao/pull/1288#issuecomment-5102411025 isn't actually the case (or _may_ not be the case)

Legend-Master:
> Yeah, that's the part i still don't get either. I don't see any docs saying that it gets ignored in any cases but i guess there's some "normal" java or android thing at play here that notices it's the same instance being added all the time idk. I'll stop thinking about this for now 😂

The doc says

> You can consider this LifecycleOwner as the composite of all of your Activities, except that [`Lifecycle.Event.ON_CREATE`](https://developer.android.com/reference/androidx/lifecycle/Lifecycle.Event#ON_CREATE) will be dispatched once
> https://developer.android.com/reference/androidx/lifecycle/ProcessLifecycleOwner#:~:text=You%20can%20consider%20this%20LifecycleOwner%20as%20the%20composite%20of%20all%20of%20your%20Activities%2C%20except%20that%20Lifecycle.Event.ON_CREATE%20will%20be%20dispatched%20once

> Not sure what exactly you mean. The PR moves Resumed/Suspended from app-wide to per-activity == from Lifecycle to Activity which makes sense now. But i think i just don't get what you mean.

right right right, that's correct since we didn't move the events yet

src/platform_impl/android/ndk_glue.rs
```diff
@@ -61,16 +61,17 @@ const DATA_URL_ENCODING_SET: &AsciiSet = &CONTROLS
...
 /// 3. the android activity that has external linking for the following functions and calls them:
-///       - `private external fun onActivityCreate(activity: WryActivity)``
-///       - `private external fun start(activity: WryActivity)`
-///       - `private external fun resume(activity: WryActivity)`
-///       - `private external fun pause(activity: WryActivity)`
-///       - `private external fun stop(activity: WryActivity)`
-///       - `private external fun onActivitySaveInstanceState()`
-///       - `private external fun onActivityDestroy(activity: WryActivity)`
-///       - `private external fun onActivityLowMemory()`
+///       - `private external fun onFirstActivityCreate(activity: WryActivity)`
```
FabianLars: we currently don't give it that in wry - we can i think because the lifecycle owner is the activity

if we were i'm thinking about moving the ndk-context thing into here but it sucks that it's called after the activity's onCreate

Legend-Master: > we currently don't give it that in wry - we can i think because the lifecycle owner is the activity

We do, it should be easier to see after https://github.com/tauri-apps/wry/pull/1790

> if we were i'm thinking about moving the ndk-context thing into here but it sucks that it's called after the activity's onCreate

I don't mind it but I'd probably keep it just for the timing, I could make a comment on the reasons if you want

FabianLars: No we don't: https://github.com/tauri-apps/wry/pull/1790/changes#diff-76280dd9be7cec50ab33d778adb7380d01ebfe63a7b5373ee637ef710fe60322R29 - i was talking about the function argument.

FabianLars: > I don't mind it but I'd probably keep it just for the timing, I could make a comment on the reasons if you want

Talking about timing, i wonder if we even should use the lifecycle thing or if we just just call it in onActivity if there's a way to have a variable across all activities to check if it has been called already. The incorrect timing here really puts me off.

Legend-Master: > No we don't: https://github.com/tauri-apps/wry/pull/1790/changes#diff-76280dd9be7cec50ab33d778adb7380d01ebfe63a7b5373ee637ef710fe60322R29 - i was talking about the function argument.

Ah, I see, was testing it in tauri and it worked so didn't notice

> Talking about timing, i wonder if we even should use the lifecycle thing or if we just just call it in onActivity if there's a way to have a variable across all activities to check if it has been called already. The incorrect timing here really puts me off.

Trying something like https://github.com/tauri-apps/wry/pull/1790/changes/11dab6bcc0a472f2bd6555d7f545bdd08fda4232 but I really don't know what java land smokes that if this works. I would honestly say just leave this alone since it didn't cause any problems so far.

Legend-Master: The timing is actually correct I think, we need to run the android entry point after we setup many things.

FabianLars: > Trying something like https://github.com/tauri-apps/wry/commit/11dab6bcc0a472f2bd6555d7f545bdd08fda4232 but I really don't know what java land smokes that if this works. I would honestly say just leave this alone since it didn't cause any problems so far.

That works in the multiwindow example i tested everything else at least

> The timing is actually correct I think, we need to run the android entry point after we setup many things.

we're talking about https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/android/ndk_glue.rs#L330 and https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/android/ndk_glue.rs#L389 right? to me `create` looks like it wants to run first.

I almost want to merge these 2 callbacks, we could do the isFirstActivity check in there instead but i think i'm just going crazy - we can keep everything as is and focus on getting it released instead 😂

Legend-Master: > we're talking about https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/android/ndk_glue.rs#L330 and https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/android/ndk_glue.rs#L389 right? to me `create` looks like it wants to run first.

Yeah, it looks like so but it doesn't and it's probably correct (not semantically, functionally) 😂

I would honestly just leave it to another day, I have documented the timings for now.

FabianLars (approved):
can't test the last commit rn 🤷

Legend-Master:
I tested or else I wouldn't know there's a `tao-macros` bug 😂

FabianLars:
i assumed as much, i just like to document what i did in the approval, hence the `(untested)` comments you may have seen once or twice 😂

PR #1291, opened 2026-07-29
refactor!: bump MSRV to v1.85

Legend-Master:
Reference: https://github.com/tauri-apps/tauri/issues/15579

Don't have time to test if we can change the JNI version yet, we could change it to 1.85 to match https://crates.io/crates/jni if that's possible to upgrade


FabianLars:
Imo we can update to 1.85 either way already. That's the edition2024 boundary right

FabianLars (approved):

Legend-Master:
yeah, that's fair too

FabianLars (approved):

Legend-Master:
I think we have exported a few methods containing JNI types so probably we can't update it in v2 yet for the time being.

FabianLars:
🤮

PR #1295, opened 2026-08-02
feat(macos): support applicationDockMenu: (right-click Dock menu)

thomnico:
## Summary

Adds support for `applicationDockMenu:` — the menu Cocoa shows when the user right-clicks (or Control-clicks / long-presses) the application's Dock icon. Currently unsupported: macOS falls back to its default minimal menu (Show/Hide, Quit) with no way for the app to customize it.

New public API on `EventLoopWindowTargetExtMacOS`:

```rust
fn set_dock_menu(&self, menu: Option<objc2::rc::Retained<objc2_app_kit::NSMenu>>);
```

Same shape as the existing `set_badge_label`/`set_dock_visibility` methods. Can be called at any point after the event loop starts, including from a menu item's own click handler (e.g. to rebuild the menu with an updated checkmark) — the next right-click picks up the change.

## Why a small module instead of an `Event` variant

`applicationDockMenu:` must return its value *synchronously* to Cocoa — there's no round-trip available through tao's event queue the way there is for e.g. `Event::Reopen`. So the menu is stored in a small module-level `Mutex` (`platform_impl/macos/dock_menu.rs`), read directly by the new `application_dock_menu` handler in `app_delegate.rs`. Same main-thread-only invariant as the rest of that file's `id` usage (documented inline).

## Testing

Used this in a real app (menu bar app built on tao 0.36, `EventLoopBuilder::<UserEvent>::with_user_event()`) — right-click on the Dock icon now shows a custom `CheckMenuItem` built with `muda`, synced with the same toggle exposed elsewhere in the app (tray menu, application menu bar). Confirmed working on macOS (tested by the end user, not just compiled).

## Notes

- Added `"NSMenu"` to the `objc2-app-kit` feature list in `Cargo.toml` (was missing, needed for `NSMenu`/`Retained<NSMenu>`).
- `cargo fmt --check` and `cargo check --target aarch64-apple-darwin` pass.
- Happy to add a `dock_menu` example under `examples/` if useful for review — didn't want to guess at what you'd want there without asking first.


thomnico:
Gentle nudge on this one 🙂 — CI is green, there are no conflicts, and it's a small, self-contained addition (+109/−1) with a change file included. Would a maintainer be able to take a look when you have a moment? Happy to rebase onto `dev` or address any feedback (tests, docs, API shape). Thanks for your work on tao!

PR #1296, opened 2026-08-04
fix(windows): flashing title bar on undecorated windows

### LINKED ISSUE #1240: Windows: transparent window flashing title bar on start
https://github.com/user-attachments/assets/af7c733d-79d5-4fb0-aeef-5c3241dbbdd3

_Originally posted by @FabianLars in https://github.com/tauri-apps/tao/issues/1231#issuecomment-4707915728_

#1231 introduced a regression on Windows 11 that the transparent window's title bar will flash on start

Legend-Master:
Fix #1240

This was the result of lots of trial and errors in the past week, I'm not yet sure if this breaks anything so feedbacks very much welcome.

Here's the initial win32 window messages comparison

image

### DIFF
--- a/src/platform_impl/windows/window_state.rs
+++ b/src/platform_impl/windows/window_state.rs
@@ -249,9 +249,6 @@ impl WindowFlags {
     if self.contains(WindowFlags::MINIMIZABLE) {
       style |= WS_MINIMIZEBOX;
     }
-    if self.contains(WindowFlags::VISIBLE) {
-      style |= WS_VISIBLE;
-    }
     if self.contains(WindowFlags::ON_TASKBAR) {
       style_ex |= WS_EX_APPWINDOW;
     }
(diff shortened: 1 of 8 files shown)


src/platform_impl/windows/window_state.rs
```diff
@@ -458,6 +444,23 @@ impl WindowFlags {
         );
       }
     }
+
+    // `SWP_FRAMECHANGED` above must run while the HWND is still hidden. It sends the initial
+    // `WM_NCCALCSIZE` that converts Tao's caption-compatible styles into an undecorated client
+    // frame. Showing first would expose the native caption until that recalculation completed.
```
Legend-Master: Didn't confirm this yet, this was written by codex

Legend-Master:
This doesn't focus the window now... currently the only working commit is d467245

Edit: should be fixed now, the `WS_VISIBLE` must not be set in `SetWindowLongW` or `ShowWindow` wouldn't work

FabianLars (approved):
not much of a code review but no titlebar when running it anymore 🥳

src/platform_impl/windows/window_state.rs
Legend-Master: Can't really confirm this, change to a fact in 3751e68

Legend-Master:
Also seems to reduce the white flashing of the transparent example when launched with lots of mouse movements. The flicker fully disappears if the window is built with `with_no_redirection_bitmap(true)`.

Legend-Master:
I wonder if this should be ported to winit, undecorated transparent window doesn't seem to have the frames at all there...

PR #1298, opened 2026-08-05
fix(android): fix the android_binding macro

nifouprog:
A mismatched meta-variable name in the `android_binding` macro was introduced during the refactoring of https://github.com/tauri-apps/tao/commit/4e7c2f4. This PR fixes the usage of the meta-variable so that the macro becomes usable again (currently, any usage of the macro without a custom tao path fails with `error: no rules expected $`).


Legend-Master (approved):
Nice catch, thanks!

PR #1302, opened 2026-08-07
refactor(windows): only call `set_skip_taskbar` when needed

justanotheranonymoususer:
Reason: If Explorer is not responding, the app will hang too. Can be verified with Process Explorer by suspending explorer.exe.

### DIFF
--- a/.changes/fix-set_skip_taskbar.md
+++ b/.changes/fix-set_skip_taskbar.md
@@ -2,4 +2,4 @@
 "tao": patch
 ---
 
-Skip `set_skip_taskbar` when not hiding. Reason: If Explorer is not responding, the app will hang too. Can be verified with Process Explorer by suspending explorer.exe.
+On Windows, skip `set_skip_taskbar` if `skip_taskbar` is false when creating the window and on taskbar restars.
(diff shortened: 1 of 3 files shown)


Legend-Master (commented):
Since the taskbar is not skipped by default without us calling `ITaskbarList.AddTab` manually, this makes sense to me. I don't know if there're unexpected consequences on this though. cc @amrbashir for some opinions on this.

Could you also add a change file?

> https://github.com/tauri-apps/tao/blob/dev/.changes/readme.md

.changes/fix-set_skip_taskbar.md
```diff
@@ -0,0 +1,5 @@
+---
+"tao": patch
+---
+
+Skip `set_skip_taskbar` when not hiding. Reason: If Explorer is not responding, the app will hang too. Can be verified with Process Explorer by suspending explorer.exe.
```
Legend-Master:
```suggestion
On Windows, skip `set_skip_taskbar` if `skip_taskbar` is false when creating the window and on taskbar restars.
```

The reasoning sounds weird, like it will still hang if you have it set to true

Legend-Master (approved):

Legend-Master:
I'll wait @amrbashir for another week, will merge this if he doesn't respond in time.

amrbashir (approved):

PR #1304, opened 2026-08-08
fix(android): drain EventLoopProxy user events on every poll iteration

Th4ler:
## Summary

On Android, `ALooper_pollAll` can return an fd event (e.g. the `ndk_glue`
event pipe, ident=0) instead of `ALOOPER_POLL_WAKE` when both the wake fd and
an fd are ready in the same epoll batch. In that case `Poll::Wake` is never
reported, the `EventSource::User` arm never drains `self.receiver`, and events
queued via `EventLoopProxy::send_event` sit in the channel forever.

This breaks every IPC response, window getter, and `run_on_main_thread` call
on Android — the app hangs on first invoke because responses never reach the
renderer.

## Fix

Drain `self.receiver` unconditionally after every poll iteration, not only
when `first_event == Some(EventSource::User)`.

Safety:
- If `EventSource::User` already drained above, `try_recv` returns immediately
  on an empty channel.
- User events are independent of the Callback/InputQueue processing order,
  they are dispatched via `Event::UserEvent`.
- The crossbeam receiver is lock-free and `try_recv` is cheap.

## Testing

Tested on Samsung S22 Ultra (Android 16) with a Tauri 2.11.5 app using
tao 0.36.0 + tauri-runtime-wry 2.11.4. Without the fix, the first
`invoke()` hangs indefinitely. With the fix, all invokes resolve in <100ms.

### DIFF
--- a/.changes/android-always-drain-user-events.md
+++ b/.changes/android-always-drain-user-events.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+Fix Android event loop hanging on first IPC call. When the `ndk_glue` event pipe and the wake fd became ready at the same time, `ALooper_pollAll` could return the fd event instead of `ALOOPER_POLL_WAKE`, so the loop never drained the user-event channel and every invoke response sat there forever. The event loop now drains pending user events on every iteration, regardless of what the poll reported.
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -424,31 +424,14 @@ impl<T: 'static> EventLoop<T> {
             }
           */
         }
-        Some(EventSource::User) => {
-          while let Ok(event) = self.receiver.try_recv() {
-            self.call_event_handler(event_handler, control_flow, event::Event::UserEvent(event));
-          }
-        }
-        None => {}
+        _ => {}
       }
 
-      // FIX: Always drain pending user events after every poll iteration,
-      // not only when `first_event == Some(EventSource::User)`.
+      // We need to treat every event as if they were from [`EventLoopProxy::send_event`]
       //
-      // On Android, `ALooper_pollAll` can return an fd event (e.g. the
-      // `ndk_glue` event pipe, ident=0) instead of `ALOOPER_POLL_WAKE` when
-      // both the wake fd and an fd are ready in the same epoll batch. In
-      // that case `Poll::Wake` is never reported and the `EventSource::User`
-      // arm above never drains `self.receiver`, so events queued via
-      // `EventLoopProxy::send_event` sit in the channel forever — breaking
-      // every IPC response, window getter, and `run_on_main_thread` call.
-      //
-      // Draining unconditionally is safe because:
-      //  * If `EventSource::User` already drained above, `try_recv` returns
-      //    immediately on an empty channel.
-      //  * User events are independent of the Callback/InputQueue
-      //    processing order — they are dispatched via `Event::UserEvent`.
-      //  * The crossbeam receiver is lock-free and `try_recv` is cheap.
+      // > All return values may also imply ALOOPER_POLL_WAKE. If you call this in a loop,
+      // > you must treat all return values as if they also indicated ALOOPER_POLL_WAKE.
+      // > https://developer.android.com/ndk/reference/group/looper#alooper_pollonce
       while let Ok(event) = self.receiver.try_recv() {
         self.call_event_handler(event_handler, control_flow, event::Event::UserEvent(event));
       }
--- a/src/platform_impl/android/mod.rs
+++ b/src/platform_impl/android/mod.rs
@@ -432,6 +432,27 @@ impl<T: 'static> EventLoop<T> {
         None => {}
       }
 
+      // FIX: Always drain pending user events after every poll iteration,
+      // not only when `first_event == Some(EventSource::User)`.
+      //
+      // On Android, `ALooper_pollAll` can return an fd event (e.g. the
+      // `ndk_glue` event pipe, ident=0) instead of `ALOOPER_POLL_WAKE` when
+      // both the wake fd and an fd are ready in the same epoll batch. In
+      // that case `Poll::Wake` is never reported and the `EventSource::User`
+      // arm above never drains `self.receiver`, so events queued via
+      // `EventLoopProxy::send_event` sit in the channel forever — breaking
+      // every IPC response, window getter, and `run_on_main_thread` call.
+      //
+      // Draining unconditionally is safe because:
+      //  * If `EventSource::User` already drained above, `try_recv` returns
+      //    immediately on an empty channel.
+      //  * User events are independent of the Callback/InputQueue
+      //    processing order — they are dispatched via `Event::UserEvent`.
+      //  * The crossbeam receiver is lock-free and `try_recv` is cheap.
+      while let Ok(event) = self.receiver.try_recv() {
+        self.call_event_handler(event_handler, control_flow, event::Event::UserEvent(event));
+      }
+
       self.call_event_handler(event_handler, control_flow, event::Event::MainEventsCleared);
 
       if let Some(window_id) = resized_window_id {


Legend-Master:
Could you provide a reproducible example of the problem first?

Th4ler:
Hi @Legend-Master, here's a minimal reproducible example:

Setup:
1. Create a new Tauri 2.x app: pnpm create tauri-app (Vue/React/Svelte, doesn't matter)
2. Add a simple command that returns immediately:

```
#[tauri::command]
fn ping() -> String { "pong".to_string() }
```

3. Invoke it from the frontend on mount:

```
import { invoke } from "@tauri-apps/api/core";
onMount(async () => {
  console.log("result:", await invoke<string>("ping"));
});
```

On Android (real device, e.g. Samsung S22 Ultra / Android 16):

- pnpm tauri android build --debug
- Install + launch
- The invoke("ping") never resolves, the onMount hangs on await.
- adb logcat shows the request reaching Rust (handler runs), but no response ever gets back to the renderer.

With this fix applied:

- Same scenario, invoke("ping") resolves in <100ms, result: pong logs to console.

Key probe evidence (from debugging a private Tauri app):

- EventLoopProxy::send_event fires, ALooper_wake() is called, queue grows (I logged the message count).
- poll_all() consistently returns Poll::Event{ident:0} (the ndk_glue event pipe fd), never Poll::Wake.
- So the EventSource::User arm that drains self.receiver never runs, events sit in the channel forever.

Hope that helps, happy to provide more instrumentation logs if needed.

Legend-Master:
Could you show a screen recording of the repro? I can't seem to reproduce this.

clozanoruiz:
Hello,

I hit this independently and built a minimal reproduction before finding this PR. Posting it here since the thread is waiting on a reproducible example — this PR's diagnosis matches what I measured, and the patch fixes it on my devices.
Btw, I do not know tao's internals, and I used Claude Code to help instrument this example.


**Reproduction:** 
[tao-raw-android-repro.zip](https://github.com/user-attachments/files/31100713/tao-raw-android-repro.zip). It sends a user event through `EventLoopProxy` before `event_loop.run()`, stays busy 3 s, runs the loop, then sends a second event five seconds later.

There is no screen recording because the app deliberately draws nothing: the evidence is the log, and a recording would show a blank activity. If a video of a real app failing would still help, the Tauri app I originally hit this in shows it as a permanently blank the very first launch after install.

### Before (released tao 0.36.0)

```
[taoraw +     0 ms] sending user event 1 BEFORE run()
[taoraw +     0 ms] busy for 3000 ms before run()
[taoraw +  3000 ms] calling event_loop.run() now        <- event 1 NOT delivered
[taoraw +  5003 ms] sending user event 2
[taoraw +  5003 ms] >>> USER EVENT 1 RECEIVED           <- both, same millisecond
[taoraw +  5003 ms] >>> USER EVENT 2 RECEIVED
```

A heartbeat prints every second between +3000 and +5003, so the loop is running and idle throughout — the event is queued, not lost, and nothing delivers it until unrelated traffic arrives.

### After (same device, same APK pipeline, 0.36.0 + this PR)

```
[taoraw +     0 ms] sending user event 1 BEFORE run()
[taoraw +     0 ms] busy for 3000 ms before run()
[taoraw +  3000 ms] calling event_loop.run() now
[taoraw +  3000 ms] >>> USER EVENT 1 RECEIVED           <- delivered as run() starts
[taoraw +  5001 ms] sending user event 2
[taoraw +  5001 ms] >>> USER EVENT 2 RECEIVED
```
Only the patch differs between those two runs.


### Environment

Reproduced on two handsets with different Android versions, and on two tao releases:

- **Google Pixel 8a**, Android 17 (SDK 37), WebView 150.0.7871.181
- **Samsung Galaxy A22 5G** (SM-A226B), Android 13 (SDK 33), WebView 151.0.7922.83
- tao **0.36.0** and **0.35.3**, rustc 1.96.0, NDK 28.2.13676358, aarch64 debug
- the before/after pair above is the Galaxy A22 on 0.36.0

Timings are within a millisecond of each other on both devices. The same pattern delivers correctly on desktop (Linux/GTK).


### A Tauri app
`tauri-runtime-wry`'s `send_user_message` dispatches a command result through this proxy whenever it is off the main thread, and Tauri's `setup` hook runs before `run()` — so the answer to a command invoked during startup can be queued into exactly this window. The visible effect is a permanently blank first launch after install, no error and no panic, recoverable only by killing the app.

Happy to run any further probe, or test a revised patch, on either device.

Legend-Master:
@clozanoruiz Thanks, I'll checkout the repro next Monday, but could you post it in a repo so I can inspect the code before downloading it? Also a video recording will help a lot, even if it's just a white screen. Also if you can reproduce this in an emulator, the real time log + emulator screen would be really helpful.

Legend-Master:
@Th4ler Also you'll need to sign your commits for me to merge this

> https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

clozanoruiz:
@Legend-Master thanks — repo, video and an emulator run below.

**Repo:** https://github.com/clozanoruiz/tao-android-eventloop-repro

Everything that differs from the zipped example is either how is built (x86_64 for the emulator) and how is watched (log on screen instead of only in logcat). 

**It reproduces on a stock emulator**:

```sh
TARGET=x86_64-linux-android ./build.sh
cd android && ./gradlew assembleDebug
adb -e install -r app/build/outputs/apk/debug/app-debug.apk
```

Verified on `sdk_gphone64_x86_64`, Android 15 (API 35) — same result as the two physical phones, so that is now three environments and two tao releases.

**Video:** 
https://github.com/user-attachments/assets/0af6a062-0449-4e54-bae3-112a758bc687. The app mirrors its own logcat on screen.

What the video shows, and the log from the same run:

```
[taoraw +     0 ms] run_app() entered
[taoraw +     0 ms] sending user event 1 BEFORE run()
[taoraw +     0 ms] busy for 3000 ms before run(), like an app doing startup work
[taoraw +  1000 ms] heartbeat (loop thread is not blocked)
[taoraw +  2000 ms] heartbeat (loop thread is not blocked)
[taoraw +  3000 ms] calling event_loop.run() now
[taoraw +  3001 ms] heartbeat (loop thread is not blocked)
[taoraw +  4001 ms] heartbeat (loop thread is not blocked)
[taoraw +  5000 ms] sending user event 2 (loop has been running for ~5 s)
[taoraw +  5000 ms] >>> USER EVENT 1 RECEIVED
[taoraw +  5000 ms] >>> USER EVENT 2 RECEIVED
[taoraw +  5001 ms] heartbeat (loop thread is not blocked)
```

The two heartbeats between +3000 and +5000 are the point: the loop is running and idle, and event 1 is still not delivered. It arrives only when event 2 is sent, and then both land in the same millisecond.

**With the patch applied** (unconditional drain after each poll iteration), same device and same tao 0.36.0, only the patch differing:

```
[taoraw +  3000 ms] calling event_loop.run() now
[taoraw +  3000 ms] >>> USER EVENT 1 RECEIVED
```

Happy to run anything else.

Th4ler:
> @Th4ler Also you'll need to sign your commits for me to merge this
> 
> > https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

The PR now contains a single SSH-signed commit (d06e08d), and the duplicate changelog has been removed.

The CI workflows are currently awaiting maintainer approval because this PR comes from a fork.

Legend-Master:
I think the problem here is we need to migrate away from `ALooper_pollAll` to use `ALooper_pollOnce` instead.

> This API will not reliably respond to ALooper_wake. As such, this API is hidden and callers should migrate to ALooper_pollOnce. Binary compatibility is preserved to support already-compiled apps.
> https://developer.android.com/ndk/reference/group/looper#alooper_pollall

Legend-Master (approved):
> I think the problem here is we need to migrate away from `ALooper_pollAll` to use `ALooper_pollOnce` instead.
> 
> > This API will not reliably respond to ALooper_wake. As such, this API is hidden and callers should migrate to ALooper_pollOnce. Binary compatibility is preserved to support already-compiled apps.
> > https://developer.android.com/ndk/reference/group/looper#alooper_pollall

That being said, this PR is still needed with or without the migration, so I'll take it first, nice catch thanks!

Legend-Master:
@clozanoruiz the repro helped a lot, big thanks!

PR #1306, opened 2026-08-12
fix(windows): delay maximize window if hidden

### LINKED ISSUE #1142: The flashing splash page when launching the app in maximized state is confusing
**Error Description**
Initially, to fix the white splash screen that flashes when launching the app, I configured the tauri.conf.json file as follows:
"visible": false,
"transparent": true,

Then I executed the following code to display the window.
```
/// Fix window rendering (avoid white screen in WebView2)
async fn fix_window_rendering(handle: AppHandle) {
if let Some(window) = handle.get_webview_window("main") {
if let Err(e) = (|| -> Result<(), tauri::Error> {
if !window.is_maximized()? {
window.center()?;
window.show()?; //Avoid conflict with maximized state
}
Ok(())
})() {
log::error!("Unable to restore hidden window: {e}");
} else {
log::info!("Window successfully shown and centered");
}
```
Seems to be caused by the set_maximized function?

**Steps to Reproduce**
Maximize the app on every launch

**Expected Behavior**
The white page should not flicker.

**Platform and Version (please fill in the following information):**
Operating System: Windows 10

**Additional Information**
What's a good way to resolve this issue? I don't want to experience the flickering issue after saving the app in a maximized state; it's too annoying.

### LINKED ISSUE #14068: The flashing splash page when launching the app in maximized state is confusing
This issue is slightly different from the previous white startup page issue in that it only occurs when the startup is in a maximized state.

Here's what happened: To fix the white splash screen that flashes when launching the app, I configured the tauri.conf.json file as follows:
      "visible": false,
       "transparent": true,
Then I executed this code to show the window.
      /// Fix window rendering (avoids white screen in WebView2)
      async fn fix_window_rendering(handle: AppHandle) {
      if let Some(window) = handle.get_webview_window("main") {
      if let Err(e) = (|| -> Result<(), tauri::Error> {
      if !window.is_maximized()? {
      window.center()?;
      window.show()?; //Avoid conflict with maximized state
      }
      Ok(())
      })() {
      log::error!("Unable to restore hidden window: {e}");
      } else {
      log::info!("Window successfully shown and centered");
      }
Fixed successfully. It wasn't until I used tauri_plugin_window_state to save the app in a maximized state that a new startup issue emerged: the entire screen would briefly flicker (I noticed my computer wallpaper flickered before the app appeared, so I suspect this is still a rendering issue during startup???). This flicker is visible to the naked eye, but difficult to capture with screen recording software. Furthermore, the flicker doesn't occur every time the app is launched maximized; sometimes it doesn't, and sometimes it does. I'm not sure how to fix this, as launching the app with a fixed window size doesn't cause any splash screen issues. Any help would be greatly appreciated.

Legend-Master:
Fix #1142
Fix https://github.com/tauri-apps/tauri/issues/14068

On Windows, delay maximizing the window if it's hidden to avoid it flashing briefly. This also changed the size and position returned to pre-maximized states if you query them with the window being hidden and called `maximize`

This is mainly for the tauri's window-state plugin since you need to set `visible: false` to avoid flashing. This will allow you to set both maximized + hidden without causing a flashing window.

### DIFF
--- a/.changes/windows-delay-maximize-if-hidden.md
+++ b/.changes/windows-delay-maximize-if-hidden.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+On Windows, delay maximizing the window if it's hidden to avoid it flashing briefly. This also changed the size and position returned to pre-maximized states if you query them with the window being hidden and called `maximize`
--- a/src/platform_impl/windows/window_state.rs
+++ b/src/platform_impl/windows/window_state.rs
@@ -357,7 +357,10 @@ impl WindowFlags {
       }
     }
 
-    if diff.contains(WindowFlags::MAXIMIZED) || new.contains(WindowFlags::MAXIMIZED) {
+    if (diff.contains(WindowFlags::MAXIMIZED) || new.contains(WindowFlags::MAXIMIZED))
+      // This is to avoid the window from flashing
+      && !(new.contains(WindowFlags::MAXIMIZED) && !new.contains(WindowFlags::VISIBLE))
+    {
       unsafe {
         let _ = ShowWindow(
           window,


Legend-Master:
This change made so that the size and position remain the pre-maximized while the window's hidden, @FabianLars could you help test if this is also the case for macOS? (it seems to be the case on Linux)

```patch
diff --git a/examples/window.rs b/examples/window.rs
index c4804fc4..7dce477c 100644
--- a/examples/window.rs
+++ b/examples/window.rs
@@ -2,6 +2,8 @@
 
+use std::{sync::Arc, thread::sleep, time::Duration};
+
 use tao::{
   event::{Event, WindowEvent},
@@ -12,18 +14,31 @@ use tao::{
 fn main() {
 
-  let mut window = Some(
+  let mut window = Some(Arc::new(
     WindowBuilder::new()
       .with_inner_size(tao::dpi::LogicalSize::new(300.0, 300.0))
       .with_min_inner_size(tao::dpi::LogicalSize::new(200.0, 200.0))
+      .with_visible(false)
       .build(&event_loop)
       .unwrap(),
-  );
+  ));
+
+  // window.as_ref().unwrap().set_visible(false);
+  window.as_ref().unwrap().set_maximized(true);
+
+  let window_clone = window.clone();
+
+  std::thread::spawn(move || {
+    sleep(Duration::from_secs(3));
+    dbg!(window_clone.as_ref().unwrap().inner_size());
+    window_clone.as_ref().unwrap().set_visible(true);
+    // window_clone.as_ref().unwrap().set_maximized(true);
+  });
 
-    println!("{event:?}");
+    // println!("{event:?}");
 
     match event {
       Event::WindowEvent {

```

FabianLars:
just woke up so i may be dense, i just apply that diff, run `cargo run --example window` and see how the window behaves?
I think it looks good.

https://github.com/user-attachments/assets/3f00bdea-4dbf-4413-b451-0e6a61f4df85

Legend-Master:
Looking like the size is the maximized size on macOS? I can't think of if this can cause problems right now but I always want to be careful with this type of changes. (before this change, on Windows, if you have a hidden window and then call maximize, it will show the window very briefly and then hide it again, now if you query the size/pos of the window, you get the states after the maximization, but now you get the pre-maximized states)

FabianLars:
> Looking like the size is the maximized size on macOS?

Yes. is this good? 😂

Legend-Master:
Maybe? Like you can do something based on the window's size and position before showing the window. That being said, flashing it to get the correct size is a no go for me anyways so maybe let's just roll with this.

FabianLars (approved):

PR #1163, opened 2025-12-13
Change run loop observer mode to common modes to prevent freezing on iOS

rapporian:
Proposed fix for issue #1162 - **Needs maintainer validation**.

On iOS, **all events are blocked while the user is scrolling**.  This breaks every Tao-based iOS app that needs real-time updates during scroll.

**Affected**

- Signal/state updates
- Async task completions
- Cross-thread UI communication

iOS run loop modes:

| Mode                    | When Active | `DefaultMode` Observers | `CommonModes` Observers |
| ----------------------- | ----------- | ----------------------- | ----------------------- |
| `kCFRunLoopDefaultMode` | Normal      | ✅ Fire                 | ✅ Fire                 |
| `UITrackingRunLoopMode` | Scrolling   | ❌ Don't fire           | ✅ Fire                 |

The run loop **observers** are registered for `kCFRunLoopDefaultMode` only. During scroll, iOS switches to `UITrackingRunLoopMode` — observers aren't registered for that mode, so they don't fire and events don't get processed.

PR #1319, opened 2026-08-31
fix(ios): dispatch events while a scroll view is tracking

guilbaudl:
### Problem

On iOS, no event is dispatched to the application while a `UIScrollView` is
tracking a finger or decelerating. Events queued during that window are
delivered only once the scroll comes to rest, so a tap on a button outside the
scroller appears to do nothing for as long as the momentum lasts.

### Cause

`setup_control_flow_observers` registers all three control-flow observers in
`kCFRunLoopDefaultMode`:

```rust
CFRunLoopAddObserver(main_loop, begin_observer, kCFRunLoopDefaultMode);
CFRunLoopAddObserver(main_loop, main_end_observer, kCFRunLoopDefaultMode);
CFRunLoopAddObserver(main_loop, end_observer, kCFRunLoopDefaultMode);
```

While scrolling, UIKit runs the main run loop in `UITrackingRunLoopMode`, so
these observers never fire and nothing drives the event loop.

The `EventLoopProxy` wake source is already registered in
`kCFRunLoopCommonModes` (line 222), so the run loop *is* woken during a scroll
— it simply has no observer in that mode to dispatch what arrived. The existing
`kCFRunLoopExit => ()` comments ("Mode is changed to others like
`UITrackingRunLoopMode`") show the mode switch was already known about.

### Fix

Register the three observers in `kCFRunLoopCommonModes`, which includes
`UITrackingRunLoopMode`, matching the mode the wake source already uses.

### Testing

Verified on a physical iPhone 14 Pro (iOS 26.6.1) running a WebView-based
Wry/Dioxus application. Before: tapping a navigation button during a momentum
scroll did nothing until the scroll stopped. After: the tap is handled
immediately, with no change to scrolling or inertia and no perceptible jank
from the loop now running during tracking.

To be explicit about the limits of that evidence: this is one device and one
application, and it is a behaviour change for every iOS consumer — the event
loop now runs during tracking, so an app doing heavy work per iteration will do
that work while the user scrolls. That is the correct trade (the alternative is
dropped interactivity), but it is worth a maintainer's judgement.

### DIFF
--- a/.changes/ios-runloop-common-modes.md
+++ b/.changes/ios-runloop-common-modes.md
@@ -1,0 +1,5 @@
+---
+"tao": patch
+---
+
+On iOS, run the event loop observers in `kCFRunLoopCommonModes` so queued events are dispatched while a `UIScrollView` is tracking or decelerating, instead of waiting for the scroll to stop.
--- a/src/platform_impl/ios/event_loop.rs
+++ b/src/platform_impl/ios/event_loop.rs
@@ -311,7 +311,7 @@ fn setup_control_flow_observers() {
       control_flow_begin_handler,
       ptr::null_mut(),
     );
-    CFRunLoopAddObserver(main_loop, begin_observer, kCFRunLoopDefaultMode);
+    CFRunLoopAddObserver(main_loop, begin_observer, kCFRunLoopCommonModes);
 
     let main_end_observer = CFRunLoopObserverCreate(
       ptr::null_mut(),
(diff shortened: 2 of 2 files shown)


FabianLars:
thanks for the pr but there already is one for this https://github.com/tauri-apps/tao/pull/1163.

Also for future contributions, we require all commits in a PR to be signed
> https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #1332, opened 2026-09-14
chore(deps): update gtk crates to 0.19

### LINKED ISSUE #985: Related to the deprecation of MainContext::channel in GTK on Linux.

A warning is currently being issued that `glib::MainContext::channel` is deprecated in the Linux event loop-related code.

```
warning: use of deprecated associated function `gtk::glib::main_context_channel::<impl gtk::glib::MainContext>::channel`: Use an async channel, from async-channel for example, on the main context using spawn_future_local() instead
   --> src/platform_impl/linux/event_loop.rs:227:71
    |
227 |     let (window_requests_tx, window_requests_rx) = glib::MainContext::channel(Priority::default());
    |                                                                       ^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated associated function `gtk::glib::main_context_channel::<impl gtk::glib::MainContext>::channel`: Use an async channel, from async-channel for example, on the main context using spawn_future_local() instead
   --> src/platform_impl/linux/event_loop.rs:241:55
    |
241 |       let (device_tx, device_rx) = glib::MainContext::channel(glib::Priority::default());
    |                                                       ^^^^^^^

```

We can update the currently used `MainContext::channel` to `async_channel` in `event_loop.rs` on Linux.

**Additional context**

See: https://gtk-rs.org/gtk4-rs/stable/latest/book/main_event_loop.html or https://discourse.gnome.org/t/help-required-to-migrate-from-dropped-maincontext-channel-api/20922

### LINKED ISSUE #1281: RUSTSEC-2024-0429: Unsoundness in `Iterator` and `DoubleEndedIterator` impls for `glib::VariantStrIter`
> Unsoundness in `Iterator` and `DoubleEndedIterator` impls for `glib::VariantStrIter`

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unsound                |
| Package             | `glib`                      |
| Version             | `0.18.5`                   |
| URL                 | [https://github.com/gtk-rs/gtk-rs-core/pull/1343](https://github.com/gtk-rs/gtk-rs-core/pull/1343) |
| Date                | 2024-03-30                         |

The `VariantStrIter::impl_get` function (called internally by implementations of the `Iterator` and `DoubleEndedIterator` traits for this type) was unsound, resulting in undefined behaviour.

An immutable reference `&amp;p` to a `*mut libc::c_char` pointer initialized to `NULL` was passed as an argument to a C function that that mutates the pointer behind `&amp;p` in-place (i.e. as an out-argument), which was unsound. After changes in recent versions of the Rust compiler, these unsound writes through `&amp;p` now seem to be completely disregarded when building the `glib` crate with optimizations.

This subsequently caused all calls of `VariantStrIter::impl_get` to violate the safety requirements of the `std::ffi::CStr::from_ptr` function - which requires its argument to be a valid pointer to a C-style string - resulting in crashes due to `NULL` pointer dereferences.

This was fixed by passing the out-argument pointer explitly as `&amp;mut p` instead of `&amp;p`.

This issue has been present since this code was initially added in `glib` v0.15.0. The mismatch in mutability was likely missed (and not raised as an error by the compiler) because the C function wrapped by `VariantStrIter::impl_get` is variadic (`glib_sys::g_variant_get_child`), and the pointer in question is one of the variadic arguments.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2024-0429.html) for additional details.

Legend-Master:
Closes #1329
Closes #1330
Closes #1331

Closes https://github.com/tauri-apps/tao/issues/1281
Closes #985
Closes https://github.com/tauri-apps/tao/security/dependabot/1

This also increases the MSRV to 1.92.

Migrate glib channel to `async-channel`, adopted the approach used in #1104, a method suggested by the glib author: https://discourse.gnome.org/t/help-required-to-migrate-from-dropped-maincontext-channel-api/20922
Disclosure: Used codex to save some typings.

Notes to reviewers: I'm not the most familiar with this so please double check if I missed something, I have tested on WSL and everything seems to work for me but still we want to take extra caution on this.


.changes/gtk-0.19.md
```diff
@@ -0,0 +1,5 @@
+---
+tao: minor
+---
+
+Updated gtk crates to 0.19
```
johncarmack1984, line 5: This is the line downstream reads when `gtk_window()` stops type-checking against their gtk 0.18, so worth saying so here:

```suggestion
Updated the gtk crates to 0.19 (glib 0.22). The `gtk` types exposed through `platform::unix` (`WindowExtUnix::gtk_window`, `default_vbox`, `EventLoopWindowTargetExtUnix::gtk_app`) change with them, so downstream crates need the same bump.
```

Cargo.toml
```diff
@@ -7,7 +7,7 @@ authors = [
   "The winit contributors",
 ]
 edition = "2021"
-rust-version = "1.85"
+rust-version = "1.92"
```
johncarmack1984, line 10: This bump lets clippy turn on `manual_is_multiple_of` (gated on 1.87), which flags one line of untouched code at `src/icon.rs:108`. Not this PR's doing; if you want `cargo clippy` clean in the same go:

```rust
      if !rgba.len().is_multiple_of(PIXEL_SIZE) {
```

(icon.rs isn't in the diff, so no suggestion button for that one.)

johncarmack1984 (commented):
Read the channel migration file by file and ran it, since the PR asked for a second pair of eyes on exactly that. Nothing blocking.

**Context ownership.** `spawn_local` in `new_gtk` runs inside the `context.with_thread_default(...)` scope, and glib 0.22's `spawn_local_with_priority` takes its own `acquire()` anyway (`main_context_futures.rs:590`). No panic path there whether or not `EventLoop::new` runs on `main`.

**portal.rs, `attach(None)` to `MainContext::default()`.** Same context in tao: the only thread-default pushes are the two `with_thread_default` scopes around `new_gtk` and `run_return`, both on the default context.

**Dropped `JoinHandle`s.** glib's handle has an explicit `abort()` and no `Drop` impl, so both tasks stay alive after the handle goes.

**`send_blocking`.** Unbounded channel, so it never blocks and only errs once the receiver is gone, the contract `glib::Sender::send` had. device.rs keeps its info-log-and-break on Err. Task lifetimes match the old sources too: the device loop still checks `run` after each event, and the window-request loop ends when the last `window_requests_tx` clone drops.

**Ran it** in a bookworm container on 1.92, Xvfb plus a session bus. `cargo check`, `clippy`, `doc` and `test` with `--all-features --locked`: green. window, multithreaded, control_flow, theme, custom_events and request_redraw_threaded run 20 s each without a panic. A probe that moves a `Window` into a worker thread and fires 600 `set_title` / `set_inner_size` / `set_outer_position` / `set_theme` calls at it gives identical counts on this branch and on `dev`, three runs each: 299 Resized, 299 Moved, 150 ThemeChanged, final size 449×349. Happy to paste the probe if it's useful.

**Nits.** One MSRV-surfaced clippy lint and a changelog wording idea are inline. `dev`'s three `MainContext::channel` deprecation warnings are gone with this. `cargo audit` on the branch lock shows only the already-allowed `paste`; nothing new from async-channel's tree.

PR #1334, opened 2026-09-15
fix(windows): make event loop runner refcount atomic

cedstrom:
## Summary

`EventLoopRunnerShared<T>` is `Rc<EventLoopRunner<T>>`. The runner is main-thread-only, so a non-atomic refcount looks correct in isolation — but the *handle* does not stay on one thread in practice, and that makes the count racy.

This changes the alias to `Arc`, so only the refcount becomes atomic. The runner keeps its `Cell`/`RefCell` interior and remains neither `Send` nor `Sync`, so it still cannot be **used** off the event loop thread — only *sharing the handle* becomes sound.

## Why the count is racy today

tauri-runtime-wry embeds `EventLoopWindowTarget` — and therefore this `Rc` — in `DispatcherMainThreadContext`, which it declares `Send + Sync`:

```rust
// SAFETY: we ensure this type is only used on the main thread.
unsafe impl<T: UserEvent> Send for DispatcherMainThreadContext<T> {}
unsafe impl<T: UserEvent> Sync for DispatcherMainThreadContext<T> {}
```

That context is reachable from `AppHandle`, `Window` and `Webview` (via `WryHandle { context: Context<T> }` and the window/webview dispatchers), and Tauri clones those handles off the main thread on ordinary paths. The most common one is not app code at all: the `#[tauri::command]` macro extracts arguments **inside** the spawned future for `async fn` commands, and `AppHandle::from_command` calls `message.webview()`, which clones a `Webview` — two `Context` clones per invocation, on an async-runtime worker thread.

Two threads then run `incq`/`decq` on the same counter with no `lock` prefix.

## Why it is hard to attribute

A lost update is silent. It only surfaces once the count wraps past zero to `usize::MAX`, at which point the next `Rc::clone` hits the overflow guard and calls `core::intrinsics::abort()`. On Windows that lowers to a bare `ud2`, so the process dies with `STATUS_ILLEGAL_INSTRUCTION` (0xC000001D) — no unwind, no panic payload, no useful backtrace.

Diagnosed from a WER record by disassembling the fault RVA. The guards all branch to one shared `ud2`:

```
1406fa875:  lock incq (%rax) ; jle  0x1406fa9b0   <- Arc
1406fa8e0:       incq (%rsi) ; je   0x1406fa9b0   <- this Rc, non-atomic
...
1406fa9b0:  ud2
```

`je` (abort when the increment yields `0`) is the signature of a count that had already wrapped, i.e. more decrements than increments — a lost update, not genuine overflow.

Downstream has hit this shape before: tauri-runtime-wry's `windows: Arc<WindowsStore>` carries the comment `// changing this to an Rc will cause frequent app crashes.` This is the same hazard, one field over.

## Scope

Mechanical. This `Rc` was the only one in the Windows backend, and no `Rc`-specific API (`ptr_eq`, `downgrade`, `into_raw`, `strong_count`) was in use — the alias, one `Rc::new`, and two imports.

## Testing

- `cargo check --all-targets` on Windows: clean, no new warnings.
- Applied the same change to the `tao` version the affected app pins (0.35.2) and built the full downstream stack through it — `tao` → `wry` → `tauri-runtime` → `tauri-runtime-wry` → `tauri` → app — which compiles clean, confirming nothing downstream depends on the handle being `Rc`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016fYE37qXvCgqHMRYQ3GKFw


Legend-Master:
See https://github.com/tauri-apps/tauri/pull/15411

PR #1339, opened 2026-09-17
fix(windows): `with_focused(false)` only applies to the first show

### LINKED ISSUE #1338: Windows: after with_focused(false), maximizing and minimizing the window go wrong
**Describe the bug**

On Windows, if a window is built with `with_focused(false)`, maximizing and minimizing it later go wrong:

- tao 0.35.3 (what Tauri uses now): minimizing a maximized window shows it at its normal size for a moment, then it minimizes. A Markpad user ran into this: https://github.com/sftwrdotdev/Markpad/issues/776
- `dev` (0.37): `set_maximized(true)` and `set_minimized(true)` get undone right away. The window goes back to its normal size.

Windows built without `with_focused(false)` don't do this.

**Steps To Reproduce**

```rust
use std::time::{Duration, Instant};
use tao::{
  event::{Event, StartCause},
  window::WindowBuilder,
};

fn main() {
    .with_focused(false)
    .build(&event_loop)
    .unwrap();

  let mut step = 0;
    if let Event::NewEvents(StartCause::Init | StartCause::ResumeTimeReached { .. }) = event {
      match step {
        0 => window.set_maximized(true),
        1 => {
          println!("after set_maximized(true): is_maximized = {}", window.is_maximized());
          window.set_minimized(true);
        }
        _ => {
          println!("after set_minimized(true): is_minimized = {}", window.is_minimized());
          return;
        }
      }
      step += 1;
      *control_flow = ControlFlow::WaitUntil(Instant::now() + Duration::from_secs(2));
    }
  });
}
```

On `dev` it prints:

```
after set_maximized(true): is_maximized = false
after set_minimized(true): is_minimized = false
```

On 0.35.3 both are `true`, but you can see the window jump to its normal size right before it minimizes.

**Expected behavior**

`with_focused(false)` only keeps the first show from taking focus. After that, the window maximizes and minimizes like any other window.

**Why it happens**

`with_focused(false)` sets a flag that is meant to be used once, on the first show. `apply_diff` clears it on its own copy, so the stored flag stays set. From then on, every window state change also calls `ShowWindow(SW_SHOWNOACTIVATE)`, which puts a maximized or minimized window back to its normal size.

On 0.35.3 that call runs before maximize/minimize, so you see a flash. Since #1296 it runs after them, so they get undone.

**Platform and Versions**

OS: Windows 11 (build 26200). The Markpad report was on Windows 10.
Rustc: 1.98.1

_Edited: the first version of this issue guessed that the `SW_MAXIMIZE` call in `apply_diff` was the cause. I didn't have a Windows machine then, and that guess was wrong._

PathGao:
Closes #1338.

`with_focused(false)` is meant to keep only the first show from taking focus. The flag was cleared on a copy, so it stayed set, and every later maximize or minimize also ran `ShowWindow(SW_SHOWNOACTIVATE)`, which puts the window back to its normal size. On 0.35 you see a flash when minimizing a maximized window. On `dev`, maximize and minimize get undone.

This clears the stored flag once the window has been shown.

Verification on Windows 11 (build 26200), with the example from #1338:

- before: `is_maximized = false`, `is_minimized = false`
- after: both `true`, and the first show still doesn't take focus

```sh
cargo test
```

Not tested on Windows 10.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


PathGao:
@Legend-Master could you take a look at this one too? Small Windows-only fix for #1338, CI is green.

Legend-Master:
Not sure I understand the bug here?

PathGao:
**What the user sees**

sftwrdotdev/Markpad#776, Windows 10. The window is built with `.focused(false)` and draws its own title bar. It is maximized, the user clicks minimize.

```
expected:  [maximized] ──> taskbar
actual:    [maximized] ──> [visibly drops to its restored size] ──> taskbar
           reopening:      [restored size first] ──> snaps back to maximized
```

Four times out of five. Minimizing the same window from the taskbar never does it, because that path does not go through `apply_diff`.

**Why**

One `set_minimized(true)` ends up sending three `ShowWindow` calls:

```
apply_diff():
  1. SW_MAXIMIZE         the window is already maximized
  2. SW_MINIMIZE         minimize it
  3. SW_SHOWNOACTIVATE   <- extra. "Displays a window in its most recent
                            size and position", which undoes 1 and 2.
                            This is the flash the reporter sees.
```

Step 3 is there for `with_focused(false)`: don't take focus on the first show. It should run once and then clear `MARKER_DONT_FOCUS`. It clears it on a copy:

```rust
fn apply_diff(mut self, ...)            // `self` is a copy, `WindowFlags` is `Copy`
    self.set(MARKER_DONT_FOCUS, false)  // mutates the copy
// returns, the copy is dropped
```

The stored flags in `WindowState.window_flags` are never touched, so the marker stays set for the window's whole life. From then on every flag change on a visible window (maximize, minimize, always-on-top, decorations, ...) ends with another step 3.

On 0.35 that is the flash above. On `dev` it is worse: `set_maximized(true)` leaves `is_maximized()` at `false`. The 30-line repro is in #1338.

Legend-Master:
I see, seems like a regression from #1296

I don't think what you described matches the PR or consistent at all though...

PathGao:
You're right, and my ordering was wrong.

On 0.35.3 the `VISIBLE` block still runs at the top of `apply_diff`, so a minimize on a maximized window is `SW_SHOWNOACTIVATE` → `SW_MAXIMIZE` → `SW_MINIMIZE`: the window flashes at its restored size on the way down, but ends up in the right state. That is what the Markpad reporter saw.

#1296 moved that block to the end, so on `dev` it is `SW_MAXIMIZE` → `SW_MINIMIZE` → `SW_SHOWNOACTIVATE` and the show now undoes them, which is why `set_maximized(true)` leaves `is_maximized()` at `false`.

So the stale `MARKER_DONT_FOCUS` predates #1296 — #1296 turned it from a flash into a wrong state. This PR clears the stored flag, which fixes both orderings. If you would rather handle it inside the block #1296 added, I am happy to redo it that way.

